PR #1, opened 2026-03-30
Add SeaORM migrations and entities for all tables

brendanbyrne:
## Summary
- Set up SeaORM with SQLite as the database layer, with a migration crate at `backend/migration/`
- Created 7 migrations covering all 10 tables from DESIGN.md: characters, bodies, wheels, gliders, cups, tracks, drink_types, users, runs, run_flags
- Handled the circular FK between `users.preferred_drink_type_id → drink_types` and `drink_types.created_by → users` by deferring `created_by` to a separate ALTER TABLE migration
- Auto-generated SeaORM entity files in `backend/src/entities/` from the migrated database
- Migrations run automatically on server startup; the hello endpoint still works
- Added `[lints.rust] warnings = "deny"` to treat warnings as errors
- Configured cargo target directory on the Linux filesystem for faster WSL2 builds

## Non-obvious details
- UUID columns (users, runs, run_flags, drink_types) are stored as TEXT in SQLite — easier to debug than BLOB, and SeaORM maps them to `String`
- Timestamps are TEXT in ISO 8601 format (SQLite has no native timestamp type)
- The `data/` directory at repo root holds the SQLite database; `.gitkeep` is tracked but `.db` files are gitignored
- Entity files are auto-generated and currently `#[allow(unused)]` since no routes use them yet

## Open questions
- Environment variable / config strategy needs design before auth work begins (added to Cowork handoff)
- Database file management strategy (git lfs? external storage?) also queued for Cowork

## Test plan
- [ ] `cd backend && cargo build` — zero warnings, zero errors
- [ ] `cd backend && cargo run` — server starts, prints "Listening on http://localhost:3000"
- [ ] `curl http://localhost:3000/api/v1/hello` — returns `{"message":"Hello from Beerio Kart!"}`
- [ ] Delete `data/beerio-kart.db` and restart — database recreated automatically


backend/migration/src/m20260330_000006_create_runs.rs
```diff
@@ -0,0 +1,102 @@
...
+    Table,
+    Id,
+    UserId,
+    TrackId,
+    CharacterId,
+    BodyId,
+    WheelsId,
+    GliderId,
+    TrackTime,
+    Lap1Time,
```
brendanbyrne, line 95: **Important:** `DeriveIden` converts `Lap1Time` → `lap1_time`, but DESIGN.md specifies `lap_1_time` (with underscore before the digit). The generated entity confirms this: the fields are `lap1_time`, `lap2_time`, `lap3_time`.

To match the spec, you can override the column name with an explicit `#[sea_orm(iden = "lap_1_time")]` on the enum variant, or rename the variants to something like `Lap_1Time` — though that's ugly. The cleaner approach is to use raw SQL for these three column definitions, similar to migration 5.

Alternatively, update DESIGN.md to use `lap1_time` — it's a cosmetic choice, but worth being intentional about.

backend/migration/src/m20260330_000006_create_runs.rs
```diff
@@ -0,0 +1,102 @@
...
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .create_table(
+                Table::create()
+                    .table(Runs::Table)
+                    .if_not_exists()
+                    .col(ColumnDef::new(Runs::Id).text().not_null().primary_key())
+                    .col(ColumnDef::new(Runs::UserId).text().not_null())
+                    .col(ColumnDef::new(Runs::TrackId).integer().not_null())
```
brendanbyrne, line 24: **Suggestion:** SQLite doesn't auto-index FK columns. Once leaderboards and personal stats start querying `runs` by `user_id`, `track_id`, and `drink_type_id`, these will need indexes. Not blocking for Phase 1, but worth a follow-up migration when the query patterns are concrete.

backend/migration/src/m20260330_000007_create_run_flags.rs
```diff
@@ -0,0 +1,69 @@
...
+                Table::create()
+                    .table(RunFlags::Table)
+                    .if_not_exists()
+                    .col(ColumnDef::new(RunFlags::Id).text().not_null().primary_key())
+                    .col(
+                        ColumnDef::new(RunFlags::RunId)
+                            .text()
+                            .not_null()
+                            .unique_key(),
+                    )
```
brendanbyrne: **Suggestion:** The `unique_key()` on `run_id` means a run can only ever have **one flag row total** (not just one active one). If a flag is resolved and the same run later needs re-flagging, you'd have to update/delete the old row rather than insert a new one. Is that intentional? If you want "one unresolved flag per run" instead, drop the unique constraint here and enforce it in application code.

backend/src/main.rs
```diff
@@ -8,7 +13,26 @@ struct HelloResponse {
...
-    let app = Router::new().route("/api/v1/hello", get(hello));
+    // Load .env file if present (non-fatal if missing)
+    dotenvy::dotenv().ok();
+
+    let database_url = std::env::var("DATABASE_URL")
+        .unwrap_or_else(|_| "sqlite:../data/beerio-kart.db?mode=rwc".to_string());
+
+    // Connect to the database. The ?mode=rwc flag creates the file if it
+    // doesn't exist yet. SeaORM wraps sqlx under the hood.
+    let db: DatabaseConnection = Database::connect(&database_url)
```
brendanbyrne, line 24: **Suggestion:** SQLite requires `PRAGMA foreign_keys = ON` per connection for FK constraints to actually be enforced — it's off by default. Without it, you can insert runs referencing nonexistent users and nothing stops you. Consider enabling this after connecting, e.g.:
```rust
db.execute_unprepared("PRAGMA foreign_keys = ON").await?;
```
Or via sqlx connection options. Can be a follow-up.

backend/migration/src/m20260330_000006_create_runs.rs
brendanbyrne: Let's go with the convention that matches what's expected by `DeriveIden`.  Go from `lap_1_time` to `lap1_time`.  Create a handoff for Cowork to update the design doc.

backend/migration/src/m20260330_000006_create_runs.rs
brendanbyrne: I don't understand what you mean.  Please expand on this PR's associated up markdown file.

backend/migration/src/m20260330_000007_create_run_flags.rs
brendanbyrne: I agree.  Record your thoughts in the handoff for cowork.

backend/src/main.rs
brendanbyrne: I agree that this is a problem.  Record your thoughts in the handoff for cowork.

backend/migration/src/m20260330_000006_create_runs.rs
brendanbyrne: Sounds good. Created a Cowork handoff at `.claude/claude-code-handoff.md` to update DESIGN.md.

backend/migration/src/m20260330_000006_create_runs.rs
brendanbyrne: Wrote up a detailed explanation in `reviews/pr-1-review.md` covering what indexes are, why FK columns need them, what the fix looks like, and why it's fine to defer to Phase 3.

backend/migration/src/m20260330_000007_create_run_flags.rs
brendanbyrne: Added to the Cowork handoff for a design decision on whether to keep the UNIQUE constraint or drop it in favor of app-level enforcement.

backend/src/main.rs
brendanbyrne: Added to the Cowork handoff. The fix is straightforward — just need a design decision on whether to address it in this PR or as a follow-up.

PR #20, opened 2026-05-02
Consolidate migrations 1–12 into single initial schema migration

brendanbyrne:
## Summary

Replace the 12 incremental `m20260330_*` migration files with a single consolidated `m20260101_000001_initial_schema.rs` that produces the same final schema.

This adopts the new prelaunch schema-change convention (CLAUDE.md → "Schema changes (prelaunch)", added in 9c0c4bf): pre-launch we keep all schema in one migration file and edit it in place rather than appending. The leading `01` differentiates this as the post-squash starting point and avoids confusion with the dated incremental files.

This is the prerequisite for PRs 3D-1 / 3D-2 (Pending Races feature, see `reviews/design/2026-04-19-pending-races-and-grace-period.md`), which will edit the consolidated migration to add the `session_race_participations` table and modify `session_participants`.

## Verification

- `sqlite3 .schema` was dumped from a fresh DB before the change (12 migrations applied) and after the change (consolidated migration applied). `diff` is empty — schemas are byte-identical.
- `cargo test` passes: 119 unit + 27 + 21 + 8 + 10 = **185 tests**, all green.
- `cargo clippy` and `cargo fmt` pass via pre-commit hook.

## Notes

- The circular FK between `users.preferred_drink_type_id` and `drink_types.created_by` is broken the same way the original 12 migrations did: create `drink_types` without `created_by`, then create `users`, then `ALTER TABLE drink_types ADD COLUMN created_by TEXT REFERENCES users(id)`. SQLite supports inline `REFERENCES` on `ADD COLUMN` but SeaORM's builder doesn't reliably emit it for `ALTER TABLE`.
- The partial unique index `idx_session_participants_one_active_session ... WHERE left_at IS NULL` is created via raw SQL because SeaORM's builder doesn't support partial-index `WHERE` clauses.
- Lookup tables, tracks, drink_types, and users use the SeaORM builder. Sessions, session_participants, session_races, runs, and run_flags use raw SQL — matching the existing pattern from migrations 008–011 where raw SQL was used for `datetime_text` typing and other SQLite-specific features.
- No behavior change, no new tables, no new columns. Pure refactor of how the schema is expressed.

## Test plan

To verify locally:

1. Check out the branch:
   ```
   git fetch origin phase-3/3d-0-consolidate-migrations
   git checkout phase-3/3d-0-consolidate-migrations
   ```
2. Delete the dev DB (per the prelaunch convention — schema rebuilds on next boot):
   ```
   rm data/db/beerio-kart.db
   ```
3. Boot the backend and confirm migrations + seeding succeed:
   ```
   cd backend && JWT_SECRET=anything cargo run
   ```
   Expected: `Seeding complete` log line, no errors.
4. Stop the server, then dump the schema:
   ```
   sqlite3 data/db/beerio-kart.db ".schema" > /tmp/schema-after.sql
   ```
5. Optionally regenerate entities to confirm SeaORM still recognizes all column types:
   ```
   just entities
   git diff backend/src/entities/   # expect no changes
   ```
6. Run the test suite:
   ```
   cd backend && cargo test
   ```
   Expected: all 185 tests pass.


brendanbyrne:
### Code review

PR #21, opened 2026-05-02
Pending races: schema, race-creation hook, query, get_session_detail integration

brendanbyrne:
## Summary

Read-side foundation for the Pending Races & Grace Period feature, per [`reviews/design/2026-04-19-pending-races-and-grace-period.md`](reviews/design/2026-04-19-pending-races-and-grace-period.md) (fully signed off). Enforcement (skip endpoint, ordered-submit check, HTTP integration tests) is intentionally **deferred to PR 3D-2**.

After this PR the system can:
- Track per-race presence at race-creation time.
- Track which races a user has pending (correctly excluding submitted, skipped, absent-at-creation, and grace-expired).
- Surface a user's pending list as `your_pending` on the session detail polling response.

After this PR the system **cannot yet**:
- Skip a pending race via API (no endpoint).
- Block out-of-order run submission (`create_run` still accepts any order).

That gap is intentional — 3D-1 is read-side only.

## Schema (consolidated migration edited in place)

- `session_participants` gains `UNIQUE(session_id, user_id)` — single mutable row per (session, user) replaces the old multi-row history design. The partial unique index `(user_id) WHERE left_at IS NULL` stays; it still enforces "active in only one session at a time."
- New `session_race_participations` table: composite PK `(session_race_id, user_id)`, `ON DELETE CASCADE` on the **race-side** FK (`skip_turn`'s delete-and-replace relies on it), plain `REFERENCES` on the user-side FK (rows are never deleted by design — preserves the audit trail), index on `user_id`.

## Service changes

- **`helpers::insert_race_participations`** snapshots currently-present users into `session_race_participations` inside the same transaction as the race insert. Wired into `next_track` and `skip_turn`. Single `insert_many` per call.
- **`join_session`** rewritten for single-row mutate-on-rejoin. Within the 5-min grace window, rejoin only clears `left_at`; outside the window, `joined_at` resets to `NOW()` — which forfeits pre-gap pending records via the `get_pending_races` derivation.
- **`get_pending_races`** implements the 6-clause derivation from DESIGN.md (presence row + not skipped + no run + within grace + post-`joined_at` + session is active). Uses `find_by_statement` per the §5.1 ORM convention for multi-table JOINs. The `sessions.status = 'active'` clause was added in commit `a975225` after PR review caught a phantom-pending bug for sessions that close while a user is still inside their grace window.
- **`get_session_detail`** now takes `Option<&str> requesting_user_id` and includes `your_pending: Vec<SessionRaceInfo>` in the response. Route handler passes the auth user.

`SessionRaceInfo.submissions` is intentionally empty for pending races to avoid an N+1 on the polling path; the design review didn't require submissions on pending entries. Easy to add later if needed.

## Tests

**19 new tests, 204 total** (was 185 on main). All pass.

- 3 race-creation hook tests in `services::sessions::tests::test_create_session_race_*` — presence snapshot correctness, atomicity (FK violation rolls back the race insert).
- 6 pending-query tests in `services::sessions::tests::test_pending_*` — happy path, exclusions for submitted/skipped/absent-at-creation, ordering, no schema cap.
- 7 grace-period tests in `services::sessions::tests::test_pending_*` and `test_*_grace_*` — in-session, within grace, expired, within-grace rejoin (preserves `joined_at`), after-grace rejoin (resets `joined_at`, forfeits pending), multiple flaps, lazy-check intent assertion.
- 1 closed-session regression test (`test_pending_excludes_closed_session_even_within_grace`) added in `a975225` — exercises the `close_stale_sessions` path.
- 2 helper unit tests in `services::helpers::tests::insert_race_participations_*` added in `2be5b18` — direct coverage for the snapshot helper, including the empty-snapshot edge case.

`backdate_participant`, `insert_session_race`, and `insert_race_participation` test helpers added for fine-grained timestamp manipulation without sleeping.

## Reviewer note

**Reset the dev DB before booting** to pick up the schema change (per the prelaunch convention):
```
rm data/db/beerio-kart.db
```
The consolidated migration rebuilds the schema on next startup; SeaORM seeds run automatically.

## What this PR does NOT do

- `POST /sessions/:id/races/:race_id/skip` endpoint — PR 3D-2.
- Ordered-submit enforcement in `create_run` — PR 3D-2.
- HTTP-layer integration tests for the new behavior — PR 3D-2.
- Frontend consumption of `your_pending` — separate UI design.

### 1. Test suite (fast path — covers all logic)

```bash
git fetch origin phase-3/3d-1-pending-foundation
git checkout phase-3/3d-1-pending-foundation
rm -f data/db/beerio-kart.db
cd backend && cargo test
```

**Expected:** `test result: ok. 204 passed; 0 failed` total across all targets — broken down as 138 lib + 27 api_integration + 21 auth_integration + 8 auth_middleware + 10 session_integration.

To run only the tests added/modified in this PR:

```bash
cd backend && cargo test --lib -- \
    services::sessions::tests::test_create_session_race \
    services::sessions::tests::test_pending \
    services::sessions::tests::test_rejoin_within_grace \
    services::sessions::tests::test_rejoin_after_grace \
    services::sessions::tests::test_multiple_short_flaps \
    services::sessions::tests::test_lazy_check \
    services::helpers::tests::insert_race_participations
```

**Expected:** 19 tests, all pass.

### 2. End-to-end smoke test (HTTP + direct DB inspection)

Requires `jq` and `sqlite3`. Run from the repo root with the backend booted in another terminal:

```bash
# Terminal 1 — backend (will block; leave running)
cd backend && rm -f ../data/db/beerio-kart.db && JWT_SECRET=smoke-test-secret COOKIE_SECURE=false cargo run
# Wait for "Seeding complete" before continuing.
```

```bash
# Terminal 2 — set up env vars and run the smoke test
API=http://localhost:3000/api/v1
DB=data/db/beerio-kart.db

# Sanity check: backend is up. `curl -s` swallows connection errors silently,
# which would otherwise leave $A empty and produce confusing `null` outputs
# downstream. Run this first to fail fast if the server isn't listening yet.
curl -fsS "$API/hello" -o /dev/null && echo "backend up" || \
    { echo "backend not reachable on $API — start it in terminal 1 first" >&2; }

# Register two users; capture access tokens and user IDs.
A=$(curl -s -X POST "$API/auth/register" \
    -d '{"username":"alice","password":"smoketest123"}')
TOKEN_A=$(echo "$A" | jq -r .access_token)
USER_A=$(echo  "$A" | jq -r .user.id)

B=$(curl -s -X POST "$API/auth/register" \
    -d '{"username":"bob","password":"smoketest123"}')
TOKEN_B=$(echo "$B" | jq -r .access_token)
USER_B=$(echo  "$B" | jq -r .user.id)

echo "alice=$USER_A bob=$USER_B"
```

**Expected:** two non-empty UUIDs printed. If either is `null`, the register call failed — re-check the backend log.

```bash
# Step A: alice creates a session, bob joins.
SESSION=$(curl -s -X POST "$API/sessions" \
    -H "Authorization: Bearer $TOKEN_A" \
    -d '{"ruleset":"random"}' | jq -r .id)

curl -s -X POST "$API/sessions/$SESSION/join" \
    -H "Authorization: Bearer $TOKEN_B" -o /dev/null -w "%{http_code}\n"
# Expected: 204
```

```bash
# Step B: alice picks the next track. This is the call that should
# create one session_race_participations row per currently-present user.
RACE=$(curl -s -X POST "$API/sessions/$SESSION/next-track" \
    -H "Authorization: Bearer $TOKEN_A" | jq -r .id)
echo "race=$RACE"
```

**Verify in the DB** — should show **2 rows** (one for alice, one for bob), both with `skipped_at = NULL`:

```bash
sqlite3 -header -column "$DB" \
    "SELECT user_id, created_at, skipped_at
     FROM session_race_participations
     WHERE session_race_id = '$RACE';"
```

**Expected output (2 rows, ordering may differ):**
```
user_id                               created_at                  skipped_at
------------------------------------  --------------------------  ----------
<alice's UUID>                        2026-05-02T...                          
<bob's UUID>                          2026-05-02T...                          
```

```bash
# Step C: alice's pending list contains the race.
curl -s "$API/sessions/$SESSION" \
    -H "Authorization: Bearer $TOKEN_A" | jq '(.your_pending | length), .your_pending[0].id'
# Expected: 1 then "$RACE"
```

```bash
# Step D: alice leaves the session. Session stays active because bob is
# still in it — host transfer keeps it open.
curl -s -X POST "$API/sessions/$SESSION/leave" \
    -H "Authorization: Bearer $TOKEN_A" -o /dev/null -w "%{http_code}\n"
# Expected: 204
```

```bash
# Step E: while still within 5 minutes, alice's pending is preserved.
curl -s "$API/sessions/$SESSION" \
    -H "Authorization: Bearer $TOKEN_A" | jq '.your_pending | length'
# Expected: 1
```

```bash
# Step F: backdate alice's left_at to 6 minutes ago to simulate the
# grace window expiring without waiting. Then re-query.
sqlite3 "$DB" \
    "UPDATE session_participants
       SET left_at = datetime('now', '-6 minutes')
     WHERE session_id = '$SESSION' AND user_id = '$USER_A';"

curl -s "$API/sessions/$SESSION" \
    -H "Authorization: Bearer $TOKEN_A" | jq '.your_pending | length'
# Expected: 0  (grace expired → no accessible pending)
```

```bash
# Step G: confirm the participation row still exists in the DB
# (lazy check — no cleanup task touches it).
sqlite3 -header "$DB" \
    "SELECT COUNT(*) AS rows_remaining
     FROM session_race_participations
     WHERE user_id = '$USER_A';"
# Expected: rows_remaining = 1
```

### 3. Closed-session regression check (added in `a975225`)

This exercises the bug from the first review run — pending should NOT be returned for a session that's closed, even when the user is well within their grace window.

```bash
# Continuing from the smoke test above.
# Backdate bob's left_at to within grace, then have bob leave so the
# session closes via the host-leaves-last cascade.
sqlite3 "$DB" \
    "UPDATE session_participants
       SET left_at = datetime('now', '-2 minutes')
     WHERE session_id = '$SESSION' AND user_id = '$USER_B';"

# Force-close the session directly (mirrors the close_stale_sessions effect).
sqlite3 "$DB" \
    "UPDATE sessions SET status = 'closed' WHERE id = '$SESSION';"

# Bob is within grace AND has a presence row for the race — but the session
# is closed, so pending must be empty.
curl -s "$API/sessions/$SESSION" \
    -H "Authorization: Bearer $TOKEN_B" | jq '.your_pending | length'
# Expected: 0  (closed session → no accessible pending, even within grace)
```

If all expected outputs match, the read-side foundation is working end-to-end including the closed-session edge case.

### DIFF
--- a/backend/migration/src/m20260101_000001_initial_schema.rs
+++ b/backend/migration/src/m20260101_000001_initial_schema.rs
@@ -269,6 +269,12 @@ impl MigrationTrait for Migration {
         .await?;
 
         // ---- Session participants ----
+        //
+        // One row per (session, user). Leave/rejoin mutates this row rather
+        // than appending a new one — `joined_at` is the start of the current
+        // presence segment, `left_at` is null while present. Per-race presence
+        // is captured by `session_race_participations` (below) at race-creation
+        // time, not derived from this table.
 
         conn.execute_unprepared(
             "CREATE TABLE IF NOT EXISTS session_participants (
@@ -276,7 +282,8 @@ impl MigrationTrait for Migration {
                     session_id TEXT NOT NULL REFERENCES sessions(id),
                     user_id TEXT NOT NULL REFERENCES users(id),
                     joined_at datetime_text NOT NULL,
-                    left_at datetime_text
+                    left_at datetime_text,
+                    UNIQUE(session_id, user_id)
                 )",
         )
         .await?;
@@ -308,6 +315,35 @@ impl MigrationTrait for Migration {
                     created_at datetime_text NOT NULL,
                     UNIQUE(session_id, race_number)
                 )",
+        )
+        .await?;
+
+        // ---- Session race participations ----
+        //
+        // One row per (race, user) for every user present at race-creation
+        // time. The row is the proof of "this user was present when this race
+        // was created" — pending-state derivation reads from here, not from a
+        // walk of `session_participants` history. `skipped_at` flips when the
+        // user explicitly forfeits the race.
+        //
+        // ON DELETE CASCADE on session_race_id matters for `skip_turn`, which
+        // deletes-and-replaces the current race; cascading drops the old
+        // participations atomically.
+
+        conn.execute_unprepared(
+            "CREATE TABLE IF NOT EXISTS session_race_participations (
+                    session_race_id TEXT NOT NULL REFERENCES session_races(id) ON DELETE CASCADE,
+                    user_id TEXT NOT NULL REFERENCES users(id) ON DELETE CASCADE,
+                    created_at datetime_text NOT NULL,
+                    skipped_at datetime_text,
+                    PRIMARY KEY (session_race_id, user_id)
+                )",
+        )
+        .await?;
+
+        conn.execute_unprepared(
+            "CREATE INDEX idx_session_race_participations_user
+                 ON session_race_participations(user_id)",
         )
         .await?;
 
@@ -375,6 +411,8 @@ impl MigrationTrait for Migration {
         conn.execute_unprepared("DROP TABLE IF EXISTS run_flags")
             .await?;
         conn.execute_unprepared("DROP TABLE IF EXISTS runs").await?;
+        conn.execute_unprepared("DROP TABLE IF EXISTS session_race_participations")
+            .await?;
         conn.execute_unprepared("DROP TABLE IF EXISTS session_races")
             .await?;
         conn.execute_unprepared("DROP TABLE IF EXISTS session_participants")
--- a/backend/migration/src/m20260101_000001_initial_schema.rs
+++ b/backend/migration/src/m20260101_000001_initial_schema.rs
@@ -326,14 +326,20 @@ impl MigrationTrait for Migration {
         // walk of `session_participants` history. `skipped_at` flips when the
         // user explicitly forfeits the race.
         //
-        // ON DELETE CASCADE on session_race_id matters for `skip_turn`, which
-        // deletes-and-replaces the current race; cascading drops the old
-        // participations atomically.
+        // ON DELETE CASCADE on session_race_id is required: `skip_turn`
+        // deletes-and-replaces the current race, and the old participations
+        // need to evaporate atomically with the race delete.
+        //
+        // user_id intentionally does NOT cascade — DESIGN.md guarantees these
+        // rows are never deleted (they're the audit trail of "who was present
+        // when"). If a user-deletion path is ever added, it must explicitly
+        // decide what to do with these rows rather than letting them silently
+        // vanish.
 
         conn.execute_unprepared(
             "CREATE TABLE IF NOT EXISTS session_race_participations (
                     session_race_id TEXT NOT NULL REFERENCES session_races(id) ON DELETE CASCADE,
-                    user_id TEXT NOT NULL REFERENCES users(id) ON DELETE CASCADE,
+                    user_id TEXT NOT NULL REFERENCES users(id),
                     created_at datetime_text NOT NULL,
                     skipped_at datetime_text,
                     PRIMARY KEY (session_race_id, user_id)
--- a/backend/src/services/helpers.rs
+++ b/backend/src/services/helpers.rs
@@ -6,8 +6,8 @@
 use chrono::Utc;
 use rand::seq::SliceRandom;
 use sea_orm::{
-    ActiveModelTrait, ColumnTrait, Condition, ConnectionTrait, EntityTrait, PrimaryKeyTrait,
-    QueryFilter, Set, sea_query::Expr,
+    ColumnTrait, Condition, ConnectionTrait, EntityTrait, PrimaryKeyTrait, QueryFilter, Set,
+    sea_query::Expr,
 };
 
 use crate::domain::enums::SessionStatus;
@@ -80,16 +80,26 @@ pub async fn insert_race_participations<C: ConnectionTrait>(
         .all(txn)
         .await?;
 
-    for participant in present {
-        session_race_participations::ActiveModel {
+    if present.is_empty() {
+        // `insert_many` with an empty iterator panics in some SeaORM
+        // versions; bail before the call. Nothing to snapshot is a valid
+        // state — e.g. all participants left between race creation and the
+        // helper call (shouldn't happen in normal flow, but defended here).
+        return Ok(());
+    }
+
+    let rows = present
+        .into_iter()
+        .map(|p| session_race_participations::ActiveModel {
             session_race_id: Set(session_race_id.to_string()),
-            user_id: Set(participant.user_id),
+            user_id: Set(p.user_id),
             created_at: Set(now),
             skipped_at: Set(None),
-        }
-        .insert(txn)
+        });
+
+    session_race_participations::Entity::insert_many(rows)
+        .exec(txn)
         .await?;
-    }
 
     Ok(())
 }
@@ -181,9 +191,12 @@ mod tests {
     use super::*;
     use sea_orm::{ActiveModelTrait, Set};
 
-    use crate::entities::{characters, cups, tracks};
+    use sea_orm::PaginatorTrait;
+
+    use crate::entities::{characters, cups, session_race_participations, tracks};
     use crate::test_helpers::{
-        create_user, insert_participant, insert_session, seed_tracks_for_test, setup_db,
+        create_user, insert_participant, insert_session, insert_session_race, seed_tracks_for_test,
+        setup_db,
     };
 
     // --- load_active_session ---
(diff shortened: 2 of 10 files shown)


brendanbyrne:
### Code review

Found 1 issue:

1. `get_pending_races` lacks a `sessions.status = 'active'` filter, so a closed session can still return non-empty `your_pending`. When the host leaves last and `transfer_host_or_close` fires `SessionClosed`, only the leaving user's `left_at` is set — other participants who were still present keep `left_at IS NULL`, so the grace clause `(sp.left_at IS NULL OR sp.left_at >= $3)` passes and the API surfaces phantom pendings the user can never resolve (the route layer also doesn't gate `GET /sessions/:id` on status). DESIGN.md explicitly states pending rows "become inaccessible via normal API paths (filtered out by grace check or `session.status = 'active'` filter)" — both filters are required.

https://github.com/brendanbyrne/beerio-kart/blob/c4160fc53de6073ecff5f7621cefa0f755a76422/backend/src/services/sessions.rs#L487-L502

backend/migration/src/m20260101_000001_initial_schema.rs
```diff
@@ -311,6 +318,35 @@ impl MigrationTrait for Migration {
...
+        // user explicitly forfeits the race.
+        //
+        // ON DELETE CASCADE on session_race_id matters for `skip_turn`, which
+        // deletes-and-replaces the current race; cascading drops the old
+        // participations atomically.
+
+        conn.execute_unprepared(
+            "CREATE TABLE IF NOT EXISTS session_race_participations (
+                    session_race_id TEXT NOT NULL REFERENCES session_races(id) ON DELETE CASCADE,
+                    user_id TEXT NOT NULL REFERENCES users(id) ON DELETE CASCADE,
```
brendanbyrne: **Important — inconsistent `ON DELETE CASCADE` on `user_id`.**

Every other `user_id` FK in this migration uses plain `REFERENCES` (default `RESTRICT`): `sessions.created_by`, `sessions.host_id`, `session_participants.user_id` (line 283), `session_races.chosen_by` (line 314), `runs.user_id` (line 355). The comment above (lines 329–331) only justifies the cascade on `session_race_id` (for `skip_turn`'s delete-and-replace).

Cascading on `user_id` also contradicts DESIGN.md's "Rows are never deleted" invariant for this table (DESIGN.md:341) — if a user-deletion path is ever added, this would silently wipe a user's full participation audit trail without any code-side opt-in.

No functional consequence today (no user-deletion code exists), but the divergence appears unintentional and the design review (`reviews/design/2026-04-19-pending-races-and-grace-period.md` §2) only specified the FK without cascade.

**Suggested fix:**

```suggestion
                    user_id TEXT NOT NULL REFERENCES users(id) ON DELETE CASCADE,
```

→

```sql
user_id TEXT NOT NULL REFERENCES users(id),
```

And correspondingly drop `on_delete = "Cascade"` from the `Users` relation in `backend/src/entities/session_race_participations.rs`. Keep the cascade on `session_race_id` — that one is intentional and documented.

*Background: in SQL, `ON DELETE` controls what happens to child rows when the parent row is deleted. `RESTRICT` (the default) makes the parent delete fail if children exist; `CASCADE` deletes the children too. We want the former for `users` so the constraint surfaces loudly if a user-deletion path is ever added — letting us decide explicitly what to do with their history.*

brendanbyrne (commented):
## Code review

**Verdict: Approve with minor follow-ups.** The blocking issue from the prior review run (`get_pending_races` missing the `sessions.status = 'active'` filter) was already fixed in commit a9752252. Remaining items are non-blocking.

**Automated checks.** `cargo clippy --all-targets --all-features` clean. `cargo test` — 202 tests pass (136 lib + 21 api + 27 auth + 8 middleware + 10 session integration), zero failures. Frontend untouched.

### Critical
None.

### Important

1. **`ON DELETE CASCADE` on `session_race_participations.user_id`** is inconsistent with every other `user_id` FK in the consolidated migration (`sessions.created_by`, `sessions.host_id`, `session_participants.user_id`, `session_races.chosen_by`, `runs.user_id` all use plain `REFERENCES`/RESTRICT). The migration's own comment only justifies the cascade on `session_race_id` (for `skip_turn`'s delete-and-replace). Cascading on `user_id` contradicts DESIGN.md's "Rows are never deleted" invariant for this table — if user-deletion is ever added, the audit trail silently disappears. No functional consequence today (no user-deletion path), but the divergence appears unintentional. Inline suggestion below.

### Suggestions

1. **N+1 INSERTs in `insert_race_participations`** — for 2–8 participants per session perf is negligible, but `Entity::insert_many` would be a single statement and read more naturally. See inline comment.
2. **Misleading comment in `join_session`** — the `existing` query filters by both `session_id` AND `user_id`, so reaching the `Some(row) / left_at IS NULL` branch means "already in *this* session," not "a different active session." Inline comment with suggested rewording below.
3. **`helpers.rs` module doc claims "Each helper is independently tested"** but `insert_race_participations` is exercised only via integration tests in `sessions.rs`. Either add a unit test in `helpers.rs::tests` (zero-present-users edge case is the obvious one) or relax the doc claim. Inline comment below.

### DESIGN.md accuracy

DESIGN.md was correctly updated in this PR (lines 341, 447, 455) — schema, the now-required `session.status = 'active'` filter with its rationale, and forfeiture-on-close are all documented. Code matches DESIGN.md after commit a9752252. No further DESIGN.md updates needed.

### What's good

- **16 new tests, all passing**, with good coverage of presence-snapshot correctness, transactional atomicity (FK-violation rollback verifies the race insert is rolled back), the 5-clause pending derivation, and grace boundaries.
- **The follow-up commit's test refactor was sharp:** `test_pending_accessible_when_within_grace` and `test_pending_inaccessible_when_grace_expired` were rewritten to use 2-participant sessions, isolating the grace check from the status filter — they were previously passing for the wrong reason in solo sessions where leaving auto-closed.
- **Consolidated migration edited in place** per the CLAUDE.md prelaunch convention.
- **`find_by_statement` correctly used** for the multi-table pending join, matching DESIGN.md's ORM Usage convention.
- **Strong doc comments** on `get_pending_races`, `join_session`, and `transfer_host_or_close` — the "earliest-joined ≈ most-tenured in current segment" note is exactly the kind of subtle invariant worth writing down.
- **`test_lazy_check_assertion`** is a clean negative-requirement assertion documenting that no background sweeper touches `session_race_participations`. No-op test bodies are unusual, but the doc comment is the load-bearing part — defensible here.

backend/src/services/helpers.rs
```diff
@@ -53,6 +53,47 @@ pub async fn require_active_participant<C: ConnectionTrait>(
...
+    for participant in present {
+        session_race_participations::ActiveModel {
+            session_race_id: Set(session_race_id.to_string()),
+            user_id: Set(participant.user_id),
+            created_at: Set(now),
+            skipped_at: Set(None),
+        }
+        .insert(txn)
+        .await?;
+    }
```
brendanbyrne: **Suggestion — N+1 INSERTs.**

This loop issues one `INSERT` per present participant. For typical session sizes (2–8 players) the perf impact is negligible, but `Entity::insert_many` would issue a single statement and read more naturally:

```rust
let rows: Vec<session_race_participations::ActiveModel> = present
    .into_iter()
    .map(|p| session_race_participations::ActiveModel {
        session_race_id: Set(session_race_id.to_string()),
        user_id: Set(p.user_id),
        created_at: Set(now),
        skipped_at: Set(None),
    })
    .collect();

if !rows.is_empty() {
    session_race_participations::Entity::insert_many(rows)
        .exec(txn)
        .await?;
}
```

Not blocking — flagging for a future cleanup.

Separately, the module doc on line 4 (`//! Each helper is independently tested.`) is now technically inaccurate: `insert_race_participations` is exercised only via integration tests in `sessions.rs` (`test_create_session_race_*`). Either add a focused unit test here (zero-present-users edge case, plus a basic happy path) or relax the doc claim.

backend/src/services/sessions.rs
```diff
@@ -479,18 +609,53 @@ pub async fn join_session(
...
+            .insert(&txn)
+            .await?;
+        }
+        Some(row) => {
+            let Some(left_at) = row.left_at else {
+                // Already present in this session. The partial unique index on
+                // user_id WHERE left_at IS NULL would catch a stray duplicate
+                // INSERT, but `check_not_in_any_session` should have already
+                // surfaced this as a 409 — reaching here implies a different
+                // active session, which is a user-level conflict.
```
brendanbyrne: **Suggestion — comment reasoning is inverted.**

The `existing` query at lines 612–619 filters by both `session_id` AND `user_id`, so reaching this `Some(row) / left_at IS NULL` arm means the user is already active in *this* session — not "a different active session." `check_not_in_any_session` would have caught a *different* active session before we ever got here.

The branch is effectively unreachable in practice (because `check_not_in_any_session` rejects first), but the comment as written misleads anyone debugging the conflict path.

**Suggested rewrite:**

```rust
// Already present in this session. This branch should be
// unreachable in practice — `check_not_in_any_session` (line
// 610) rejects the call before we get here whenever the user
// has any active participant row, including in this session.
// Returning Conflict defensively in case that guard is ever
// weakened. The partial unique index on user_id WHERE
// left_at IS NULL is the database-level backstop.
```

PR #26, opened 2026-05-04
PR-B2: per-connection PRAGMAs via SqliteConnectOptions

brendanbyrne:
## Summary

Switch from `Database::connect(url)` + post-hoc `PRAGMA foreign_keys = ON` to building the SQLx pool with `SqliteConnectOptions` and wrapping it via `SqlxSqliteConnector::from_sqlx_sqlite_pool`. Per [`seaorm.md` § 8](docs/coding-standards/seaorm.md), several SQLite PRAGMAs (`busy_timeout`, `synchronous`, `foreign_keys`) are *per-connection* — running them once at startup only configures the connection that served that statement.

This is **PR-B2** in [`docs/compliance-plan.md`](docs/compliance-plan.md) Phase B.

## The bug

```rust
// pre-PR-B2
let db = Database::connect(&database_url).await?;
db.execute_unprepared("PRAGMA foreign_keys = ON").await?;
```

`PRAGMA foreign_keys = ON` only takes effect on the connection that runs it. With `max_connections > 1`, when a query triggers a new pool connection (the original is busy or has been recycled), that connection has FKs **disabled** by default. SQLite then silently accepts inserts referencing non-existent foreign keys.

This is silent corruption — no error, no warning, just a row referencing a nonexistent parent. It would surface only when the query happens to land on the original connection that ran the startup PRAGMA, making the bug pool-size-dependent and intermittent.

## The fix

Refactored connection setup into `backend/src/db.rs` (new module). `main.rs` now calls `db::connect(url)`. The function:

```rust
let sqlx_opts = SqliteConnectOptions::from_str(url)?
    .create_if_missing(true)
    .journal_mode(SqliteJournalMode::Wal)
    .synchronous(SqliteSynchronous::Normal)
    .busy_timeout(Duration::from_secs(5))
    .foreign_keys(true)
    .log_statements(log::LevelFilter::Debug);

let pool = SqlitePoolOptions::new()
    .max_connections(5)
    .min_connections(1)
    .acquire_timeout(Duration::from_secs(5))
    .idle_timeout(Some(Duration::from_secs(60)))
    .connect_with(sqlx_opts)
    .await?;

Ok(SqlxSqliteConnector::from_sqlx_sqlite_pool(pool))
```

`SqliteConnectOptions` fires its PRAGMA configuration at **every** new pool connection (sqlx's connection-setup hook), so every connection in the pool has FKs enforced.

## Settings

| Setting | Value | Why |
|---|---|---|
| `create_if_missing` | `true` | Replaces the `?mode=rwc` URL flag — works even without it. |
| `journal_mode` | `WAL` | Better concurrent-read behavior than the default `delete` mode. WAL is sticky (database-file-level) — applies once and persists. |
| `synchronous` | `Normal` | WAL-mode best practice; durable enough without `Full`'s fsync penalty. |
| `busy_timeout` | `5s` | Per-connection; absorbs short write contention without surfacing `database is locked` to callers. |
| `foreign_keys` | `true` | Per-connection; the bug fix. |
| `log_statements` | `Debug` | Replaces SeaORM's `sqlx_logging_level(LevelFilter::Debug)` — same logging behavior in the SqlxSqliteConnector path. |
| `max_connections` | `5` | SQLite serializes writes — bigger pools just fight for the write lock. |
| `min_connections` | `1` | Keeps one warm connection. |
| `acquire_timeout` | `5s` | Bound the wait when the pool is saturated. |
| `idle_timeout` | `60s` | Recycle idle connections so PRAGMA state stays fresh. |

## Tests

`backend/src/db.rs` has two new tests:

- **`test_connect_enforces_foreign_keys_on_inserts`** — the behavioral test. Uses `sqlite:file:<uuid>?mode=memory&cache=shared` (so multiple pool connections share the same DB), runs migrations, then attempts to insert a `users` row with `preferred_character_id = 99999`. Asserts the insert fails — proves FK enforcement reaches the actual data path.
- **`test_connect_uses_wal_journal_mode`** — temp-file SQLite. Memory DBs report `"memory"` for `PRAGMA journal_mode`, so an on-disk path is needed to verify WAL is actually applied. Cleans up the `.db`, `.db-shm`, `.db-wal` files at the end.

`test_helpers::setup_db()` is intentionally not migrated to use `db::connect()`. It uses `sqlite::memory:` (per-connection ephemeral DB) and applies the PRAGMA inline — works fine with `max_connections=1`-equivalent behavior. Migrating it would force decisions about `cache=shared` test isolation and is out of scope here.

## Dependencies

Two direct deps added (both already transitively present; pinned to match sea-orm 1.x's internal versions):

- `sqlx = { version = "0.8", default-features = false, features = ["sqlite", "runtime-tokio-rustls"] }` — for `SqliteConnectOptions` / `SqlitePoolOptions`.
- `log = "0.4"` — for `log::LevelFilter::Debug`.

## Risk

Low–medium. Changes pool sizing and PRAGMA application. The migration is mechanical (the SeaORM team's documented pattern for per-connection PRAGMAs), but it's the kind of change that can have surprising production behavior on first deploy if any code path was implicitly relying on FK-disabled semantics. None observed in this codebase.

## Verification

- [x] `cargo build --all-targets` — clean
- [x] `cargo clippy --all-targets` — clean
- [x] `cargo +nightly fmt --check` — clean
- [x] `cargo test` — all 232 tests pass (155 lib + 27 + 21 + 8 + 19 integration; +2 from `db::tests`)

## Test plan

- [ ] `cd backend && cargo test --lib db::tests` — runs the 2 new connection tests; both pass.
- [ ] `cd backend && cargo test` — full suite, 232 pass.
- [ ] `cargo clippy --all-targets` — clean.
- [ ] `cargo +nightly fmt --check` — clean.
- [ ] **Reset the dev DB before booting** (PRAGMA changes don't require it, but if you reuse an existing DB the WAL files will appear next to it on first boot — that's expected): `rm data/db/beerio-kart.db*` then `just dev`. Boot should succeed, `data/db/beerio-kart.db-wal` and `.db-shm` should appear (WAL mode confirmed at runtime).
- [ ] Smoke: register a user via `POST /auth/register`, then `cd backend && sqlite3 ../data/db/beerio-kart.db "PRAGMA foreign_keys"`. Should return `1` *(note: this only confirms FKs are on for the sqlite3 connection — the more meaningful check is `cargo test --lib db::tests` which exercises the pool path).*

## Follow-ups (not in scope)

- After this merges, check the PR-B2 row in [`docs/compliance-plan.md`](docs/compliance-plan.md).
- **PR-B3** (Argon2 spawn_blocking + semaphore) is independent of this; can land any time.
- `test_helpers::setup_db()` could eventually be migrated to use `db::connect()` for consistency, but only as part of a larger test-infrastructure refactor (the schema-drift test in PR-X2 will need similar treatment).


brendanbyrne:
### Code review

PR #85, opened 2026-05-08
ci: harden workflows with concurrency, least-privilege, timeouts, lychee retries

brendanbyrne:
## Summary

Four CI hardening changes across both workflows. All are low-risk best-practice tightening — no functional changes to what the workflows do.

- **Concurrency on link-check.yml** — new pushes to a PR cancel the prior run instead of stacking, matching the existing pattern in `coverage.yml`. Group key falls back to `github.sha` for pushes to `main` so main builds never cancel each other.
- **`permissions: contents: read` on link-check.yml** — least-privilege. Without an explicit block, the workflow inherits the repo-default `GITHUB_TOKEN` scopes, which can be broader than this read-only workflow needs. Coverage already had this; link-check now matches.
- **`timeout-minutes` per job** — 20 min for coverage, 10 min for link-check. GitHub's default cap is 360 minutes (6 hours), so a hung step today silently burns Actions minutes for hours. Picked numbers ~2x current p95 runtime so flaky-but-recoverable runs aren't killed.
- **`persist-credentials: false` on `actions/checkout`** — by default checkout writes the `GITHUB_TOKEN` into `.git/config` so subsequent steps can `git push`. Neither workflow pushes, so this is unused token surface. Recommended by [GitHub's hardening guide](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-secrets).

A follow-up PR will SHA-pin third-party actions and add Dependabot to keep them current.

## Test plan

The Actions tab is the verification surface — there's no local way to exercise these.

- [ ] **Trigger both workflows on this PR.** This PR touches `.github/workflows/*.yml` but no `.md`, so the `paths:` filter on `link-check` will skip it. To force a run, comment any markdown file (or push a trivial md edit on this branch). Coverage runs unconditionally on PR.
- [ ] **Confirm timeouts show up.** On the Actions run page, expand the `coverage` / `link-check` job — the header should show `Timeout: 20 minutes` (resp. 10).
- [ ] **Confirm cancel-in-progress works for link-check.** Push a second commit to this branch within ~30s of the first. The first link-check run should transition to "cancelled" and the second should start. (Requires the run to actually be in progress when the second push lands — easiest to test by editing two `.md` files in quick succession.)
- [ ] **Confirm permissions are read-only.** On the run page, expand the workflow yaml view — `permissions: contents: read` should be visible at the top of `link-check.yml`.
- [ ] **Sanity-check coverage still uploads to Codecov.** The `Upload to Codecov` step should succeed; `persist-credentials: false` only affects the local checkout, not the Codecov action which uses its own token via the `CODECOV_TOKEN` secret (or OIDC).

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -16,8 +16,11 @@ jobs:
   coverage:
     name: Backend coverage
     runs-on: ubuntu-latest
+    timeout-minutes: 20
     steps:
       - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
 
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@stable
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -25,14 +25,26 @@ jobs:
       - uses: actions/checkout@v4
         with:
           persist-credentials: false
-      - name: Run lychee
+      # Two-pass retry: most runs end at attempt 1. Attempt 2 fires only
+      # if attempt 1 fails, and catches transient transport-level blips
+      # (connection-resets, brief upstream outages) that lychee's internal
+      # --max-retries can't ride out on its own. URL exclusions live in
+      # `.lycheeignore`. Lychee's retry config: --max-retries / --retry-
+      # wait-time / --timeout — exponential backoff, ~35s of retry window
+      # per URL across 4 attempts.
+      - name: Run lychee (attempt 1)
+        id: lychee1
+        continue-on-error: true
         uses: lycheeverse/lychee-action@v2
         with:
-          # URL exclusions live in `.lycheeignore` at the repo root.
-          # --max-retries / --retry-wait-time / --timeout ride out transient
-          # upstream blips (github.com 502s, slow blog hosts) without
-          # excluding otherwise-good URLs. Backoff is exponential, so 3s
-          # initial gives ~35s of retry window per URL across 4 attempts.
+          args: --verbose --no-progress --max-retries 4 --retry-wait-time 3 --timeout 30 --exclude-path docs/drafts --exclude-path docs/handoffs './**/*.md'
+          fail: true
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Run lychee (attempt 2 if attempt 1 failed)
+        if: steps.lychee1.outcome == 'failure'
+        uses: lycheeverse/lychee-action@v2
+        with:
           args: --verbose --no-progress --max-retries 4 --retry-wait-time 3 --timeout 30 --exclude-path docs/drafts --exclude-path docs/handoffs './**/*.md'
           fail: true
         env:
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -10,11 +10,21 @@ on:
     paths:
       - '**/*.md'
 
+concurrency:
+  group: link-check-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
 jobs:
   link-check:
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
       - name: Run lychee
         uses: lycheeverse/lychee-action@v2
         with:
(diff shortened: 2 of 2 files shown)


brendanbyrne (commented):
## Verdict: Approve

**What this PR does:** Adds `concurrency`, `permissions: contents: read`, `timeout-minutes`, and `persist-credentials: false` to both workflows, bringing `link-check.yml` into parity with `coverage.yml` and tightening both. No functional change to what the workflows do.

**Areas reviewed:** `design.md`, `.claude/CLAUDE.md`. No coding-standards file covers GitHub Actions workflow security; review is judgment-based.

### What's verified

- Concurrency group key falls back to `github.sha` for `main` pushes — correct (main builds shouldn't cancel each other).
- 20 min / 10 min timeouts at ~2x p95 well-reasoned.
- `persist-credentials: false` is safe for both workflows (neither pushes via the local checkout token).
- `permissions: contents: read` is the right scope for both — coverage reads source to compute `lcov.info`, lychee reads `.md` files.
- Test plan correctly identifies the `paths:` trap (this PR has no `.md`, so link-check skips it) and gives the workaround.

### Findings

None blocking.

**Suggestion** — Optional defense-in-depth: `coverage.yml` could add `permissions: contents: read` at the **job level** in addition to the workflow level. If a future second job is added to the same workflow, it'd otherwise inherit the workflow scope without a fresh review of whether it needs more. Not a blocker — there's no second job today and adding it pre-emptively is premature.

### Design-doc accuracy

No changes required. `design.md` § Coverage & CI describes CI at the right altitude ("`cargo-llvm-cov` runs on every PR and push, uploads to Codecov"); concurrency / timeouts / permissions are below that abstraction level.

### Cross-PR note

#86 will conflict on the `actions/checkout` line in both files. Per #86's description, the resolution is mechanical (keep both: SHA pin **and** `persist-credentials: false`). Merge order: this PR first, then rebase #86, then merge #86.

brendanbyrne:
Acknowledged — skipping the job-level `permissions` suggestion per your own framing (premature with one job, easy to add when a second is introduced). Will revisit if/when we add a second job to coverage.yml.

brendanbyrne (commented):
## Re-review: Approve (still)

Two new commits since the previous review expand scope beyond the original "concurrency, least-privilege, timeouts":

1. `cdafbeb` — Tunes lychee args: `--max-retries 4 --retry-wait-time 3 --timeout 30`.
2. `3c37f7f` — Wraps lychee in a step-level two-pass retry (attempt 1 with `continue-on-error: true`, attempt 2 conditional on `steps.lychee1.outcome == 'failure'`).

### What's verified on the new retry layer

- **`outcome` vs `conclusion` is correct.** `outcome` is the step's true result before `continue-on-error` suppression; `conclusion` is post-suppression. Using `outcome == 'failure'` is the right gate.
- **Exit semantics are correct.** Attempt 1 success → attempt 2 skipped → job green. Attempt 1 fail + attempt 2 success → job green (last step succeeded, attempt 1 swallowed by `continue-on-error`). Attempt 1 fail + attempt 2 fail → job red (attempt 2 has no `continue-on-error`).
- **Comment block earns its keep.** The "two-layer retry" rationale (lychee's internal retries can't ride out transport-level blips before its HTTP stack initializes) is the kind of WHY a future maintainer would otherwise have to reverse-engineer.
- **CI is green.** Link-check on this branch ran 3m26s, consistent with a legit retry having fired during testing — good signal the pattern actually works.

### Findings

**🔵 Suggestion — PR title is now stale.** Title says *"harden workflows with concurrency, least-privilege, timeouts"* but the PR also adds a lychee retry layer. Consider: `ci: harden workflows with concurrency, least-privilege, timeouts, lychee retries`. Squash-merge will keep the title as the commit message, so a future `git log` reader gets a more accurate signal.

**🔵 Suggestion — step-level retry fires on any failure, not only transient ones.** Real broken-link failures will trigger attempt 2 too — no way to distinguish "genuine 404" from "DNS hiccup at startup" without inspecting lychee's output. Cost is ~lychee runtime once per fail (low single-digit minutes); acceptable. Calling out so it's a known property, not a surprise.

### Cross-PR

Conflict with #86 unchanged from prior review — same plan: merge this, rebase #86, merge #86.

brendanbyrne:
Updated title to include the lychee retry layer per your suggestion. On the second suggestion (attempt 2 fires on real broken-link failures too): acknowledged as a known property — the cost is at most one extra lychee runtime per genuinely-broken-link PR, which is acceptable given how rarely we add external links and how much noise the retry layer suppresses on normal PRs. Won't surprise me later because of the comment block in the workflow.

PR #86, opened 2026-05-08
ci: SHA-pin third-party actions and add Dependabot

brendanbyrne:
## Summary

Replaces every `uses:` floating tag with the commit SHA it currently resolves to, with the human-readable version preserved as a trailing comment. Adds `.github/dependabot.yml` so Dependabot watches for new releases and opens a PR each week to bump the SHAs. Without Dependabot, SHA pins go stale and stop receiving security fixes — the two changes are a package, not separable.

### Why pin

Floating tags (`@v4`, `@stable`, `@v2`) are mutable: the action maintainer can repoint them at any commit, including malicious code if their account is compromised. This is not theoretical — `tj-actions/changed-files` was hijacked in March 2025 and leaked secrets from thousands of repos. Pinning to a 40-char SHA makes the reference immutable; the only way it changes is via an explicit PR (which Dependabot generates and you review).

Recommended by [GitHub's hardening guide](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) and [OpenSSF Scorecard's `Pinned-Dependencies` check](https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies).

### Per-action notes

Two actions used non-standard tag patterns where the tag *is* the input parameter (`@stable`, `@cargo-llvm-cov`). Dependabot can't reason about those well, and pinning a branch ref's SHA is brittle. I switched both to canonical versioned tags with the input passed explicitly:

| Action | Before | After |
|---|---|---|
| `dtolnay/rust-toolchain` | `@stable` (branch) | `@e97e2d8…  # v1` + `toolchain: stable` |
| `taiki-e/install-action` | `@cargo-llvm-cov` (shortcut) | `@cca35ed…  # v2.77.1` + `tool: cargo-llvm-cov` |

The behavior is identical; the references are now Dependabot-trackable.

### Dependabot grouping

Six tracked actions × weekly schedule would mean up to six PRs/week. The `groups: actions: patterns: ["*"]` block bundles them into a single weekly PR. If you'd rather see them individually (to roll back one without the others), drop the `groups` block.

### Conflict with [#85](https://github.com/brendanbyrne/beerio-kart/pull/85)

This PR is branched from `main`, not from #85, per the project's squash-merge convention (stacking would silently lose this PR's content when #85 squashes). After #85 merges, this PR will conflict on the `uses:` lines — the resolution is mechanical (keep both: the SHA pin **and** the `persist-credentials: false` block on `actions/checkout`). I'm happy to rebase once #85 lands; just ping me.

### Out of scope (intentionally)

`package-ecosystem: cargo` and `package-ecosystem: bun` (or `npm`) for the backend and frontend dependency trees. Those are bigger decisions about dep-bump cadence — open as a separate question once you've seen how the GH Actions cadence feels.

## Test plan

- [ ] **Re-run both workflows on this PR.** This PR touches `.github/workflows/*.yml` and `.github/dependabot.yml` but no `.md`, so the `paths:` filter on `link-check` will skip. Force a run by editing any markdown file (or just push a trivial md edit on this branch). Coverage runs unconditionally on PR.
- [ ] **Confirm coverage workflow still passes.** All four pinned actions need to actually load and run: checkout, rust-toolchain, install-action, rust-cache, codecov-action. A typo in any SHA would surface as `Unable to resolve action`.
- [ ] **Confirm link-check workflow still passes.** Same check for the lychee pin.
- [ ] **Confirm Dependabot picks up the config.** After merge, visit the repo's "Insights → Dependency graph → Dependabot" tab. The `github-actions` ecosystem should appear there. (No PR will fire immediately — Dependabot waits for its weekly schedule, but the config is detected within a few minutes.)
- [ ] **Optional sanity check via local YAML parse.** `python3 -c 'import yaml; yaml.safe_load(open(".github/dependabot.yml"))'` — should produce no output if the file is valid.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -18,20 +18,23 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 20
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
         with:
           persist-credentials: false
 
       - name: Install Rust toolchain
-        uses: dtolnay/rust-toolchain@stable
+        uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # v1
         with:
+          toolchain: stable
           components: llvm-tools-preview
 
       - name: Install cargo-llvm-cov
-        uses: taiki-e/install-action@cargo-llvm-cov
+        uses: taiki-e/install-action@cca35edeb1d01366c2843b68fc3ca441446d73d3 # v2.77.1
+        with:
+          tool: cargo-llvm-cov
 
       - name: Cache cargo artifacts
-        uses: Swatinem/rust-cache@v2
+        uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2.9.1
         with:
           workspaces: backend
 
@@ -40,7 +43,7 @@ jobs:
         run: cargo llvm-cov --workspace --lcov --output-path lcov.info
 
       - name: Upload to Codecov
-        uses: codecov/codecov-action@v6
+        uses: codecov/codecov-action@57e3a136b779b570ffcdbf80b3bdc90e7fab3de2 # v6.0.0
         with:
           files: backend/lcov.info
           fail_ci_if_error: true
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -34,7 +34,7 @@ jobs:
           tool: cargo-llvm-cov
 
       - name: Cache cargo artifacts
-        uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2.9.1
+        uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
         with:
           workspaces: backend
 
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      actions:
+        patterns:
+          - "*"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,8 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    commit-message:
+      prefix: "ci"
     groups:
       actions:
         patterns:
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
         with:
           persist-credentials: false
       # Two-pass retry: most runs end at attempt 1. Attempt 2 fires only
@@ -35,7 +35,7 @@ jobs:
       - name: Run lychee (attempt 1)
         id: lychee1
         continue-on-error: true
-        uses: lycheeverse/lychee-action@v2
+        uses: lycheeverse/lychee-action@8646ba30535128ac92d33dfc9133794bfdd9b411 # v2.8.0
         with:
           args: --verbose --no-progress --max-retries 4 --retry-wait-time 3 --timeout 30 --exclude-path docs/drafts --exclude-path docs/handoffs './**/*.md'
           fail: true
@@ -43,7 +43,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - name: Run lychee (attempt 2 if attempt 1 failed)
         if: steps.lychee1.outcome == 'failure'
-        uses: lycheeverse/lychee-action@v2
+        uses: lycheeverse/lychee-action@8646ba30535128ac92d33dfc9133794bfdd9b411 # v2.8.0
         with:
           args: --verbose --no-progress --max-retries 4 --retry-wait-time 3 --timeout 30 --exclude-path docs/drafts --exclude-path docs/handoffs './**/*.md'
           fail: true


.github/workflows/coverage.yml
```diff
@@ -17,18 +17,21 @@ jobs:
...
 
       - name: Install cargo-llvm-cov
-        uses: taiki-e/install-action@cargo-llvm-cov
+        uses: taiki-e/install-action@cca35edeb1d01366c2843b68fc3ca441446d73d3 # v2.77.1
+        with:
+          tool: cargo-llvm-cov
 
       - name: Cache cargo artifacts
-        uses: Swatinem/rust-cache@v2
+        uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2.9.1
```
brendanbyrne: 🔴 **Critical** — this SHA is **not** v2.9.1.

`e18b497796c12c097a38f9edb9d0641fb99eee32` is a post-release commit dated 2026-03-12 with the message *"update changelog, not sure why npm suddenly does not commit this"* — not the v2.9.1 release. The annotated tag `v2.9.1` actually points to commit `c19371144df3bb44fab255c43d04cbc2ab54d1c4`.

Why it matters: the entire point of SHA-pinning is "this exact, audited code." A reviewer reading the comment trusts `# v2.9.1` to mean "I can audit this against the v2.9.1 release diff" — but the audit wouldn't catch what's actually running. On a PR explicitly motivated by the `tj-actions/changed-files` supply-chain hijack, the failure mode is exactly the one the technique is supposed to prevent: comment claims one thing, SHA delivers another.

Likely cause: `gh api .../git/refs/tags/v2.9.1 --jq .object.sha` returns the *tag-object* SHA for annotated tags, not the commit SHA — so the resolution recipe must dereference one extra level for annotated tags. (`codecov/codecov-action v6.0.0` is also annotated but happened to land on the right commit, masking the bug.)

**Fix:**

```yaml
uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
```

brendanbyrne (commented):
## Verdict: Request changes

**What this PR does:** Replaces every floating `uses:` tag with the commit SHA of the corresponding release, plus adds `.github/dependabot.yml` to keep them current.

**Areas reviewed:** `design.md` (no impact), `.claude/CLAUDE.md` (no convention contradicted). No coding-standards file covers Actions pinning.

### Findings

**🔴 Critical — `Swatinem/rust-cache` SHA does not match v2.9.1.** See inline comment on `coverage.yml` line 34. The other five pins were verified against their release tags (table below).

**🔵 Suggestion** — Dependabot config could add `commit-message: prefix: "ci"` so auto-PRs match your existing `ci:` prefix convention. Pure ergonomics, not a blocker.

**🔵 Suggestion** — Re the rebase-after-#85 plan in your PR description: branching from `main` (rather than stacking on #85's branch) is the right call per the project's squash-merge convention. The conflict is the small price.

### SHA verification

Dereferenced every tag to its commit SHA via the GitHub API. Five of six match; one does not.

| Action | Tag | Pin in PR | Commit SHA at tag | Match |
|---|---|---|---|---|
| actions/checkout | v4.3.1 | `34e1148…` | `34e1148…` | ✅ |
| dtolnay/rust-toolchain | v1 | `e97e2d8…` | `e97e2d8…` | ✅ |
| taiki-e/install-action | v2.77.1 | `cca35ed…` | `cca35ed…` | ✅ |
| **Swatinem/rust-cache** | **v2.9.1** | **`e18b497…`** | **`c193711…`** | **❌** |
| codecov/codecov-action | v6.0.0 | `57e3a13…` | `57e3a13…` | ✅ |
| lycheeverse/lychee-action | v2.8.0 | `8646ba3…` | `8646ba3…` | ✅ |

Reproduce via:

```bash
resolve() {
  local raw=$(gh api "repos/$1/git/refs/tags/$2")
  if [ "$(echo "$raw" | jq -r .object.type)" = "tag" ]; then
    gh api "repos/$1/git/tags/$(echo "$raw" | jq -r .object.sha)" --jq .object.sha
  else
    echo "$raw" | jq -r .object.sha
  fi
}
resolve Swatinem/rust-cache v2.9.1
```

### Design-doc accuracy

No changes required. `design.md` doesn't describe action versions or pin policy at this granularity.

### What's good

- The two non-standard tag patterns (`@stable`, `@cargo-llvm-cov`) handled by switching to versioned tags + explicit input parameters — that's the right move; Dependabot can track these and the behavior is identical.
- Dependabot grouping with `patterns: ["*"]` is the right tradeoff for a hobby-project cadence (six PRs/week → one).
- "Out of scope" section in the PR body prevents future scope creep on this work.
- Test plan handles the same `paths:` filter trap as #85.

brendanbyrne:
Thanks — the Swatinem mismatch was a real bug. Root cause: I dereferenced the `v2` floating annotated tag (`42dc69e1…` → commit `e18b497…`), and the tag's *message* says "2.9.1" so I trusted that as the version label. But Swatinem evidently maintains the `v2` floating tag on a different commit than the `v2.9.1` immutable tag (`23869a5b…` → commit `c193711…`). That's not the convention the other five actions follow — for them, `v<major>` and `v<major>.<minor>.<patch>` resolve to the same commit, so my "deref the floater, label with its message" shortcut happened to work and I generalized from a small N. Fix in 67799a8: pin to v2.9.1's commit (`c193711…`).

Verification table after the fix — every pin now matches the immutable specific-version tag:

| Action | Tag | Pin | Match |
|---|---|---|---|
| actions/checkout | v4.3.1 | `34e1148…` | ✅ |
| dtolnay/rust-toolchain | v1 | `e97e2d8…` | ✅ |
| taiki-e/install-action | v2.77.1 | `cca35ed…` | ✅ |
| Swatinem/rust-cache | v2.9.1 | `c193711…` | ✅ (was `e18b497…`) |
| codecov/codecov-action | v6.0.0 | `57e3a13…` | ✅ |
| lycheeverse/lychee-action | v2.8.0 | `8646ba3…` | ✅ |

Took the `commit-message: prefix: "ci"` suggestion in the same commit. Skipped the rebase-after-#85 note since that was an FYI confirming the plan already in the PR description.

brendanbyrne (commented):
## Re-review: Approve (verdict flipped)

Previous verdict was "Request changes" over the rust-cache SHA mismatch. That's resolved.

### What changed since last review

- **🔴 rust-cache SHA fixed.** [`coverage.yml`:34](.github/workflows/coverage.yml#L34) now pins `c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1`. I re-ran the annotated-tag dereference recipe — that's the actual v2.9.1 commit. ✅
- **🔵 Adopted Dependabot prefix suggestion.** `commit-message: prefix: "ci"` added to `dependabot.yml`. ✅
- All other five pins unchanged from prior review (were already correct).

Backend coverage and link-check both green on the head SHA. No remaining findings.

Merge plan unchanged: merge #85 first, rebase this onto main, merge.

PR #92, opened 2026-05-09
audit: adopt ID newtypes project-wide (§4.2)

### LINKED ISSUE #82: Audit §4.2: revert unused ID newtypes (or adopt project-wide)
## Context

The 2026-04-15 Rust audit ([`docs/designs/2026-04-15-rust-audit.md`](../blob/main/docs/designs/2026-04-15-rust-audit.md) §4.2) recommended introducing newtype wrappers — `UserId`, `SessionId`, `RunId`, `SessionRaceId` — to prevent mixing up IDs at function boundaries. The audit itself flagged the trade-off: "Touches every service function signature. Big diff, modest practical benefit for a small codebase. Worth it if you're worried about ID confusion; skip if you're not." Box checked Approved.

**Current state:** [`backend/src/domain/ids.rs`](../blob/main/backend/src/domain/ids.rs) defines all four newtypes (127 lines, full test coverage — `Deref<Target=str>`, `AsRef<str>`, transparent serde, hash/eq). The module is `pub mod ids;` in `domain/mod.rs`. **Zero callers anywhere in the codebase:** `grep -rn "domain::ids" backend/src/ | grep -v '^backend/src/domain/ids.rs'` returns nothing. Service signatures still take `&str` and `String`.

This is the worst of both worlds: maintenance cost of a tested public type with no benefit to actual code paths.

### Option 1 — revert (recommended)

Delete `backend/src/domain/ids.rs` and remove `pub mod ids;` from `domain/mod.rs`. Document the decision in a one-paragraph ADR (`docs/decisions/NNNN-id-newtypes-not-adopted.md`) so the next person who wonders "why don't we have these?" finds the answer. The audit's own caveat justifies skipping for a small codebase.

Effort: ~30 minutes (delete + mod.rs edit + ADR + history entry).

### Option 2 — adopt project-wide

Plumb `UserId`, `SessionId`, `RunId`, `SessionRaceId` through every service and route signature. `Deref<Target=str>` means most internal call sites compile unchanged once parameters change. Mostly mechanical refactor — large flat diff, no logic changes. Worth doing only if there's a concrete reason (a near-miss bug, a planned feature where ID confusion is likely).

Effort: ~3–4 hours of refactor + test pass + review.

## Implementation plan

**If Option 1:** delete `domain/ids.rs`, edit `domain/mod.rs`, write ADR `docs/decisions/NNNN-id-newtypes-not-adopted.md` (use the template at `docs/decisions/template.md`), update audit doc §4.2 to mark as reverted-with-ADR, append history entries to both audit doc and any other touched file. PR title: `audit: revert unused ID newtypes (§4.2)`.

**If Option 2:** change service/route function signatures across `backend/src/services/*.rs` and `backend/src/routes/*.rs` to use the newtypes; update entity-to-newtype conversions at the DB boundary; run all tests; update audit doc §4.2 with adoption status; PR title: `audit: adopt ID newtypes project-wide (§4.2)`.

## References

- Audit doc: [`docs/designs/2026-04-15-rust-audit.md`](../blob/main/docs/designs/2026-04-15-rust-audit.md) §4.2
- Code: [`backend/src/domain/ids.rs`](../blob/main/backend/src/domain/ids.rs) (the orphan), `domain/mod.rs` (the export)
- ADR template: [`docs/decisions/template.md`](../blob/main/docs/decisions/template.md)
- Reconciliation: `docs/drafts/WIP_audit-reconciliation-2026-05-08.md` (transient)

brendanbyrne:
## Summary

Per audit §4.2 (Approved), the four newtypes in `domain/ids.rs` — `UserId`, `SessionId`, `RunId`, `SessionRaceId` — are now plumbed through the backend instead of sitting unused. Closes #82.

- **Service signatures** take and return the newtypes (`services/{sessions,runs,users,helpers,auth,session_context}.rs`).
- **Middleware** `AuthUser` / `AdminUser` expose `user_id: UserId`. JWT `claims.sub` stays a `String` and is wrapped at extractor construction.
- **Route adapters** wrap `Path<String>` into the appropriate newtype at the boundary. Request body types kept as `String` (untrusted client input).
- **Response DTOs** use newtypes directly. `serde(transparent)` keeps wire format identical — frontend and integration tests unchanged.
- **Entity columns** stay `String` (auto-generated SeaORM); newtype↔String conversion happens at the service↔entity boundary via `Set(id.as_str().to_string())` / `UserId::new(model.user_id)`.
- **Ergonomics:** added `From<&Self>` / `From<Self>` for `sea_orm::Value` and `String` to the newtype macro so `Column::Foo.eq(user_id)` and `find_by_id(session_id)` accept newtype IDs directly without `.as_str()` at every call site.
- **Audit doc** §4.2 updated with adoption status; Document history section added.

## Test plan

Automated:

- [ ] `cd backend && cargo build --all-targets` — clean.
- [ ] `cd backend && cargo clippy --all-targets -- -D warnings` — clean.
- [ ] `cd backend && cargo test` — 235 tests pass (159 lib + 27 + 21 + 8 + 1 + 19 integration).

Manual smoke (the wire format is supposed to be unchanged; this verifies):

1. Start backend: `cd backend && cargo run`. Wait for "Listening on" line.
2. Create a user (note the returned id, treat as opaque): `curl -s -X POST http://127.0.0.1:3000/auth/register -H 'Content-Type: application/json' -d '{"username":"newtype-test","password":"hunter22hunter22"}'` — confirm 201, response body has `{"user":{"id":"<uuid>", ...}, "tokens":{...}}`.
3. Log in: `curl -s -X POST http://127.0.0.1:3000/auth/login -H 'Content-Type: application/json' -d '{"username":"newtype-test","password":"hunter22hunter22"}'` — confirm 200, capture the `tokens.access_token` (export as `TOK`).
4. Create a session: `curl -s -X POST http://127.0.0.1:3000/sessions -H "Authorization: Bearer $TOK" -H 'Content-Type: application/json' -d '{"ruleset":"random"}'` — confirm 201, response has `{"id":"<uuid>", "host_id":"<uuid>", "participants":[{"user_id":"<uuid>", ...}], ...}`. The IDs are bare strings, not nested objects.
5. Hit `/sessions/mine` with the same token: `curl -s http://127.0.0.1:3000/sessions/mine -H "Authorization: Bearer $TOK"` — confirm `{"session_id":"<uuid>"}`.
6. Open the frontend dev server (`cd frontend && bun dev`) and log in as the same user. Confirm session list, session detail, and run-recording flows work end-to-end without surface-visible regressions.

## Notes for reviewer

- The agent flagged one performance call-out: `SessionContext` clones `self.session.id` into a `SessionId` per `require_participant` / `touch` call (single allocation, ~50 bytes per UUID, not in a hot loop). Acceptable today; structural fix would be caching a `SessionId` next to the entity model.
- `Set(id.as_str().to_string())` could be `Set(id.to_string())` (Display gives the same output). Kept the explicit-`.as_str().to_string()` form for grep-ability — it tells the reader "I'm reading the inner string and re-allocating for the entity write."
- This is **Option 2** from #82 (full adoption), not Option 1 (revert).

### DIFF
--- a/backend/src/routes/drink_types.rs
+++ b/backend/src/routes/drink_types.rs
@@ -81,7 +81,7 @@ pub async fn create_drink_type(
         name: Set(name),
         alcoholic: Set(req.alcoholic),
         created_at: Set(now),
-        created_by: Set(Some(user.user_id)),
+        created_by: Set(Some(user.user_id.into_string())),
     };
 
     let inserted = sea_orm::ActiveModelTrait::insert(model, &state.db).await?;
--- a/backend/src/routes/sessions.rs
+++ b/backend/src/routes/sessions.rs
@@ -6,7 +6,13 @@ use axum::{
 };
 use serde::{Deserialize, Serialize};
 
-use crate::{AppState, error::AppError, middleware::auth::AuthUser, services::sessions};
+use crate::{
+    AppState,
+    domain::{SessionId, SessionRaceId},
+    error::AppError,
+    middleware::auth::AuthUser,
+    services::sessions,
+};
 
 #[derive(Deserialize)]
 pub struct CreateSessionRequest {
@@ -25,7 +31,7 @@ pub async fn create_session(
 
 #[derive(Serialize)]
 pub struct MySessionResponse {
-    pub session_id: Option<String>,
+    pub session_id: Option<SessionId>,
 }
 
 /// GET /sessions/mine — get the user's current active session ID.
@@ -52,7 +58,8 @@ pub async fn get_session(
     State(state): State<AppState>,
     Path(id): Path<String>,
 ) -> Result<Json<sessions::SessionDetail>, AppError> {
-    let detail = sessions::get_session_detail(&state.db, &id, Some(&user.user_id)).await?;
+    let session_id = SessionId::new(id);
+    let detail = sessions::get_session_detail(&state.db, &session_id, Some(&user.user_id)).await?;
     Ok(Json(detail))
 }
 
@@ -62,7 +69,8 @@ pub async fn join_session(
     State(state): State<AppState>,
     Path(id): Path<String>,
 ) -> Result<impl IntoResponse, AppError> {
-    sessions::join_session(&state.db, &id, &user.user_id).await?;
+    let session_id = SessionId::new(id);
+    sessions::join_session(&state.db, &session_id, &user.user_id).await?;
     Ok(StatusCode::NO_CONTENT)
 }
 
@@ -72,7 +80,8 @@ pub async fn leave_session(
     State(state): State<AppState>,
     Path(id): Path<String>,
 ) -> Result<impl IntoResponse, AppError> {
-    sessions::leave_session(&state.db, &id, &user.user_id).await?;
+    let session_id = SessionId::new(id);
+    sessions::leave_session(&state.db, &session_id, &user.user_id).await?;
     Ok(StatusCode::NO_CONTENT)
 }
 
@@ -82,7 +91,8 @@ pub async fn next_track(
     State(state): State<AppState>,
     Path(id): Path<String>,
 ) -> Result<(StatusCode, Json<sessions::SessionRaceInfo>), AppError> {
-    let race = sessions::next_track(&state.db, &id, &user.user_id).await?;
+    let session_id = SessionId::new(id);
+    let race = sessions::next_track(&state.db, &session_id, &user.user_id).await?;
     Ok((StatusCode::CREATED, Json(race)))
 }
 
@@ -92,7 +102,8 @@ pub async fn skip_turn(
     State(state): State<AppState>,
     Path(id): Path<String>,
 ) -> Result<(StatusCode, Json<sessions::SessionRaceInfo>), AppError> {
-    let race = sessions::skip_turn(&state.db, &id, &user.user_id).await?;
+    let session_id = SessionId::new(id);
+    let race = sessions::skip_turn(&state.db, &session_id, &user.user_id).await?;
     Ok((StatusCode::CREATED, Json(race)))
 }
 
@@ -103,6 +114,8 @@ pub async fn skip_pending_race(
     State(state): State<AppState>,
     Path((session_id, race_id)): Path<(String, String)>,
 ) -> Result<impl IntoResponse, AppError> {
+    let session_id = SessionId::new(session_id);
+    let race_id = SessionRaceId::new(race_id);
     sessions::skip_pending_race(&state.db, &session_id, &race_id, &user.user_id).await?;
     Ok(StatusCode::NO_CONTENT)
 }
@@ -113,6 +126,7 @@ pub async fn list_races(
     State(state): State<AppState>,
     Path(id): Path<String>,
 ) -> Result<Json<Vec<sessions::RaceInfo>>, AppError> {
-    let races = sessions::list_races(&state.db, &id).await?;
+    let session_id = SessionId::new(id);
+    let races = sessions::list_races(&state.db, &session_id).await?;
     Ok(Json(races))
 }
--- a/backend/src/services/session_context.rs
+++ b/backend/src/services/session_context.rs
@@ -13,9 +13,14 @@ use crate::{
     services::helpers,
 };
 
+/// `session_id` mirrors `session.id` as a typed `SessionId` so the helper
+/// methods can borrow `&self.session_id` instead of allocating a fresh wrapper
+/// on every call. Both fields hold the same UUID; the typed copy is the one
+/// callers borrow when they need a `&SessionId`.
 #[derive(Debug, Clone)]
 pub struct SessionContext {
     pub session: sessions::Model,
+    pub session_id: SessionId,
 }
 
 impl SessionContext {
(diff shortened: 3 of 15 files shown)


backend/src/routes/drink_types.rs
```diff
@@ -81,7 +81,7 @@ pub async fn create_drink_type(
         name: Set(name),
         alcoholic: Set(req.alcoholic),
         created_at: Set(now),
-        created_by: Set(Some(user.user_id)),
+        created_by: Set(Some(user.user_id.into_string())),
```
brendanbyrne, line 84: 🔵 **Suggestion** — Style inconsistency. Everywhere else this PR uses `Set(id.as_str().to_string())` (the PR description justifies it for grep-ability — "I'm reading the inner string and re-allocating for the entity write"), but here it's `into_string()`. `into_string()` is actually the better default: it consumes the newtype and avoids the `as_str + to_string` re-allocation. Either align on `into_string()` everywhere a `Set(...)` consumes the value, or document why `drink_types.rs` is the exception.

Not a blocker — just two patterns drifting in one PR.

backend/src/services/session_context.rs
```diff
@@ -39,17 +40,19 @@ impl SessionContext {
     pub async fn require_participant<C: ConnectionTrait>(
         &self,
         db: &C,
-        user_id: &str,
+        user_id: &UserId,
```
brendanbyrne, line 52: 🔵 **Suggestion** — Per-call clone of `self.session.id`. The PR description acknowledges this: every `require_participant` / `touch` call clones the inner string and rewraps it as a `SessionId`. Not a hot path, but the structural fix is one extra field:

```rust
pub struct SessionContext {
    session: sessions::Model,
    session_id: SessionId,  // wrap once at load_active
}
```

…then `require_participant` / `touch` borrow `&self.session_id` instead of cloning + wrapping. Same idea for the `touch` impl a few lines below. Optional — fine to defer if you'd rather keep this PR strictly mechanical.

backend/src/routes/sessions.rs
```diff
@@ -52,7 +58,8 @@ pub async fn get_session(
     State(state): State<AppState>,
     Path(id): Path<String>,
 ) -> Result<Json<sessions::SessionDetail>, AppError> {
-    let detail = sessions::get_session_detail(&state.db, &id, Some(&user.user_id)).await?;
+    let session_id = SessionId::new(id);
```
brendanbyrne, line 61: 🔵 **Suggestion** — `Path<SessionId>` would work directly. Every handler in this file does `Path<String>` then `let session_id = SessionId::new(id);`. Since `SessionId` derives `Deserialize` and is `#[serde(transparent)]`, the Axum `Path` extractor can deserialize straight into the newtype:

```rust
Path(session_id): Path<SessionId>,
```

That removes the wrap line in every handler (`get_session`, `join_session`, `leave_session`, `next_track`, `skip_turn`, `list_races`) and the `(SessionId, SessionRaceId)` tuple in `skip_pending_race`.

Fine to defer — the PR description notes a deliberate choice to wrap at the boundary on the body-DTO side because of "untrusted client input," but path params are equally untrusted, so the consistency argument cuts the other way. Up to you.

brendanbyrne (commented):
## Verdict

Approve with minor suggestions. Mechanical, well-scoped adoption of the four ID newtypes per audit §4.2 / Issue #82. Wire format preserved via `serde(transparent)`; no `design.md` / `api-contract.md` impact.

## What this PR does

Plumbs `UserId`, `SessionId`, `RunId`, `SessionRaceId` through service signatures, middleware (`AuthUser` / `AdminUser`), route adapters, response DTOs, and test helpers. Adds four `From` impls to the `string_id_newtype!` macro — `From<&Self>` / `From<Self>` for `sea_orm::Value` and `String` — so SeaORM filters and `find_by_id` accept newtypes directly without `.as_str()` boilerplate.

## What was verified

- `cargo fmt --check` — clean (only the standard nightly-feature warnings about `imports_granularity` / `group_imports`, pre-existing).
- `cargo clippy --all-targets -- -D warnings` — clean.
- `cargo test --lib` — 158 lib tests pass.
- Audit doc §4.2 adoption status + dated history entry per `docs/CLAUDE.md` Document history rule.
- `design.md` / `data-model.md` / `api-contract.md` correctly untouched — wire format is preserved by `serde(transparent)` on the existing newtypes.
- New `into_sea_orm_value` macro test covers both ref and owned conversions.

## What's good

- The four new macro `From` impls give exactly the right ergonomic surface: borrowed (clone path, for filter args you don't own) and owned (move path, for `Set(...)`).
- `services/sessions.rs::leave_session` refactor extracts `is_host_leaving` *before* consuming `session` into `ActiveModel`, eliminating a `session.clone()`. Nice.
- Test helpers updated to return newtypes (`create_user → UserId`, `insert_session → SessionId`, etc.) so call sites stay clean instead of growing `.as_str()` calls.

Three suggestions inline below — none block merge.

backend/src/routes/drink_types.rs
brendanbyrne: align on `into_string()` to minimize the number of unnecessary allocations, or tell me why it isn't possible.

backend/src/routes/sessions.rs
brendanbyrne: do it

backend/src/services/session_context.rs
brendanbyrne: Done in 2ca63b3 — added a `session_id: SessionId` field, wrapped once in `load_active`, and `require_participant` / `touch` now borrow `&self.session_id`.

Issue #98, opened 2026-05-09
Lint cleanup (e): nursery warnings


brendanbyrne:
## Scope
Remove module-level `#[allow(clippy::...)]`s for `clippy::nursery` warnings introduced by PR-A1 (#24).

Nursery lints are experimental — some are noisy or disagree with idiomatic style. For each warning:
- If the lint catches a real issue, fix the code.
- If the lint is wrong about this codebase's style, keep an allow but move it from module-level to a single `#[allow]` on the specific item, with a one-line `// reason: ...` comment justifying it.

## Verification
Workspace-level `clippy::nursery = warn` is enforced everywhere; no broad module-level nursery allows remain. Any per-item allows carry a `// reason:` comment.

PR #112, opened 2026-05-11
103: clear pedantic batch 2 (module_name_repetitions + cast_*)

### LINKED ISSUE #103: Lint cleanup (d): pedantic style lints, batch 2
## Scope
Remove the second batch of `clippy::pedantic` warnings. Concrete list determined when (c) closes — what's left after (c) plus any new pedantic warnings introduced by intervening PRs.

Likely candidates:
- `clippy::if_not_else`, `clippy::match_bool`, `clippy::single_match_else` — readability rewrites.
- `clippy::redundant_closure_for_method_calls` — `.map(|x| x.foo())` → `.map(Type::foo)`.
- `clippy::option_if_let_else`, `clippy::map_unwrap_or` — combinator preference.

Goal: wrap pedantic in two batches; if more remain after (d), open additional sub-issues.

## Verification
Workspace-level `clippy::pedantic = warn` is enforced everywhere; no module-level pedantic allows remain.

brendanbyrne:
Closes #103. Fourth PR in the PR-H1+ lint cleanup sequence.

## Summary

Removes **five** workspace allows in `backend/Cargo.toml` (`module_name_repetitions` + the four `cast_*` lints) by renaming the 13 module-name-repetition sites and fixing the 6 cast sites. After this PR, the only pedantic lints still allow-listed are `missing_errors_doc` / `missing_panics_doc` (queued for PR-G3, doc-comment audit).

## Renames (13 sites)

| Old | New |
|---|---|
| `AppConfig` | `config::Config` |
| `AppError` | `error::Error` |
| `AuthUser` | `middleware::auth::User` |
| `AuthResponse` | `routes::auth::Response` |
| `DrinkTypesQuery` | `routes::drink_types::Filters` *(avoids collision with `axum::extract::Query`)* |
| `RaceSetupUpdate` | `domain::race_setup::Update` |
| `list_drink_types` / `list_runs` / `list_sessions` / `list_users` | `list` |
| `list_active_sessions` | `list_active` |
| `close_stale_sessions` | `close_stale` |

All renames are mechanical — `sed -i 's/\bOLD\b/NEW/g'` across `backend/**/*.rs`, with `DrinkTypesQuery → Filters` (not `Query`) to dodge the `axum::extract::Query` collision the obvious rename would have caused.

`AppState` (in `lib.rs`, not in an `app` module) was **not** flagged by clippy and is left as-is.

## Cast cleanups (6 sites)

| Site | Before | After |
|---|---|---|
| `services/auth.rs:98, 122` | `config.jwt_*_expiry_minutes as i64` | field type changed `u64 → i64` |
| `routes/auth.rs:59` | `config.jwt_refresh_expiry_days as i64 * 86400` | cast removed (field is now `i64`) |
| `services/sessions.rs:680` | `r.race_number as usize` | `usize::try_from(r.race_number).unwrap_or(0)` |
| `services/sessions.rs:923` | `used_races.len() as i32` | `i32::try_from(used_races.len()).unwrap_or(i32::MAX)` |
| `test_helpers.rs:73` | `(i + 1) as i32` | `i32::try_from(i + 1).expect("test cup index fits in i32")` (test code; `expect_used` allowed) |

Field-type change rationale: `chrono::TimeDelta::{minutes,days}` take `i64`. The fields were `u64` to express non-negativity, but the cast was happening at every call site. Storing them as `i64` removes the cast.

## Docs updated

Live prose only — historical PR notes preserve the old names:

- `backend/CLAUDE.md` § Errors
- `docs/design.md` § Observability → Error response pattern + Document history entry
- `docs/api-contract.md` § 3 (error response contract example) + Document history entry

## Workspace `Cargo.toml` after this PR

Remaining pedantic allows:
- `missing_errors_doc`, `missing_panics_doc` — queued for **PR-G3** (doc-comment audit).
- `cargo_common_metadata`, `multiple_crate_versions` — leave alone (publishing concerns).
- `doc_markdown`, `too_long_first_doc_paragraph` — queued for **PR-G3**.
- `needless_raw_string_hashes`, `too_many_lines` — leave alone (style).

That closes the H1+ a/b/c/d sequence per the handoff. **#98 (e — nursery)** remains.

## Verification

```
cd backend
cargo clippy --all-targets   # clean
cargo test                    # 164 unit + 27 + 21 + 8 + 1 + 19 integration tests pass
```

Manual reproduction:

1. **The cleared lints stay clean.** From `backend/`:
   ```
   cargo clippy --all-targets -- \
     -W clippy::module_name_repetitions \
     -W clippy::cast_possible_wrap \
     -W clippy::cast_possible_truncation \
     -W clippy::cast_sign_loss \
     -W clippy::cast_precision_loss
   ```
   Should be silent.

2. **The renamed types still resolve.** `cargo doc --no-deps` produces module pages under `config::Config`, `error::Error`, `middleware::auth::User`, etc. (No `AppError` page anywhere in the generated docs.)

3. **Auth still works end-to-end.** Run `cargo test --test auth_integration` — 8 tests covering register / login / refresh / logout / change_password / cookie handling.

4. **Session lifecycle still works.** `cargo test --test session_integration` — 19 tests, including `test_stale_session_cleanup` which exercises the renamed `close_stale` path.

### DIFF
--- a/backend/src/services/sessions.rs
+++ b/backend/src/services/sessions.rs
@@ -670,9 +670,15 @@ pub async fn get_session_detail(
     // 1-indexed and gapless: `next_track` appends monotonically, and
     // `skip_turn` replaces in-place (preserves race_number). No deletion
     // path exists. Under this invariant, last().race_number == COUNT(*).
-    let race_number = races
-        .last()
-        .map_or(1, |r| usize::try_from(r.race_number).unwrap_or(0));
+    let race_number = match races.last() {
+        None => 1,
+        Some(r) => usize::try_from(r.race_number).map_err(|_| {
+            Error::Internal(anyhow::anyhow!(
+                "race_number invariant violated: got {}",
+                r.race_number
+            ))
+        })?,
+    };
 
     Ok(SessionDetail {
         id: SessionId::new(session.id),
@@ -915,7 +921,12 @@ pub async fn next_track(
         .filter(session_races::Column::SessionId.eq(session_id))
         .all(db)
         .await?;
-    let race_count = i32::try_from(used_races.len()).unwrap_or(i32::MAX);
+    let race_count = i32::try_from(used_races.len()).map_err(|_| {
+        Error::Internal(anyhow::anyhow!(
+            "session has more races than i32 can represent: {}",
+            used_races.len()
+        ))
+    })?;
     let used_track_ids: Vec<i32> = used_races.iter().map(|r| r.track_id).collect();
 
     let chosen = helpers::pick_random_track(db, &used_track_ids, &[]).await?;
--- a/backend/src/services/sessions.rs
+++ b/backend/src/services/sessions.rs
@@ -15,7 +15,7 @@ use crate::{
         cups, runs, session_participants, session_race_participations, session_races, sessions,
         users,
     },
-    error::AppError,
+    error::Error,
     services::helpers,
 };
 
@@ -31,7 +31,7 @@ struct ActiveParticipantRow {
 async fn check_not_in_any_session(
     db: &impl ConnectionTrait,
     user_id: &UserId,
-) -> Result<(), AppError> {
+) -> Result<(), Error> {
     let existing =
         ActiveParticipantRow::find_by_statement(sea_orm::Statement::from_sql_and_values(
             db.get_database_backend(),
@@ -50,7 +50,7 @@ async fn check_not_in_any_session(
         .await?;
 
     if let Some(row) = existing {
-        return Err(AppError::Conflict(format!(
+        return Err(Error::Conflict(format!(
             "Already in session {}",
             row.session_id
         )));
@@ -64,7 +64,7 @@ async fn check_not_in_any_session(
 pub async fn get_active_session_id(
     db: &DatabaseConnection,
     user_id: &UserId,
-) -> Result<Option<SessionId>, AppError> {
+) -> Result<Option<SessionId>, Error> {
     let row = ActiveParticipantRow::find_by_statement(sea_orm::Statement::from_sql_and_values(
         db.get_database_backend(),
         r#"
@@ -90,7 +90,7 @@ pub async fn create_session(
     db: &DatabaseConnection,
     user_id: &UserId,
     ruleset: &str,
-) -> Result<SessionDetail, AppError> {
+) -> Result<SessionDetail, Error> {
     let parsed: Ruleset = ruleset.parse()?;
 
     check_not_in_any_session(db, user_id).await?;
@@ -151,9 +151,7 @@ struct SessionSummaryRow {
 
 /// List active sessions sorted by last_activity_at DESC.
 /// Uses a single JOIN query instead of N+1 queries.
-pub async fn list_active_sessions(
-    db: &DatabaseConnection,
-) -> Result<Vec<SessionSummary>, AppError> {
+pub async fn list_active_sessions(db: &DatabaseConnection) -> Result<Vec<SessionSummary>, Error> {
     let rows = SessionSummaryRow::find_by_statement(sea_orm::Statement::from_sql_and_values(
         db.get_database_backend(),
         r#"
@@ -300,16 +298,13 @@ pub struct SessionDetail {
 
 /// Look up the host's username. Returns `Internal` if the FK-referenced
 /// user row is missing (data corruption — FKs should prevent this).
-async fn load_host_username(
-    db: &impl ConnectionTrait,
-    host_id: &UserId,
-) -> Result<String, AppError> {
+async fn load_host_username(db: &impl ConnectionTrait, host_id: &UserId) -> Result<String, Error> {
     users::Entity::find_by_id(host_id)
         .one(db)
         .await?
         .map(|u| u.username)
         .ok_or_else(|| {
-            AppError::Internal(anyhow::anyhow!("Host user not found for host_id {host_id}"))
+            Error::Internal(anyhow::anyhow!("Host user not found for host_id {host_id}"))
         })
 }
 
@@ -317,7 +312,7 @@ async fn load_host_username(
 async fn load_participants(
     db: &impl ConnectionTrait,
     session_id: &SessionId,
-) -> Result<Vec<ParticipantInfo>, AppError> {
+) -> Result<Vec<ParticipantInfo>, Error> {
     let rows = ParticipantRow::find_by_statement(sea_orm::Statement::from_sql_and_values(
         db.get_database_backend(),
         r#"
@@ -348,7 +343,7 @@ async fn load_participants(
 async fn load_current_race_with_submissions(
     db: &impl ConnectionTrait,
     session_id: &SessionId,
-) -> Result<Option<SessionRaceInfo>, AppError> {
+) -> Result<Option<SessionRaceInfo>, Error> {
     let race_row = CurrentRaceRow::find_by_statement(sea_orm::Statement::from_sql_and_values(
         db.get_database_backend(),
         r#"
@@ -409,7 +404,7 @@ async fn load_current_race_with_submissions(
 async fn load_race_history(
     db: &impl ConnectionTrait,
     session_id: &SessionId,
-) -> Result<Vec<RaceInfo>, AppError> {
+) -> Result<Vec<RaceInfo>, Error> {
     let rows = RaceHistoryRow::find_by_statement(sea_orm::Statement::from_sql_and_values(
         db.get_database_backend(),
         r#"
@@ -491,7 +486,7 @@ pub async fn get_pending_races(
     db: &impl ConnectionTrait,
     session_id: &SessionId,
     user_id: &UserId,
-) -> Result<Vec<SessionRaceInfo>, AppError> {
+) -> Result<Vec<SessionRaceInfo>, Error> {
     let now = Utc::now().naive_utc();
     let grace_cutoff = now - chrono::Duration::minutes(REJOIN_GRACE_MINUTES);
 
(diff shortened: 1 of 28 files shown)


backend/src/services/sessions.rs
```diff
@@ -677,7 +670,9 @@ pub async fn get_session_detail(
     // 1-indexed and gapless: `next_track` appends monotonically, and
     // `skip_turn` replaces in-place (preserves race_number). No deletion
     // path exists. Under this invariant, last().race_number == COUNT(*).
-    let race_number = races.last().map_or(1, |r| r.race_number as usize);
+    let race_number = races
+        .last()
+        .map_or(1, |r| usize::try_from(r.race_number).unwrap_or(0));
```
brendanbyrne: 🟡 **Important** — `unwrap_or(0)` silently masks the invariant the comment above documents ("1-indexed and gapless"). If `r.race_number` is ever negative or otherwise unrepresentable as `usize`, this serializes `race_number: 0` to the client — a value that is **not valid** under the documented invariant — and proceeds as if nothing happened.

Per `rust.md` § 1 (use `?` and `From` for discriminable failures; don't collapse signal into a fallback) and § 11 (handlers/services: don't silently consume errors), an invariant violation in service code should propagate as `Error::Internal`:

```rust
let race_number = match races.last() {
    None => 1,
    Some(r) => usize::try_from(r.race_number).map_err(|_| {
        Error::Internal(anyhow::anyhow!(
            "race_number invariant violated: got {}",
            r.race_number
        ))
    })?,
};
```

This keeps the cast lint clean **and** surfaces invariant violations as a 500 with a debuggable error chain instead of corrupt response data.

backend/src/services/sessions.rs
```diff
@@ -920,7 +915,7 @@ pub async fn next_track(
         .filter(session_races::Column::SessionId.eq(session_id))
         .all(db)
         .await?;
-    let race_count = used_races.len() as i32;
+    let race_count = i32::try_from(used_races.len()).unwrap_or(i32::MAX);
```
brendanbyrne: 🟡 **Important** — `unwrap_or(i32::MAX)` is subtly worse than a panic here, because `race_count` is consumed at L925 as `let new_race_number = race_count + 1;`. If the fallback ever fires (>2³¹ rows for one session — DB corruption territory, not a real session), `i32::MAX + 1` overflows: debug panics, release wraps to `i32::MIN`, which then becomes the `race_number` column value persisted on the new `session_races` row.

Per `rust.md` § 1, this should propagate:

```rust
let race_count = i32::try_from(used_races.len()).map_err(|_| {
    Error::Internal(anyhow::anyhow!(
        "session has more races than i32 can represent: {}",
        used_races.len()
    ))
})?;
```

For contrast, the third converted site — [`test_helpers.rs:73`](backend/src/test_helpers.rs#L73) — uses `.expect("test cup index fits in i32")`, which is the correct shape for tests per § 11. Production code shouldn't borrow the silent-fallback shape from the cast lint clearance.

backend/src/services/users.rs
```diff
@@ -64,22 +64,22 @@ pub async fn update_profile(
...
         .await?
-        .ok_or_else(|| AppError::NotFound(format!("User {target_user_id} not found")))?;
+        .ok_or_else(|| Error::NotFound(format!("User {target_user_id} not found")))?;
 
     let mut active: users::ActiveModel = user.into();
 
-    // Race setup: all-or-nothing via RaceSetupUpdate
-    if let Some(setup) = RaceSetupUpdate::try_from_optional(
+    // Race setup: all-or-nothing via Update
+    if let Some(setup) = Update::try_from_optional(
```
brendanbyrne: 🔵 **Suggestion** — Renaming `RaceSetupUpdate` → `race_setup::Update` solves the `module_name_repetitions` lint, but the import strips back to bare `Update`:

```rust
use crate::domain::{UserId, race_setup::Update};
…
if let Some(setup) = Update::try_from_optional(…)
```

A reader scanning this file sees `Update::try_from_optional` with no hint that it's race-setup-specific — exactly the readability cost the old name was paying for. Consider importing the module rather than the type:

```rust
use crate::domain::{UserId, race_setup};
…
if let Some(setup) = race_setup::Update::try_from_optional(…)
```

Same lint-clean result, call site stays self-explanatory. Only one consumer of `race_setup::Update` exists today, so the change is local. The sibling `Filters` rename in `routes/drink_types.rs` doesn't have this problem because it's used inside its defining module.

brendanbyrne (commented):
## Code review — PR #112

**Verdict:** Request Changes (one Important pattern worth resolving; otherwise high-quality cleanup).

**What this PR does:** Final installment of the PR-H1+ pedantic cleanup. Removes 5 workspace `allow`s (`module_name_repetitions`, `cast_precision_loss`, `cast_possible_truncation`, `cast_possible_wrap`, `cast_sign_loss`) by renaming 13 type/function sites and rewriting 6 cast sites. Mostly mechanical, with one judgment call on cast fallbacks.

**Areas covered:** `rust.md` (§ 1, § 2, § 8, § 11), `design.md`, `api-contract.md`, `backend/CLAUDE.md`, `compliance-plan.md` (Stream H).

**Automated checks (all pass):**
- `cargo fmt --check` — clean
- `cargo test` — exit 0

**Milestone context:** PR-H1+ (d) per compliance-plan.md § Stream H. The remaining pedantic allows (`missing_errors_doc`, `missing_panics_doc`) are deferred to PR-G3 (doc-comment audit), and PR-H1+ (e — nursery) is the last cleanup PR.

---

## Findings

The two 🟡 Important findings are inline on `backend/src/services/sessions.rs:675` and `:918`. One 🔵 Suggestion is inline on `backend/src/services/users.rs:82`. One additional 🔵 Suggestion below isn't line-anchorable cleanly (spans two fields and a loader function).

### 🔵 Suggestion — `jwt_*_expiry` field-type change weakens the "non-negative duration" invariant

**Files:** `backend/src/config.rs:15` (`jwt_access_expiry_minutes: i64`), `:20` (`jwt_refresh_expiry_days: i64`), and `:36–44` (`from_env`).

The PR description explains the choice: `u64` → `i64` removes the per-call-site cast at `TimeDelta::minutes`/`days`. That's a real cost reduction. But the type now admits negative values that the previous `u64` excluded, and `from_env` doesn't re-check the sign:

```rust
let jwt_access_expiry_minutes = std::env::var("JWT_ACCESS_EXPIRY_MINUTES")
    .ok().and_then(|v| v.parse().ok()).unwrap_or(15);
```

A misconfigured `JWT_ACCESS_EXPIRY_MINUTES=-5` now silently issues already-expired tokens. Pre-PR, `u64::parse("-5")` would fail and fall through to the `15` default.

Per `rust.md` § 2 ("parse, don't validate" — types should encode invariants), the structurally-correct version is a newtype (`nutype` with `validate(greater_or_equal = 0)`) or a checked parse in `from_env`:

```rust
let jwt_access_expiry_minutes: i64 = std::env::var("JWT_ACCESS_EXPIRY_MINUTES")
    .ok().and_then(|v| v.parse::<i64>().ok()).filter(|&v| v > 0).unwrap_or(15);
```

Not blocking — value validation isn't the project's prevailing pattern and this is acknowledged prelaunch ops-discipline territory. Worth a one-line ack if not a code change.

### Design-doc accuracy: ✅ updated, verified accurate

- `docs/design.md` § Observability → Error response pattern: prose, variants-table heading, and bullet text all updated to `error::Error`. Document history entry (2026-05-10) covers every rename comprehensively (`AppError`, `AppConfig`, `AuthUser`, `AuthResponse`, `RaceSetupUpdate`, `list_<resource>`).
- `docs/api-contract.md` § 3: error-response-contract example updated; doc history entry added.
- `backend/CLAUDE.md` § Errors: prose updated.

No stale references in live prose.

## What's good

- **Rename-collision avoidance was thoughtful.** `DrinkTypesQuery → Filters` (not the obvious `Query`) explicitly dodges the `axum::extract::Query` collision. Calling that out in the PR description made review faster.
- **Cast cleanups generally chose the right tier.** Changing the `config.rs` field types (vs. casting at every call site) fixes the issue at the source, once.
- **Doc updates were comprehensive.** Five rename categories × four doc files, all consistent.
- **Workspace `Cargo.toml` ends in a defensible state** — only `missing_*_doc` and the cargo-group / style-only lints remain allowed, each with a written reason and a pointer to the PR that will clear it.
- **Test code stayed honest.** [`test_helpers.rs:73`](backend/src/test_helpers.rs#L73) uses `.expect("test cup index fits in i32")` — correct for tests per `rust.md` § 11, rather than borrowing the production `unwrap_or` pattern.

---

🤖 Posted by `/code-review`. Verdict is informational — only Brendan approves.

brendanbyrne:
Addressed all three findings in commit 3033f7c:

**🟡 services/sessions.rs:675** — `race_number` now propagates `Error::Internal` via `match races.last() { None => 1, Some(r) => usize::try_from(r.race_number).map_err(|_| Error::Internal(anyhow!("race_number invariant violated: got {}", r.race_number)))? }`. Invariant violations surface as a 500 with the offending value in the chain instead of silently serializing `race_number: 0`.

**🟡 services/sessions.rs:918** — `race_count` does the same: `i32::try_from(...).map_err(|_| Error::Internal(anyhow!("session has more races than i32 can represent: {}", used_races.len())))?`. The `race_count + 1` overflow risk is gone since the `Err` branch propagates before that line.

**🔵 services/users.rs:82** — Switched the import from `domain::race_setup::Update` to `domain::race_setup`; call site now reads `race_setup::Update::try_from_optional(...)`, preserving the context the old `RaceSetupUpdate` name carried.

**🔵 config.rs `jwt_*_expiry`** — Added the suggested `.filter(|&v| v > 0)` guard in `from_env` plus a comment explaining why. A misconfigured `JWT_ACCESS_EXPIRY_MINUTES=-5` now falls back to the 15-minute default instead of silently minting already-expired tokens. The earlier `u64` non-negativity invariant is restored.

`cargo clippy --all-targets` clean; all test suites (164 + 27 + 21 + 8 + 1 + 19) still pass.

Issue #70, opened 2026-05-06
User profile endpoints


brendanbyrne:
## Context

Users need endpoints to view and update their profile, including preferred race setup and preferred drink type. The frontend defaulting fallback chain that consumes these endpoints lives in a separate Issue.

## Acceptance criteria

- [ ] `GET /users` lists all users (public profiles): id, username, optional email (if present).
- [ ] `GET /users/:id` returns full user profile: id, username, email, preferred race setup (4 IDs), preferred drink type ID, created_at.
- [ ] `PUT /users/:id` updates profile (self only): username, email, preferred race setup, preferred drink type; returns 403 for non-owner.
- [ ] Tests added: integration tests for list, get, and put (happy-path, 403 for non-owner); unit test for self-only access check.

## References

- Data model: `docs/data-model.md` § Users.
- Design: `docs/design.md` § Workflow 1 (Onboarding), Workflow 6 (Personal Stats).

---
_Filed by Cowork on Brendan's behalf._

brendanbyrne:
Closing as completed — `list_users`, `get_user`, and `update_user` handlers all live in [`backend/src/routes/users.rs`](https://github.com/brendanbyrne/beerio-kart/blob/main/backend/src/routes/users.rs).

- `GET /users` returns the `UserPublicProfile` shape (id, username, preferred-setup IDs, preferred-drink ID, `created_at`) — email and `password_hash` deliberately excluded.
- `GET /users/:id` returns the same shape for a specific user.
- `PUT /users/:id` is gated by self-only access in the service layer; non-owner attempts return 403.

Closed during 2026-05-16 open-ticket audit (Cowork on Brendan's behalf).

PR #115, opened 2026-05-11
84: split Conflict/BadRequest into client/detail; stop driver-string leakage

### LINKED ISSUE #84: Sanitize driver-string leakage from `From<DbErr>` (Conflict/BadRequest paths)
## Status as of 2026-05-11

This Issue was filed 2026-05-08, before PR-C1 / PR-C2 / PR-H1+ landed. Bringing it forward to today:

- The error type was renamed `AppError` → `error::Error` in PR #112. All references below use the current name.
- The `From<sea_orm::DbErr>` impl now sits at `backend/src/error.rs` line 100 (the leaky branches are at lines 106–107).
- This and Issue #114 (PR-G3 doc-comment audit) are the only two open Iter 1 items. Closing both ends Iter 1 cleanly before Star kickoff.
- Quality Pass intent: PR-C1 (#105) and PR-C2 (#107) tightened `Internal` to never leak; `IntoResponse` substitutes `"Internal server error"` for 500s. The same hygiene is missing on 409/400 paths because their messages flow straight through to the client. This Issue closes the gap.
- **Direction confirmed 2026-05-11: Option 3** — structured `Conflict { client, detail }` / `BadRequest { client, detail }`. Reasoning, alternatives considered, and concrete shape are in the Chosen direction section below.

## Context

PR-B1 (#25) implemented `seaorm.md` § 7 literally:

```rust
Some(SqlErr::UniqueConstraintViolation(m)) => Self::Conflict(m),
Some(SqlErr::ForeignKeyConstraintViolation(m)) => Self::BadRequest(m),
```

The `m` here is the raw backend error message — e.g., `"UNIQUE constraint failed: users.username"` from SQLite. That gets returned to API clients via `error::Error::IntoResponse` (`error.rs` lines 61–82), which leaks schema details (table and column names) into the response body. This contradicts the existing convention that internal-detail messages are log-only and sanitized.

Reviewer flagged this on PR-B1 at score 75 (just under the 80 blocker threshold). Recorded as a follow-up rather than blocking. Lived in `.claude/claude-code-notes.md` § "Future PR" until extracted here as part of [#79](https://github.com/brendanbyrne/beerio-kart/issues/79)'s reset-the-notes pass.

## UNIQUE-constraint leak inventory (researched 2026-05-11)

| Constraint | Pre-check today? | User-reachable today? |
|---|---|---|
| `bodies.name` / `wheels.name` / `gliders.name` / `cups.name` / `tracks.name` | n/a (seeded static data) | No |
| `users.username` | Yes (`routes/auth.rs:111`) | TOCTOU only |
| `users.email` | **No** | **No** — email is `Set(None)` everywhere; no code path sets a real value yet |
| `session_participants (session_id, user_id)` | Likely (join flow) | Yes via race |
| `session_participants.user_id` partial unique | Likely | Yes via race (joining a second session while still in another) |
| `session_races (session_id, race_number)` | System-set | Concurrent `next_track` (already tracked as #75) |
| `runs (session_race_id, user_id)` | Yes (`"Already submitted"`) | Race only |

The live leak surface today is small. The `users.email` gap is forward-looking — it'll fire the first time a profile-update / register-with-email endpoint sets a real email value. Option 3 closes the gap structurally so future code can't silently regress when a pre-check is missed.

## Options considered

1. **Sanitize at the `From` impl with generic strings.** `Conflict("Resource already exists".into())`. Simple, two-line change. Loses the driver detail entirely — not even available for logs. Rejected: throws away debuggable info that has value on the log side.
2. **Parse the column out and substitute a structured message.** Couples error layer to schema; brittle on rename. Rejected: pre-checks at the service layer get the same outcome (specific user-facing message) without the coupling.
3. **✅ Preserve specificity in `error::Error`, sanitize in `IntoResponse`.** Structured `Conflict { client, detail }`. Cleanest semantic split: client gets the sanitized text, logs get the raw driver detail via `tracing::warn!` at the boundary. **Chosen.**
4. **Accept the leak; rely on service-layer pre-checks alone.** Leaves the From<DbErr> path leaky as a footgun for future code. Rejected: defense in depth is cheap here.

### Variant shape

```rust
#[error("{client}")]
Conflict {
    client: String,
    /// Internal-only detail (e.g., the raw DB driver message). Logged
    /// at the IntoResponse boundary; not returned to the client.
    detail: Option<String>,
},
#[error("{client}")]
BadRequest {
    client: String,
    detail: Option<String>,
},
```

### Ergonomic helpers

Add associated functions on `Error` so the common case stays terse:

```rust
impl Error {
    pub fn conflict(msg: impl Into<String>) -> Self {
        Self::Conflict { client: msg.into(), detail: None }
    }
    pub fn bad_request(msg: impl Into<String>) -> Self {
        Self::BadRequest { client: msg.into(), detail: None }
    }
}
```

Existing sites migrate from `Error::Conflict("Username already taken".into())` to `Error::conflict("Username already taken")` — net shorter than today. Sites that want to attach detail use the struct-literal form: `Error::Conflict { client: "...".into(), detail: Some(d) }`.

### `From<sea_orm::DbErr>`

```rust
Some(SqlErr::UniqueConstraintViolation(m)) => Self::Conflict {
    client: "Resource already exists".into(),
    detail: Some(m),
},
Some(SqlErr::ForeignKeyConstraintViolation(m)) => Self::BadRequest {
    client: "Referenced record does not exist".into(),
    detail: Some(m),
},
```

### `IntoResponse`

At the response boundary, log `detail` when present, return only `client`:

```rust
Self::Conflict { client, detail } => {
    if let Some(d) = detail {
        tracing::warn!(detail = %d, "Conflict: {client}");
    }
    (StatusCode::CONFLICT, client.clone())
}
// same shape for BadRequest with warn! and StatusCode::BAD_REQUEST
```

## Call-site migration scope

Approximately 20 existing construction sites of `Error::Conflict(...)` and `Error::BadRequest(...)` need updating. Almost all migrate via a search-replace from `Error::Conflict("X".into())` → `Error::conflict("X")` (and `"X".to_string()` → `"X"`). The format-string sites (`Error::Conflict(format!("..."))`) become `Error::conflict(format!("..."))` similarly.

The sites are in:
- `services/sessions.rs` (— 8 sites), `services/runs.rs` (— 6), `services/helpers.rs`, `services/users.rs` (tests), `routes/auth.rs`, `routes/drink_types.rs`, `domain/race_setup.rs`, `domain/enums.rs`. Grep `Error::Conflict\|Error::BadRequest` for the live count.

## Acceptance criteria

- [ ] `error::Error::Conflict` and `error::Error::BadRequest` restructured to struct variants `{ client: String, detail: Option<String> }` per the shape above.
- [ ] `Error::conflict()` and `Error::bad_request()` associated functions added for the no-detail common case.
- [ ] `From<sea_orm::DbErr>` impl in `backend/src/error.rs` updated so `Conflict` and `BadRequest` arms carry the sanitized client message and the raw driver string as `detail`.
- [ ] `IntoResponse` impl emits `tracing::warn!` with `detail` when present; only `client` reaches the response body.
- [ ] Tests added that assert the response body contains the sanitized text and does NOT contain `"UNIQUE constraint failed"`, table names, or column names for both Conflict and BadRequest paths.
- [ ] Test added that asserts the `tracing::warn!` *does* still emit the driver detail (e.g., via `tracing-test` capture). This is the "debugging stays possible" backstop.
- [ ] All ~20 existing `Error::Conflict(...)` / `Error::BadRequest(...)` construction sites migrated to the helper-or-struct form. `cargo build --all-targets` clean.
- [ ] `docs/coding-standards/seaorm.md` § 7 updated to show the new pattern. Two collateral stalenesses to fix in the same edit: (a) all `AppError::*` references in § 7 should be `error::Error::*`; (b) the `Internal` line in the example uses the pre-C2 `format!("Database error: {e}")` shape — update to the current `anyhow::Error::new(e).context("Database error")` shape per design.md § Observability.
- [ ] Also catch the stale `AppError::NotFound` reference in `seaorm.md` § 1's example (around line 42) — it survived PR #112's rename pass.
- [ ] Add a one-line note to `seaorm.md` § 7's prose: "When adding a new UNIQUE constraint that's reachable from a user-facing endpoint, add a pre-check at the service layer so the 409 message can be specific." Positions pre-checks as the way to get specificity beyond the generic safety-net message.
- [ ] `seaorm.md` Document history entry appended.
- [ ] `docs/design.md` § Observability → Error response pattern: the table currently lists `BadRequest(msg)` and `Conflict(msg)` as `String` payloads. Update to the struct shape and add one bullet to the Key behaviors list describing the client / detail split.

## References

- PR-B1 (#25) — the PR where the reviewer flagged this at score 75.
- `docs/coding-standards/seaorm.md` § 7 — the existing convention this Issue revises.
- `backend/src/error.rs` lines 100–112 — current `From<sea_orm::DbErr>` impl.
- `backend/src/error.rs` lines 59–84 — current `IntoResponse` impl.
- `backend/src/error.rs` lines 150–212 — existing tests; new response-body content assertions and the tracing-capture test sit alongside these.
- `docs/design.md` § Observability → Error response pattern.

---

_Filed by Cowork on Brendan's behalf 2026-05-08. Refreshed 2026-05-11 to reflect the `AppError → error::Error` rename, the constraint-leak inventory, and the Option 3 direction Brendan chose after Cowork's design research._

brendanbyrne:
Closes #84.

## Why

`error::Error::Conflict` and `error::Error::BadRequest` were `String`-payload tuple variants. The `From<sea_orm::DbErr>` impl (`seaorm.md` § 7) was implementing that contract literally:

```rust
Some(SqlErr::UniqueConstraintViolation(m)) => Self::Conflict(m),
```

`m` here is the raw backend error message — e.g., `"UNIQUE constraint failed: users.username"` from SQLite. That flowed straight through `IntoResponse` to the wire, leaking table and column names. PR-B1's reviewer flagged it at score 75; this Issue closes the gap.

Direction was settled before pickup (Option 3 in #84 — structured `Conflict { client, detail }` with sanitized client + log-only detail). The Issue body has the constraint-leak inventory and the three rejected alternatives; no further design decision required.

## What this PR does

**Variant restructure (the core).** `Conflict` and `BadRequest` become struct variants `{ client: String, detail: Option<String> }`. `client` is what the response body carries; `detail` (when `Some`) is logged via `tracing::warn!(detail = %d, "Conflict: {client}")` at the `IntoResponse` boundary and never reaches the wire.

**Ergonomic helpers.** `error::Error::conflict(msg)` and `error::Error::bad_request(msg)` build the no-detail common case. Existing call sites migrate from `Error::Conflict("X".into())` to `Error::conflict("X")` — net shorter than today. Struct-literal form is used only when a caller has driver-string detail worth logging (currently just inside `From<DbErr>`).

**`From<DbErr>` mapping.** UNIQUE → `Conflict { client: "Resource already exists", detail: Some(m) }`. FK → `BadRequest { client: "Referenced record does not exist", detail: Some(m) }`. Service-layer pre-checks remain the canonical way to get a *specific* 409/400 message; the generic safety-net text fires only when a pre-check is missed or loses a TOCTOU race.

**Call-site migration.** All ~20 existing construction sites (in `services/sessions.rs`, `services/runs.rs`, `services/helpers.rs`, `routes/auth.rs`, `routes/drink_types.rs`, `domain/race_setup.rs`, `domain/enums.rs`) migrated. Pattern-match sites (in tests across `services/`, `domain/`) destructure the new shape (`{ client, .. }`).

**Tests added (`backend/src/error.rs`):**

- `test_conflict_helper_sets_no_detail`, `test_bad_request_helper_sets_no_detail` — the helpers build the right shape.
- `test_conflict_response_body_omits_driver_detail`, `test_bad_request_response_body_omits_driver_detail` — the regression-blocker: `"UNIQUE"` / `"FOREIGN KEY"` / `users.email` never appear in the response body.
- `test_conflict_with_detail_emits_warn_with_driver_detail`, `test_bad_request_with_detail_emits_warn_with_driver_detail` — the debuggability backstop: when `detail` is `Some`, the warn-level event captures the driver string (verified via `tracing-test`).
- `test_conflict_without_detail_emits_no_warn` — control: the helper-built no-detail case emits nothing.
- Existing `test_unique_constraint_violation_maps_to_conflict` and `test_foreign_key_constraint_violation_maps_to_bad_request` updated to assert the new client/detail shape end-to-end against a real SQLite constraint violation.

**New dev-dep: `tracing-test = "0.2"`.** Drives the warn-emission tests via `#[traced_test]` + `logs_contain(...)`. Per the AC bullet.

## Doc updates (synchronized)

- **`docs/coding-standards/seaorm.md` § 7** — rewrote the `From<DbErr>` example to the struct shape; replaced the pre-C2 `Internal(format!("Database error: {e}"))` with `Internal(anyhow::Error::new(e).context("Database error"))`; added two new rules (struct-shape convention + helpers; pre-check guidance for new user-reachable UNIQUE constraints).
- **`docs/coding-standards/seaorm.md` § 1 (line 42), §§ 3, 4, 12** — collateral `AppError::*` → `error::Error::*` rename sweep that PR #112 missed in this file.
- **`docs/design.md` § Observability → Error response pattern** — variants table updated to `BadRequest { client, detail }` and `Conflict { client, detail }`; new Key behaviors bullet describing the client/detail split, why it exists, and that service-layer pre-checks remain the way to get specific messages.

## Forward-looking gap (not fixed here)

`users.email` has a UNIQUE constraint but no service-layer pre-check — currently dead schema (`email: Set(None)` everywhere). The new "add a pre-check when adding a new user-reachable UNIQUE constraint" rule in `seaorm.md` § 7 handles the policy; the actual email pre-check lands when whatever profile-update endpoint introduces a real email value lands (probably #70).

## Test plan

Reviewer should run the following from `backend/`:

```bash
# 1. Build and the full test suite pass on this branch.
cargo build --all-targets
cargo test --all-targets

# 2. The new error.rs tests run and pass (171 lib tests total; the relevant ones):
cargo test --lib error::tests
# Expect: test_conflict_response_body_omits_driver_detail, test_bad_request_response_body_omits_driver_detail,
#         test_conflict_with_detail_emits_warn_with_driver_detail, test_bad_request_with_detail_emits_warn_with_driver_detail,
#         test_conflict_without_detail_emits_no_warn, test_conflict_helper_sets_no_detail, test_bad_request_helper_sets_no_detail,
#         test_unique_constraint_violation_maps_to_conflict, test_foreign_key_constraint_violation_maps_to_bad_request
#         all show `... ok`.

# 3. Clippy clean (workspace lints + per-crate workspace = true).

# 4. Confirm no legacy form remains in live source:
git grep -nE "Error::Conflict\(|Error::BadRequest\(|AppError::Conflict|AppError::BadRequest" -- backend/src backend/tests
# Expect: no output.
```

Manual API-level smoke (verifies the wire body is sanitized):

```bash
# Boot the backend (defaults are fine).
cd backend && cargo run &
sleep 2

# 1. Trigger a UNIQUE-violation via the From<DbErr> path. The `register` route
#    has a pre-check, so to exercise the safety net we need the duplicate to slip
#    past — easiest is to register two users at the same username via a tight loop
#    (TOCTOU). For a manual one-off, the pre-check will fire first, but the response
#    text should still be the specific "Username already taken" — confirming the
#    pre-check still wins:
curl -s -X POST -H 'content-type: application/json' \
  -d '{"username":"reviewer","password":"password1"}' \
  http://localhost:3000/auth/register | jq
curl -s -X POST -H 'content-type: application/json' \
  -d '{"username":"reviewer","password":"password1"}' \
  http://localhost:3000/auth/register | jq
# Expect second call: HTTP 409, body {"error":"Username already taken"}.

# 2. Trigger an FK violation (no pre-check on this surface). Log in as `reviewer`,
#    then POST a profile update with a bogus character_id:
TOKEN=$(curl -s -X POST -H 'content-type: application/json' \
  -d '{"username":"reviewer","password":"password1"}' \
  http://localhost:3000/auth/login | jq -r .access_token)
curl -s -X PATCH -H "Authorization: Bearer $TOKEN" -H 'content-type: application/json' \
  -d '{"preferred_character_id":99999,"preferred_body_id":1,"preferred_wheel_id":1,"preferred_glider_id":1}' \
  http://localhost:3000/users/me | jq
# Expect: HTTP 400, body {"error":"Invalid character_id"} (pre-check via require_exists),
#   NOT a body containing "FOREIGN KEY" / table names.

# 3. Inspect the server's stderr/logs from steps 1-2: with RUST_LOG=warn or default
#    info, you should NOT see driver-string warns for the pre-check path (because no
#    DbErr fires). To see the safety-net warn fire, start the server with the
#    pre-check temporarily commented out (or rely on the test suite — the
#    `tracing-test`-driven unit tests above exercise the warn! path directly).
```


backend/src/error.rs
```diff
@@ -56,14 +68,44 @@ pub enum Error {
...
+    }
+}
+
 impl IntoResponse for Error {
     fn into_response(self) -> Response {
         let (status, user_message) = match &self {
-            Self::BadRequest(msg) => (StatusCode::BAD_REQUEST, msg.clone()),
+            Self::BadRequest { client, detail } => {
+                if let Some(d) = detail {
+                    tracing::warn!(detail = %d, "BadRequest: {client}");
```
brendanbyrne: 🔵 **Suggestion** — Per `rust.md` § 10, structured fields are preferred over format strings: `info!(user_id = %id, "login succeeded")` not `info!("login succeeded for {}", id)`. The current call mixes the two — `detail` is a structured field, but `client` is interpolated into the message string, so it'll show up in the rendered text rather than as a filterable JSON field.

Consider:
```rust
tracing::warn!(detail = %d, client = %client, "BadRequest");
```

The matching tests at `test_bad_request_with_detail_emits_warn_with_driver_detail` currently assert `logs_contain("BadRequest: Referenced record does not exist")`, which would need adjusting to two separate `logs_contain` calls (one for `client=...`, one for the message body) if you take this. Not blocking — functional behavior is correct as-is.

backend/src/error.rs
```diff
@@ -56,14 +68,44 @@ pub enum Error {
...
+                }
+                (StatusCode::BAD_REQUEST, client.clone())
+            }
             Self::Unauthorized(msg) => (StatusCode::UNAUTHORIZED, msg.clone()),
             Self::Forbidden(msg) => (StatusCode::FORBIDDEN, msg.clone()),
             Self::NotFound(msg) => (StatusCode::NOT_FOUND, msg.clone()),
-            Self::Conflict(msg) => (StatusCode::CONFLICT, msg.clone()),
+            Self::Conflict { client, detail } => {
+                if let Some(d) = detail {
+                    tracing::warn!(detail = %d, "Conflict: {client}");
```
brendanbyrne: 🔵 **Suggestion** — Same `rust.md` § 10 point as the `BadRequest` arm above. Consider:
```rust
tracing::warn!(detail = %d, client = %client, "Conflict");
```
(Apply both arms together so the convention stays uniform.)

backend/src/error.rs
```diff
@@ -283,4 +351,141 @@ mod tests {
...
+        );
+        assert!(
+            logs_contain("BadRequest: Referenced record does not exist"),
+            "warn message missing from captured logs"
+        );
+    }
+
+    #[tokio::test]
+    #[tracing_test::traced_test]
+    async fn test_conflict_without_detail_emits_no_warn() {
```
brendanbyrne, line 489: 🔵 **Suggestion** — Asymmetric coverage: there's `test_conflict_without_detail_emits_no_warn` but no corresponding `test_bad_request_without_detail_emits_no_warn`. The two helper paths (`Error::conflict()` and `Error::bad_request()`) have identical `if let Some(d) = detail { warn! }` gating, so coverage for one and not the other is asymmetric for no functional reason — a matching `BadRequest` test would round out the pattern.

The with-detail case has symmetric coverage (`test_conflict_with_detail_emits_warn_with_driver_detail` + `test_bad_request_with_detail_emits_warn_with_driver_detail`), so it's just this one gap.

brendanbyrne (commented):
## Verdict: Approve (no blocking findings)

**What this PR does.** Restructures `error::Error::Conflict` and `error::Error::BadRequest` from `String`-payload tuple variants to struct variants `{ client, detail }` so the `From<DbErr>` path can carry the raw driver string (e.g., `"UNIQUE constraint failed: users.email"`) into a log-only `tracing::warn!` at the `IntoResponse` boundary while the response body sees only a sanitized generic message. Closes #84.

**Areas covered:** `rust.md` (§ 1, 7, 10, 11, 15), `seaorm.md` (§ 7), `design.md` § Observability, `api-contract.md` (response format unchanged).

**Automated checks (run locally on head SHA):**
- `cargo fmt --check` — clean (only stable-rustfmt warnings about nightly-only options, pre-existing).
- `cargo test` — all suites pass (lib 171, plus integration). The nine new `error::tests` entries (helpers, body-sanitization, warn-emission) all pass.
- `git grep -nE "Error::Conflict\(|Error::BadRequest\(|AppError::Conflict|AppError::BadRequest" -- backend/src backend/tests` — no output (legacy form fully purged).

**Acceptance criteria (#84):** Every box ticks — struct variants, helpers, `From<DbErr>` shape, `IntoResponse` warn, body-sanitization tests, warn-capture test (via `tracing-test = "0.2"`), all ~20 call-sites migrated, `seaorm.md` § 7 + § 1 (line 42 stale `AppError::NotFound`) + the new pre-check rule, `design.md` table + Key behaviors bullet, both Document-history entries.

**Design-doc accuracy:** `design.md` needed updates (variants table row + Key behaviors bullet) — both are present, accurate against the implementation, and historically dated.

**What's good**

- Helper-constructor pattern lands cleanly — `Error::conflict("X")` is net shorter than the old `Error::Conflict("X".into())`, so the migration is a readability win, not just churn.
- The body-sanitization tests assert both the absence of `"UNIQUE"` / `"FOREIGN KEY"` / `"users.email"` from the response body **and** the presence of those substrings in the captured tracing buffer. Future code can't silently leak driver strings without one of those tests failing — exactly the regression-blocker shape #84 called for.
- Caught the pre-existing stale `AppError::NotFound` reference in `seaorm.md` § 1 line 42 that PR #112's rename pass missed — good incidental cleanup carried atomically with the topic.
- The new pre-check guidance rule in `seaorm.md` § 7 frames service-layer pre-checks as the canonical way to get *specific* 409/400 messages and the `From<DbErr>` text as the defense-in-depth safety net — clearer than just describing the mechanism in isolation.
- Forward-looking `users.email` gap explicitly called out as deferred to the future profile-update endpoint (probably #70) — appropriate scope discipline; the new pre-check rule in `seaorm.md` § 7 is what makes the deferral safe.
- `tracing-test` dev-dep is justified inline in `Cargo.toml` with a comment naming what it's used for and pointing back to Issue #84.

Two minor suggestions are inline below — neither blocking.

Issue #101, opened 2026-05-09
Lint cleanup (a): unwrap_used / expect_used


brendanbyrne:
## Scope
Remove `#[allow(clippy::unwrap_used)]` and `#[allow(clippy::expect_used)]` from non-test modules where PR-A1 (#24) added them. For each warning site:
- Propagate via `?` against the typed `AppError` (cleaner once C1 lands).
- In rare controlled-context cases, log + return a fallback.

Test files keep their allows per `rust.md` § 8 — those are the documented exception, gated on `cfg_attr(test, ...)`.

## Why this lint group is high-signal
Per `rust.md` § 8: "Forbidding outright pushes people to `.expect(\"INFALLIBLE\")` which is no better. A warn nudges thinking; tests legitimately want to panic." The lint catches real panic risks.

## Verification
- `cargo clippy --all-targets` produces no warnings for `clippy::unwrap_used` or `clippy::expect_used` in non-test code.
- The workspace-level lint settings (per `rust.md` § 8) are enforced without per-module allows.

PR #142, opened 2026-05-12
132: tower middleware — timeout / concurrency / body / rate limits (PR-F3)

### LINKED ISSUE #132: Tower middleware: timeout / concurrency / body-size / rate limits (PR-F3)
## Context

Adds standard request-shape limits so a misbehaving or malicious client can't trivially exhaust server resources. Each layer is small; the combined effect is a meaningful protection floor before we worry about WAF-style work.

## Scope

- `TimeoutLayer` (e.g., 30s request-level).
- `ConcurrencyLimitLayer` (e.g., 100).
- `RequestBodyLimitLayer` (e.g., 10 MiB to match the upload size cap from `design.md`).
- Add `tower-governor` for rate limiting (default 60 req/min/IP; tunable).

Numbers above are starting points — wire as config knobs from the get-go so we can tune from Unraid without a redeploy.

## Standards refs

- [`docs/coding-standards/tokio.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/coding-standards/tokio.md) § 12

## Dependencies

None.

## Risk

Medium — rate limiter could falsely throttle. Start permissive; tighten with metrics.

## Effort

S.

## Verification

- `curl --data-binary @big-file` against an upload endpoint past the size limit returns 413.
- Concurrent request flood returns 503/429.
- A normal session-flow doesn't trip the rate limiter at expected user pace.

## Compliance-plan reference

[`docs/compliance-plan.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/compliance-plan.md) § PR-F3.

brendanbyrne:
## Reviewer notes

- **Departure from [tokio.md § 12](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/coding-standards/tokio.md#12-backpressure-and-resource-limits)**: used `tower_http::timeout::TimeoutLayer::with_status_code` instead of the rule's literal `tower::timeout::TimeoutLayer`. The tower version's error type is `BoxError`, which `axum::Router::layer` refuses because Axum requires `Error: Into<Infallible>`. Updated the rule itself in the same PR so the doc no longer points at a layer that won't compile here; added a Why callout naming the bound + 408-response benefit.
- **Rate-limit key extractor**: defaults to `PeerIpKeyExtractor` (keys on TCP peer). If we ever sit behind a reverse proxy (Unraid setups often do), this collapses to a global limit. Flagged in a code comment recommending `SmartIpKeyExtractor` for that deploy mode. Deferred — needs an explicit "trusted proxy" decision.
- **`Config` sweep**: 9 sites (4 integration tests + 5 in services/auth.rs) gained the four new fields. Considered factoring a `Config::for_tests()` helper but kept scope tight; the duplication is real but pre-existing and worth a separate cleanup pass.
- **Integration test coverage is narrow by design**: one test for `RequestBodyLimitLayer` at the 413 boundary. The other three layers aren't tested in code because (a) `GovernorLayer` needs `ConnectInfo<SocketAddr>` which axum-test doesn't populate without extra plumbing, (b) `TimeoutLayer` and `ConcurrencyLimitLayer` need timing shenanigans for deterministic asserts. The Issue's Verification section is manual `curl`, which catches those.

## Summary

Wires four Tower layers into the Axum router with config knobs sourced from env vars (tunable from Unraid without redeploy):

- `tower_http::timeout::TimeoutLayer` — `REQUEST_TIMEOUT_SECONDS` (default 30)
- `tower::limit::ConcurrencyLimitLayer` — `REQUEST_CONCURRENCY_LIMIT` (default 100)
- `tower_http::limit::RequestBodyLimitLayer` — `MAX_REQUEST_BODY_BYTES` (default 10 MiB)
- `tower_governor::GovernorLayer` — `RATE_LIMIT_PER_MINUTE` (default 60), per-peer-IP, with a background `retain_recent()` cleanup loop

Server binding switches to `into_make_service_with_connect_info::<SocketAddr>` so `ConnectInfo<SocketAddr>` is populated for the governor key extractor.

## Linked work

- Closes #132
- Per [`docs/compliance-plan.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/compliance-plan.md) § PR-F3
- Standards refs: [`docs/coding-standards/tokio.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/coding-standards/tokio.md) § 12 (rule corrected in this PR)

## How to verify

```bash
cd backend
cargo test --no-fail-fast
```

- [x] All test suites green. Test counts after this PR: lib 173, api_integration 27, auth_integration 21, auth_middleware 8, schema_drift 1, session_integration 19, **middleware_limits 1 (new)**.
- [ ] The new middleware_limits test exercises the body-limit boundary:
```bash
cd backend
cargo test --test middleware_limits
```

Manual checks per the Issue's Verification section — run with the server up (`cargo run` in `backend/` or via Docker):

- [ ] Body limit returns 413:
```bash
# Default 10 MiB cap. Generate an 11 MiB payload and POST it to any endpoint
# that accepts a body (register is the simplest available today).
head -c 11534336 /dev/urandom > /tmp/big.bin
  --data-binary @/tmp/big.bin
# expected: HTTP/1.1 413 Payload Too Large
```
- [ ] Rate limiter throttles a flood from one IP:
```bash
# 70 quick requests against a public endpoint; with default 60/min, the
# last several should return 429.
for i in $(seq 1 70); do
  curl -s -o /dev/null -w "%{http_code}\n" http://localhost:3000/api/v1/hello
done | sort | uniq -c
# expected: a mix of 200s and 429s with roughly 60 of the former.
```
- [ ] Tune knobs without rebuild:
```bash
# Set a tight limit, restart, confirm tighter behavior.
RATE_LIMIT_PER_MINUTE=10 cargo run
```

## Author checklist

- [x] Tests added or updated — `tests/middleware_limits.rs` (body-limit boundary).
- [x] Docs updated — `tokio.md` § 12 timeout-layer rule corrected; Document history entry added.
- [x] Schema changes: none.
- [x] Tested locally end-to-end (`cargo test` + `cargo build`; manual curl deferred to verification step above).

### DIFF
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -4,7 +4,7 @@ use std::{net::SocketAddr, sync::Arc, time::Duration};
 
 use anyhow::Context;
 use axum::{
-    BoxError, Json, Router,
+    Json, Router,
     error_handling::HandleErrorLayer,
     http::StatusCode,
     routing::{get, post, put},
@@ -12,7 +12,9 @@ use axum::{
 use beerio_kart::{
     ARGON2_MAX_CONCURRENT, AppState,
     config::{self, Config},
-    db, routes, services,
+    db,
+    middleware::limits::handle_load_shed_error,
+    routes, services,
 };
 use migration::{Migrator, MigratorTrait};
 use serde::Serialize;
@@ -284,17 +286,4 @@ async fn hello() -> Json<HelloResponse> {
     Json(HelloResponse {
         message: "Hello from Beerio Kart!".to_string(),
     })
-}
-
-/// Translate the only error type expected to reach `HandleErrorLayer` —
-/// `tower::load_shed::error::Overloaded` from `LoadShedLayer` — into a 503.
-/// Anything else is a programming bug (no other layer above this one errors),
-/// so map it to 500 rather than panic.
-async fn handle_load_shed_error(err: BoxError) -> (StatusCode, &'static str) {
-    if err.is::<tower::load_shed::error::Overloaded>() {
-        (StatusCode::SERVICE_UNAVAILABLE, "Service overloaded")
-    } else {
-        tracing::error!(error = %err, "unexpected error reached load-shed handler");
-        (StatusCode::INTERNAL_SERVER_ERROR, "Internal error")
-    }
 }
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -1,6 +1,6 @@
 mod seed;
 
-use std::{sync::Arc, time::Duration};
+use std::{net::SocketAddr, sync::Arc, time::Duration};
 
 use anyhow::Context;
 use axum::{
@@ -11,8 +11,18 @@ use beerio_kart::{ARGON2_MAX_CONCURRENT, AppState, config::Config, db, routes, s
 use migration::{Migrator, MigratorTrait};
 use serde::Serialize;
 use tokio::sync::Semaphore;
+use tower::limit::ConcurrencyLimitLayer;
+use tower_governor::{GovernorLayer, governor::GovernorConfigBuilder};
+// `tower::timeout::TimeoutLayer` (cited in tokio.md § 12) produces a service
+// whose error type is `BoxError`, which `axum::Router::layer` refuses because
+// it needs `Into<Infallible>`. `tower_http::timeout::TimeoutLayer` is the
+// HTTP-aware sibling: on elapsed it returns a real `408 Request Timeout`
+// response, which is the right user-facing behavior anyway. Functionally
+// satisfies the same "cap per-request wall time" rule.
 use tower_http::{
+    limit::RequestBodyLimitLayer,
     services::{ServeDir, ServeFile},
+    timeout::TimeoutLayer,
     trace::TraceLayer,
 };
 use tracing_subscriber::EnvFilter;
@@ -60,6 +70,13 @@ async fn main() -> anyhow::Result<()> {
     seed::run(&db).await.context("Seeding database")?;
     tracing::info!("Seeding complete");
 
+    // Snapshot the limit knobs before `config` is moved into `state`. These
+    // power the Tower middleware stack below.
+    let request_timeout = Duration::from_secs(config.request_timeout_seconds);
+    let concurrency_limit = config.request_concurrency_limit;
+    let max_body_bytes = config.max_request_body_bytes;
+    let rate_limit_per_minute = config.rate_limit_per_minute;
+
     let state = AppState {
         db,
         config,
@@ -69,6 +86,32 @@ async fn main() -> anyhow::Result<()> {
     // Clone the DB connection for the background cleanup task before `state`
     // is moved into the router.
     let cleanup_db = state.db.clone();
+
+    // Per-peer-IP rate limit. The default `PeerIpKeyExtractor` reads the
+    // remote address from the `ConnectInfo<SocketAddr>` extension, so the
+    // server below must be served via `into_make_service_with_connect_info`.
+    // If we ever sit behind a *trusted* reverse proxy, swap to
+    // `SmartIpKeyExtractor` so the limit keys on the real client IP via
+    // `X-Forwarded-For` / `X-Real-IP` instead of the proxy's address.
+    let per_second = std::cmp::max(1, u64::from(rate_limit_per_minute) / 60);
+    let governor_conf = Arc::new(
+        GovernorConfigBuilder::default()
+            .per_second(per_second)
+            .burst_size(rate_limit_per_minute)
+            .finish()
+            .context("invalid rate-limit config")?,
+    );
+    // The governor caches one entry per observed IP. Without periodic
+    // pruning, that map grows unbounded across the process lifetime; the
+    // standard `retain_recent()` cleanup loop is from `tower-governor`'s
+    // README. The background thread lives for the process lifetime.
+    let governor_limiter = governor_conf.limiter().clone();
+    std::thread::spawn(move || {
+        loop {
+            std::thread::sleep(Duration::from_secs(60));
+            governor_limiter.retain_recent();
+        }
+    });
 
     // STATIC_DIR defaults to ../frontend/dist for local dev (running from backend/).
     // In Docker, set to /app/static where the built frontend is copied.
@@ -150,6 +193,22 @@ async fn main() -> anyhow::Result<()> {
             "/api/v1/runs/{id}",
             get(routes::runs::get_run).delete(routes::runs::delete_run),
         )
+        // Request-shape limits (tokio.md § 12). Order is request-flow:
+        // trace wraps everything (so rejections are still logged), governor
+        // is next so abusive IPs lose nothing else, body-size and
+        // concurrency reject before any handler work, timeout is innermost
+        // so it budgets the actual handler — not the wait for a permit.
+        // `.layer()` adds layers as outer wrappers, so the last call is the
+        // outermost: in code order, innermost first.
+        .layer(TimeoutLayer::with_status_code(
+            axum::http::StatusCode::REQUEST_TIMEOUT,
+            request_timeout,
+        ))
+        .layer(ConcurrencyLimitLayer::new(concurrency_limit))
+        .layer(RequestBodyLimitLayer::new(max_body_bytes))
+        .layer(GovernorLayer {
+            config: governor_conf,
+        })
         .layer(TraceLayer::new_for_http())
         .with_state(state)
         // Serve frontend static files. If no API route or static file matches,
@@ -176,9 +235,15 @@ async fn main() -> anyhow::Result<()> {
         .await
         .context("Binding TCP listener on 0.0.0.0:3000")?;
     tracing::info!("Listening on http://localhost:3000");
-    axum::serve(listener, app)
-        .await
-        .context("HTTP server returned an error")?;
+    // `into_make_service_with_connect_info::<SocketAddr>` populates the
+    // `ConnectInfo<SocketAddr>` request extension that `tower-governor`'s
+    // `PeerIpKeyExtractor` reads to bucket rate-limit counters per IP.
+    axum::serve(
+        listener,
+        app.into_make_service_with_connect_info::<SocketAddr>(),
+    )
+    .await
+    .context("HTTP server returned an error")?;
 
     Ok(())
 }
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -4,14 +4,20 @@ use std::{net::SocketAddr, sync::Arc, time::Duration};
 
 use anyhow::Context;
 use axum::{
-    Json, Router,
+    BoxError, Json, Router,
+    error_handling::HandleErrorLayer,
+    http::StatusCode,
     routing::{get, post, put},
 };
-use beerio_kart::{ARGON2_MAX_CONCURRENT, AppState, config::Config, db, routes, services};
+use beerio_kart::{
+    ARGON2_MAX_CONCURRENT, AppState,
+    config::{self, Config},
+    db, routes, services,
+};
 use migration::{Migrator, MigratorTrait};
 use serde::Serialize;
 use tokio::sync::Semaphore;
-use tower::limit::ConcurrencyLimitLayer;
+use tower::{ServiceBuilder, limit::ConcurrencyLimitLayer, load_shed::LoadShedLayer};
 use tower_governor::{GovernorLayer, governor::GovernorConfigBuilder};
 // `tower::timeout::TimeoutLayer` (cited in tokio.md § 12) produces a service
 // whose error type is `BoxError`, which `axum::Router::layer` refuses because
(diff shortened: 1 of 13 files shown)


backend/src/main.rs
```diff
@@ -70,6 +87,32 @@ async fn main() -> anyhow::Result<()> {
...
+    // `SmartIpKeyExtractor` so the limit keys on the real client IP via
+    // `X-Forwarded-For` / `X-Real-IP` instead of the proxy's address.
+    let per_second = std::cmp::max(1, u64::from(rate_limit_per_minute) / 60);
+    let governor_conf = Arc::new(
+        GovernorConfigBuilder::default()
+            .per_second(per_second)
+            .burst_size(rate_limit_per_minute)
+            .finish()
+            .context("invalid rate-limit config")?,
+    );
```
brendanbyrne: 🟡 **Important — rate-limit math doesn't match the env knob's contract.**

`tower_governor::GovernorConfigBuilder::per_second(N)` sets *the interval between replenishments in seconds* — i.e., "1 cell every N seconds", not "N cells per second". Verified against [`tower_governor-0.7.0/src/governor.rs:183`](https://docs.rs/tower_governor/0.7.0/src/tower_governor/governor.rs.html#183):

> Set the interval after which one element of the quota is replenished in seconds.

With `per_second = max(1, n / 60)`:

| `RATE_LIMIT_PER_MINUTE` | computed `per_second` | actual sustained rate |
|---|---|---|
| 60 (default) | 1 | **60/min** ✓ |
| 10 (PR's "tight" demo) | `max(1, 0) = 1` | **60/min** ✗ — 6× the configured value |
| 30 | `max(1, 0) = 1` | **60/min** ✗ |
| 120 | 2 | **30/min** ✗ — half the configured value |

The env var only behaves correctly at exactly 60. The PR description's verification step (`RATE_LIMIT_PER_MINUTE=10 cargo run`) won't actually produce a 10/min limit — it'll produce 10-burst + 60/min sustained.

**Fix** — use the millisecond variant for sub-second granularity:

```rust
let period_ms = (60_000_u64 / u64::from(rate_limit_per_minute)).max(1);
let governor_conf = Arc::new(
    GovernorConfigBuilder::default()
        .per_millisecond(period_ms)
        .burst_size(rate_limit_per_minute)
        .finish()
        .context("Invalid rate-limit config")?,
);
```

Worth adding a test (even just a unit test on a helper that computes `period_ms` from a `u32`) so this stays correct under future tuning.

backend/src/main.rs
```diff
@@ -70,6 +87,32 @@ async fn main() -> anyhow::Result<()> {
...
+    // If we ever sit behind a *trusted* reverse proxy, swap to
+    // `SmartIpKeyExtractor` so the limit keys on the real client IP via
+    // `X-Forwarded-For` / `X-Real-IP` instead of the proxy's address.
+    let per_second = std::cmp::max(1, u64::from(rate_limit_per_minute) / 60);
+    let governor_conf = Arc::new(
+        GovernorConfigBuilder::default()
+            .per_second(per_second)
+            .burst_size(rate_limit_per_minute)
+            .finish()
+            .context("invalid rate-limit config")?,
```
brendanbyrne: 🟡 **Important — lowercase context message.**

Per [`rust.md` § 1](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/coding-standards/rust.md#1-error-handling):

> Error-message strings we *construct* start with a capital letter and have no trailing punctuation. This applies to `AppError` variant payloads, **`anyhow` static contexts** (`.context("Failed to build Set-Cookie header")`), and synthetic `anyhow!` messages.

Fix: `.context("Invalid rate-limit config")?,`

brendanbyrne (commented):
## Review summary

**Verdict:** 🟡 Request Changes — defaults work, but the rate-limit math doesn't match the env knob's contract (the PR's own "tight" demo case `RATE_LIMIT_PER_MINUTE=10` actually allows 60/min sustained).

**What this PR does:** Wires four Tower layers (timeout, concurrency, body limit, governor rate-limit) with env-tunable knobs on `Config`. Updates `tokio.md` § 12 in the same PR to correct the `tower::timeout::TimeoutLayer` → `tower_http::timeout::TimeoutLayer` recipe.

**Areas reviewed:** `rust.md` § 1 (errors), § 8 (lints/comments), § 13 (file length), § 17 (config); `tokio.md` § 8 (background tasks), § 12 (request limits); `design.md` § Observability.

**Automated checks (all green on `8344b6c`):**
- `cargo fmt --check`: clean (modulo nightly-only warnings, pre-existing).
- `cargo clippy --all-targets --all-features -- -D warnings`: 0 warnings.
- `cargo test --no-fail-fast`: 173 lib + 27 api_integration + 21 auth_integration + 8 auth_middleware + **1 middleware_limits (new)** + 1 schema_drift + 19 session_integration. All pass.

---

### Findings not anchored to specific lines

**🔵 Suggestion — Test-config duplication across 5 sites** (auth.rs unit-tests, `tests/api_integration.rs`, `tests/auth_integration.rs`, `tests/auth_middleware.rs`, `tests/session_integration.rs`)

You flagged this in the PR description. Reaffirming as a Suggestion: a `Config::for_tests()` helper (or `..Config::for_tests()` struct-update spread) would absorb future field additions without diff churn. Defer to a separate cleanup PR if the scope-trim preference holds.

### Design-doc accuracy

- ✅ `tokio.md` § 12 timeout-layer rule corrected in this PR; the "Why timeout comes from `tower-http`" Why callout reads exactly right and prevents the same trap next time.
- ✅ Document history bullet added with date, PR number, and motivation.
- ➖ No update needed in `design.md` — request-shape limits aren't currently catalogued there. The four new env vars are only documented via field doc-comments on `Config`. There's no centralized env-var catalog yet, so this isn't a finding — but worth noting for the eventual ops doc.

### What's good

- The `tokio.md` § 12 correction is exemplary process: hitting a doc rule that doesn't compile and fixing the rule in the same PR with a Why callout that names the bound (`Error: Into<Infallible>`).
- `parse_positive_env` is clean and the doc comment explains the actual failure mode ("zero or negative would silently disable the protection"), not just the mechanism.
- The "Snapshot the limit knobs before `config` is moved into `state`" comment is the right kind of *why* comment — answers a question the diff alone leaves dangling.
- PR description preemptively flags the trusted-proxy / `SmartIpKeyExtractor` deferral and the test-config duplication. Self-review saves reviewer cycles.

backend/src/main.rs
```diff
@@ -70,6 +87,32 @@ async fn main() -> anyhow::Result<()> {
...
+    // pruning, that map grows unbounded across the process lifetime; the
+    // standard `retain_recent()` cleanup loop is from `tower-governor`'s
+    // README. The background thread lives for the process lifetime.
+    let governor_limiter = governor_conf.limiter().clone();
+    std::thread::spawn(move || {
+        loop {
+            std::thread::sleep(Duration::from_secs(60));
+            governor_limiter.retain_recent();
+        }
+    });
```
brendanbyrne, line 128: 🔵 **Suggestion — use `tokio::spawn` + `tokio::time::interval` instead of a dedicated OS thread.**

`retain_recent()` is sync but cheap (it grabs a `dashmap` shard lock briefly). Running it from a tokio task instead of a `std::thread::spawn` saves an OS thread for a once-per-minute task and keeps both background loops in this file consistent — the stale-session cleanup at [main.rs:223-232](https://github.com/brendanbyrne/beerio-kart/blob/132/middleware-limits/backend/src/main.rs#L223-L232) already uses `tokio::spawn` + `tokio::time::sleep`.

If you adopt this, the canonical shape per [`tokio.md` § 8](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/coding-standards/tokio.md#8-background-tasks) is `interval(...)` with `MissedTickBehavior::Skip`:

```rust
tokio::spawn(async move {
    let mut tick = tokio::time::interval(Duration::from_secs(60));
    tick.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
    loop {
        tick.tick().await;
        governor_limiter.retain_recent();
    }
});
```

Not blocking — both background loops also share the pre-existing `CancellationToken` / structured-shutdown gap, which is a separate compliance task. But this PR is a natural moment to align the new task with the existing pattern.

backend/src/main.rs
```diff
@@ -150,6 +193,22 @@ async fn main() -> anyhow::Result<()> {
...
+        .layer(TimeoutLayer::with_status_code(
+            axum::http::StatusCode::REQUEST_TIMEOUT,
+            request_timeout,
+        ))
+        .layer(ConcurrencyLimitLayer::new(concurrency_limit))
+        .layer(RequestBodyLimitLayer::new(max_body_bytes))
+        .layer(GovernorLayer {
+            config: governor_conf,
+        })
         .layer(TraceLayer::new_for_http())
```
brendanbyrne, line 246: 🟡 **Important — concurrency-permit wait is unbounded by the request timeout, and `ConcurrencyLimitLayer` queues rather than sheds.**

The layer order is `TimeoutLayer` *innermost*, with `ConcurrencyLimitLayer` outside it. The comment accurately notes that timeout "budgets the actual handler — not the wait for a permit" — but doesn't surface the consequence:

- Request #101 (when 100 are in flight) waits for a permit forever, with no timeout firing and no 503 shed.
- The linked Issue's Verification line says "Concurrent request flood returns 503/429" — but `tower::limit::ConcurrencyLimitLayer` *queues*, it doesn't shed. Only governor returns 429, and only at the IP-pacing rate; a single client opening 101 concurrent connections gets 100 served and 1 hung indefinitely (until the client disconnects).

Two reasonable fixes:

- **Move `TimeoutLayer` outside `ConcurrencyLimitLayer`** so permit-wait counts against the budget. Then a hung client eventually sees a 408. Trade-off: handlers under permit pressure get less than their full budget.
- **Wrap concurrency in `tower::load_shed::LoadShedLayer`** so "would block" becomes a 503. Matches the Issue's stated behavior more directly.

Either is fine; pick one and update the layer-stack comment to name the chosen trade-off ("503 on saturation" vs "408 on long wait" vs "queue indefinitely"). The current comment reads as if a trade-off was considered, but the third option — queue indefinitely — is what the code actually does.

brendanbyrne (commented):
## Re-review at `f29e247`

**Verdict:** 🟡 Request Changes — one new contract-violation finding in the fix code; otherwise excellent. Easy fix.

- `cargo fmt --check`: clean.
- `cargo clippy --all-targets --all-features -- -D warnings`: 0 warnings.
- `cargo test --no-fail-fast`: **178 lib** (was 173, +5 from the new `governor_period_ms` rstest cases) + 27 api_integration + 21 auth_integration + 8 auth_middleware + 1 middleware_limits + 1 schema_drift + 19 session_integration. All pass.

---

### Prior findings — all addressed

| # | Prior finding | Status |
|---|---|---|
| 1 | 🟡 Rate-limit math doesn't match env knob | ✅ Fixed via `config::governor_period_ms` + `per_millisecond`. Rstest covers 60/10/120/30/0 → expected ms intervals. Spot-checked: `60 → 1000ms` (1/s = 60/min ✓), `10 → 6000ms` (1 per 6s = 10/min ✓), `120 → 500ms` (2/s = 120/min ✓) |
| 2 | 🟡 Lowercase `.context("invalid …")` | ✅ Fixed: `"Invalid rate-limit config"` |
| 3 | 🟡 Concurrency wait unbounded / no 503 | ✅ Fixed with `ServiceBuilder` + `HandleErrorLayer` + `LoadShedLayer` + `ConcurrencyLimitLayer`. The layer-stack comment now names the actual trade-off ("503 on saturation") instead of "queue indefinitely" |
| 4 | 🔵 `std::thread::spawn` for governor pruning | ✅ Fixed: `tokio::spawn` + `tokio::time::interval` + `MissedTickBehavior::Skip`, matches the stale-session pattern |

Also nice: the `tower` feature comment in `Cargo.toml` was updated to document the new `load-shed` feature; the `BoxError` / `HandleErrorLayer` / `StatusCode` imports are tidy; the precondition for `HandleErrorLayer` (top-level error type must be `Infallible`) is explained in the comment.

---

### What's good

- The `governor_period_ms` helper's doc comment cites the upstream source line (`tower_governor-0.7.0/src/governor.rs:183`) and explains *why* `per_millisecond` over `per_second`. Saves the next reader an hour of reading governor internals.
- The rstest table covers the right boundaries: default, sub-60 (where the prior bug lived), super-60 (where the divisor truncation lived), and the `0` clamp. Each case has a descriptive `#[case::label(...)]` name.
- The expanded layer-stack comment in `main.rs` explicitly calls out the trade-off (`queue indefinitely` vs `503 immediately`) rather than just describing what the code does — a meaningful improvement over the prior comment.
- The `HandleErrorLayer` / `LoadShedLayer` / `ConcurrencyLimitLayer` triple is bundled in a `ServiceBuilder` with the *why* (the `Error: Into<Infallible>` bound on `Router::layer`) — same memory-aid discipline as the timeout-layer comment.

backend/src/main.rs
```diff
@@ -188,3 +285,16 @@ async fn hello() -> Json<HelloResponse> {
...
+/// Anything else is a programming bug (no other layer above this one errors),
+/// so map it to 500 rather than panic.
+async fn handle_load_shed_error(err: BoxError) -> (StatusCode, &'static str) {
+    if err.is::<tower::load_shed::error::Overloaded>() {
+        (StatusCode::SERVICE_UNAVAILABLE, "Service overloaded")
+    } else {
+        tracing::error!(error = %err, "unexpected error reached load-shed handler");
+        (StatusCode::INTERNAL_SERVER_ERROR, "Internal error")
+    }
+}
```
brendanbyrne: 🟡 **Important — `handle_load_shed_error` breaks the JSON error contract.**

This handler returns `(StatusCode, &'static str)`, which axum's `IntoResponse` renders as `Content-Type: text/plain; charset=utf-8` with the raw string as the body. So on saturation the response is:

```
HTTP/1.1 503 Service Unavailable
Content-Type: text/plain; charset=utf-8

Service overloaded
```

But [`docs/design.md` § Observability](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/design.md#error-response-pattern) commits to:

> **Response format:** All errors return JSON: `{ "error": "<message>" }`

And [`error.rs:118-124`](https://github.com/brendanbyrne/beerio-kart/blob/132/middleware-limits/backend/src/error.rs#L118-L124) implements that contract for every other error path via `Json(ErrorBody { error: user_message })`. Frontend code that parses the body as JSON will fail to parse when the 503 (or this handler's 500 fallback) fires.

**Fix** — return a `Response` with the JSON wrapper:

```rust
async fn handle_load_shed_error(err: BoxError) -> Response {
    if err.is::<tower::load_shed::error::Overloaded>() {
        (
            StatusCode::SERVICE_UNAVAILABLE,
            Json(serde_json::json!({ "error": "Service overloaded" })),
        )
            .into_response()
    } else {
        tracing::error!(error = %err, "Unexpected error reached load-shed handler");
        (
            StatusCode::INTERNAL_SERVER_ERROR,
            Json(serde_json::json!({ "error": "Internal server error" })),
        )
            .into_response()
    }
}
```

Two side-notes folded into the same diff:

- The 500 fallback message is `"Internal error"`; the existing convention from `error.rs:113` is `"Internal server error"`. Match the existing string for client-side consistency (a frontend grouping errors by message would treat these as different).
- `tracing::error!(... "unexpected error ...")` — minor: the rest of the file uses capital-first tracing messages (e.g. `"Stale session cleanup failed"` at `main.rs:229`). [`rust.md` § 1](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/coding-standards/rust.md#1-error-handling)'s capital-first rule technically applies to *constructed error strings* (`AppError`, anyhow contexts, `anyhow!`), not tracing fields — but the existing codebase reads as if the convention extends to tracing. Suggestion-level.

Issue #137, opened 2026-05-12
ActiveModelBehavior::before_save for timestamps (PR-E1)


brendanbyrne:
## Context

Centralizes `created_at` / `updated_at` maintenance in `ActiveModelBehavior::before_save` so service code never sets timestamps by hand. Removes a class of bug where someone forgets to update `updated_at` on an edit path.

## Scope

- For every entity with `created_at` and/or `updated_at`, implement `ActiveModelBehavior::before_save` in a sibling file `entities/{entity}_behavior.rs`, wired in via `entities/mod.rs`.
- Remove all manual `created_at: Set(now)` / `updated_at: Set(now)` calls from service code.

## Standards refs

- [`docs/coding-standards/seaorm.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/coding-standards/seaorm.md) § 1, § 6

## Dependencies

PR-A2 (audit; landed — see [`docs/designs/2026-04-15-rust-audit.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/designs/2026-04-15-rust-audit.md)). Audit identifies which entities need the behavior — confirm the list against the audit before starting.

## Risk

Low. Centralizing behavior reduces bugs, doesn't introduce them.

## Effort

M.

## Verification

- Tests confirm `updated_at` advances on update.
- Integration test inserts a row and reads back the timestamp.
- A grep for manual `created_at: Set` / `updated_at: Set` calls in `backend/src/services/` returns nothing after the PR.

## Compliance-plan reference

[`docs/compliance-plan.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/compliance-plan.md) § PR-E1.

brendanbyrne:
Done in PR #143 (merged as 454430f).

PR #143, opened 2026-05-12
137: ActiveModelBehavior::before_save for managed timestamps

brendanbyrne:
## Summary

- Adds `ActiveModelBehavior::before_save` for the seven entities with `created_at` / `updated_at` columns, each in a sibling `entities/{entity}_behavior.rs`.
- Service / route call sites now leave timestamp fields as `NotSet`; `before_save` populates them on insert (and update for `users.updated_at`).
- Closes the work for compliance-plan PR-E1 ([#137](https://github.com/brendanbyrne/beerio-kart/issues/137)). Aligns with [`seaorm.md` § 1](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/coding-standards/seaorm.md#1-activemodel-vs-model).

## Behavior shape

- **`users`** — `before_save` stamps `created_at` on insert and `updated_at` on insert + update.
- **`sessions`** — `before_save` stamps `created_at` on insert. `last_activity_at` is *application-managed* (advanced by `helpers::touch_session`, not by every write); it stays explicitly set by hand. Treating it as a generic `updated_at` would defeat the stale-session cleanup loop.
- **`drink_types`, `runs`, `session_races`, `session_race_participations`, `run_flags`** — `before_save` stamps `created_at` on insert. These have no `updated_at` column.

## Why a small dep was added

`async-trait = "0.1"` was added to `backend/Cargo.toml`. SeaORM 1.x defines `ActiveModelBehavior::before_save` via `#[async_trait]`, and SeaORM doesn't re-export the macro. The crate is already in the dependency tree transitively via SeaORM, so this just makes the dep direct.

## Two intentional manual `Set` survivors

A grep for `created_at: Set` / `updated_at: Set` in `backend/src/` returns two results, both intentional:

1. **`services/helpers.rs::insert_race_participations`** — uses `Entity::insert_many(...).exec()`, which bypasses `before_save` (only single-row `ActiveModel::insert` / `update` / `save` invoke it; verified against `sea-orm-1.1.19/src/entity/active_model.rs:280`). The bulk path is used deliberately to keep the per-race participant snapshot to one round-trip; the trade is one explicit timestamp, documented inline.
2. **`test_helpers::insert_session_race`** — accepts an explicit `created_at: NaiveDateTime` parameter so backdated-race tests (e.g. stale-session cleanup) can construct races at a chosen point in the past. This is the one helper whose contract is "control the timestamp."

Everything else in `backend/src/services/` is timestamp-Set-free (the Issue's strict verification target).

## Tests

Two new tests in `backend/src/entities/users_behavior.rs::tests` (users is the only entity with both timestamps, so it covers both verification criteria):

- `before_save_stamps_created_at_and_updated_at_on_insert` — inserts a user with `NotSet` timestamps, reads back, asserts both columns are populated and equal (same `before_save` call).
- `before_save_advances_updated_at_on_update` — sleeps 10 ms, mutates a non-timestamp column, asserts `updated_at` advances and `created_at` stays unchanged.

The other `*_behavior.rs` files share the same shape (single-line `created_at`-on-insert), so per-entity duplication wouldn't add coverage.

## Test plan

- [ ] `cd backend && cargo build` succeeds.
- [ ] `cd backend && cargo clippy --all-targets` is clean.
- [ ] `cd backend && cargo test` — full suite green (268+ tests). The two new tests live at `entities::users_behavior::tests::before_save_*`; `cargo test before_save -- --nocapture` runs just them.
- [ ] `grep -rn 'created_at: Set\|updated_at: Set' backend/src/` returns the two documented exceptions and nothing else.

### DIFF
--- a/backend/src/entities/users_behavior.rs
+++ b/backend/src/entities/users_behavior.rs
@@ -1,0 +1,103 @@
+//! `ActiveModelBehavior` for [`super::users`].
+//!
+//! Centralizes `created_at` / `updated_at` maintenance so service code never
+//! sets these timestamps by hand. `before_save` runs on insert and update;
+//! `created_at` is only stamped on insert, `updated_at` on both. See
+//! `docs/coding-standards/seaorm.md` § 1.
+
+use async_trait::async_trait;
+use chrono::Utc;
+use sea_orm::{ActiveModelBehavior, ConnectionTrait, DbErr, Set};
+
+use super::users::ActiveModel;
+
+#[async_trait]
+impl ActiveModelBehavior for ActiveModel {
+    async fn before_save<C>(mut self, _db: &C, insert: bool) -> Result<Self, DbErr>
+    where
+        C: ConnectionTrait,
+    {
+        let now = Utc::now().naive_utc();
+        if insert {
+            self.created_at = Set(now);
+        }
+        self.updated_at = Set(now);
+        Ok(self)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    //! Verifies `before_save` for `users`. The other `*_behavior.rs` files
+    //! share the same shape (stamp `created_at` on insert), differing only in
+    //! which column they touch — exercising the pattern once here is enough.
+
+    use std::time::Duration;
+
+    use sea_orm::{ActiveModelTrait, EntityTrait, IntoActiveModel, Set};
+
+    use crate::{
+        entities::users,
+        test_helpers::{create_user, setup_db},
+    };
+
+    #[tokio::test]
+    async fn before_save_stamps_created_at_and_updated_at_on_insert() {
+        // `create_user` constructs a `users::ActiveModel` with `created_at`
+        // and `updated_at` left as `NotSet`; `before_save` must populate both.
+        let db = setup_db().await;
+        let user_id = create_user(&db, "alice").await;
+
+        let user = users::Entity::find_by_id(user_id.as_str())
+            .one(&db)
+            .await
+            .expect("query user")
+            .expect("user exists");
+
+        // On a fresh insert the two timestamps were stamped in the same
+        // `before_save` call, so they must match exactly.
+        assert_eq!(user.created_at, user.updated_at);
+        let now = chrono::Utc::now().naive_utc();
+        let drift = (now - user.created_at).num_seconds().abs();
+        assert!(drift < 5, "stamped time should be ~now (drift {drift}s)");
+    }
+
+    #[tokio::test]
+    async fn before_save_advances_updated_at_on_update() {
+        let db = setup_db().await;
+        let user_id = create_user(&db, "bob").await;
+
+        let user_t0 = users::Entity::find_by_id(user_id.as_str())
+            .one(&db)
+            .await
+            .expect("query user")
+            .expect("user exists");
+        let created_t0 = user_t0.created_at;
+        let updated_t0 = user_t0.updated_at;
+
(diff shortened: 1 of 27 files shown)


backend/src/entities/users_behavior.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    use sea_orm::{ActiveModelTrait, EntityTrait, IntoActiveModel, Set};
+
+    use crate::{
+        entities::users,
+        test_helpers::{create_user, setup_db},
+    };
+
+    #[tokio::test]
+    async fn before_save_stamps_created_at_and_updated_at_on_insert() {
```
brendanbyrne: 🔵 **Suggestion** — Per `rust.md` § 7 (`test_login_with_wrong_password_returns_401`) and `backend/CLAUDE.md` § Testing, the project convention is the `test_` prefix on every test. Every other test in the lib uses it. The two new tests in this file skip the prefix.

Suggested rename:

```rust
async fn test_before_save_stamps_created_at_and_updated_at_on_insert() { ... }
async fn test_before_save_advances_updated_at_on_update() { ... }
```

Trivial; not a blocker.

brendanbyrne (commented):
## Code review — `137: ActiveModelBehavior::before_save for managed timestamps`

**Verdict:** Approve (one minor naming nit, inline below)

**What this PR does.** Adds `ActiveModelBehavior::before_save` for the seven entities with `created_at` / `updated_at` columns, each in a sibling `entities/{entity}_behavior.rs`. Service / route / test call sites now leave timestamps as `NotSet` and let `before_save` populate them. Closes compliance-plan PR-E1 (#137).

**Areas covered.** `seaorm.md` (§§ 1, 6, 12), `rust.md` (§§ 1, 6, 7, 11, 13, 15), `tokio.md` (async-trait shape).

### Automated checks (all green)

- `cargo test` — 232 passing (182 lib + 50 integration), including both new `before_save_*` tests.
- `grep -rn 'created_at: Set\|updated_at: Set' backend/src/` — returns exactly the two documented exceptions and nothing else.

### Findings

One 🔵 suggestion (test naming convention) — inline on the test names. No 🔴 / 🟡.

### Design-doc accuracy

No changes required. This is an implementation-only PR — no data-model, API, or architecture changes. `seaorm.md` § 1 already prescribes this exact pattern; the codebase now actually does it. `seaorm.md` § 12's pitfall ("don't auto-derive `Default`… set timestamps via `before_save`") still reads correctly — it remains a hazard for any entity *without* a `before_save` impl.

### What's good

- The two intentional manual-`Set` survivors are both load-bearing and well-justified inline. Especially `services/helpers.rs` lines 104–109: the comment explains *why* `insert_many` bypasses `before_save` and *why* we still want the bulk path. That kind of comment pays for itself months later.
- `sessions_behavior.rs`'s module doc-comment explicitly calls out that `last_activity_at` is *not* a generic `updated_at` — getting that wrong would silently neutralize the stale-session cleanup loop. Worth the prose.
- The `next_track` / `skip_turn` change to echo `inserted.created_at.and_utc()` (rather than capturing a separate `now` before insert) is *more correct* than the previous code — the response now reflects the actually-stored timestamp, not a slightly-earlier wall-clock read.
- Per-entity test duplication was correctly skipped: `users` is the only entity with both columns, so testing it covers both verification criteria; the other six share the same `created_at`-on-insert shape.
- New direct dep `async-trait = "0.1"` is justified inline in `Cargo.toml`. Strong-semver utility crate, already in the transitive tree via SeaORM — passes `rust.md` § 15.

Issue #122, opened 2026-05-12
Introduce nutype, define ID newtypes (PR-D1)


brendanbyrne:
## Context

First PR of compliance-plan Stream D (type-driven design). Wraps every primitive ID column (`String`/UUID, `i32`) in a `nutype`-backed newtype so the type system enforces "don't pass a `RunId` where a `UserId` is expected."

## Scope

- Add `nutype` to `[dependencies]`.
- Define ID newtypes in `backend/src/domain/ids.rs`: `UserId`, `RunId`, `SessionId`, `SessionRaceId`, `SessionParticipantId`, `RunFlagId`, `DrinkTypeId`, `TrackId`, `CharacterId`, `BodyId`, `WheelId`, `GliderId`, `CupId`. UUID-backed types wrap `uuid::Uuid`; INTEGER-backed wrap `i32`.
- Migrate API request/response DTOs (in `routes/`) to use the newtypes via `serde(transparent)` / nutype's transparent derives.
- Convert at the entity↔service boundary: where service code reads `model.id` (a `String` or `i32`), wrap in the newtype before passing further. Where service code writes back, unwrap to the primitive.
- Update tests to use the newtype constructors.
- If this PR introduces the codebase's first `proptest` test (e.g. a UUID round-trip on `UserId`), also add `proptest` to `[dev-dependencies]` in the same PR. `proptest` was originally bundled into PR-G2 (#136) but moved here so it lands alongside its first real consumer. See [`docs/compliance-plan.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/compliance-plan.md) § PR-G2 for the lineage.

## Standards refs

- [`docs/coding-standards/rust.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/coding-standards/rust.md) § 2
- [`docs/coding-standards/seaorm.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/coding-standards/seaorm.md) § 6

## Dependencies

PR-C1 / PR-C2 (`AppError::BadRequest` for parse failures) — both landed (#105, #107). No blockers.

## Risk

Medium. Long diff; each conversion site is a chance to break.

## Effort

L. Touches every route, every service, every DTO.

## Verification

- All tests pass.
- API contract unchanged (verified by running through a few endpoints manually and confirming request/response shapes are identical).

## Compliance-plan reference

[...]

PR #144, opened 2026-05-13
127: #[tracing::instrument] audit on services and handlers

### LINKED ISSUE #127: #[tracing::instrument] audit on services and handlers (PR-F5)
## Context

Surfaced and scoped-out during PR #27 review (Argon2 spawn_blocking) — the reviewer correctly flagged that the new async helpers lack `#[tracing::instrument]`, a real `tokio.md` § 10 violation, but scoped it out because no other public async fn in `services/` or `routes/` is annotated either. This Issue makes the gap a tracked work item so it doesn't get rediscovered and re-dismissed each PR.

## Scope

- Annotate every public `async fn` in `backend/src/services/` and `backend/src/routes/` with `#[tracing::instrument]`.
- Use `skip(...)` for non-`Debug` or large arguments (e.g. `db`, password fields, full request bodies); use `fields(...)` to surface IDs (`user_id`, `session_id`, etc.) so log lines inside the function carry them automatically.
- Don't instrument trivial helpers in tight loops (per `tokio.md` § 10).

## Standards refs

- [`docs/coding-standards/tokio.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/coding-standards/tokio.md) § 10

## Dependencies

None. Could land in parallel with most Stream D/E PRs.

## Risk

Low. Adds tracing spans; no behavior change.

## Effort

M. Mostly mechanical, but every service/handler needs a thoughtful `skip` / `fields` choice.

## Verification

- Run with `RUST_LOG=info,beerio_kart=debug`; hit a handler that traverses two service calls and confirm log lines from inside both nested spans carry the parent's `user_id` / `session_id` fields.
- `grep -rL '#\[tracing::instrument' backend/src/services backend/src/routes` reports only files with no public async fns.

## Compliance-plan reference

[`docs/compliance-plan.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/compliance-plan.md) § PR-F5.

brendanbyrne:
## Summary

Closes #127 (PR-F5, in `Hardening: Backend compliance plan`).

- Adds `#[tracing::instrument]` to **every public `async fn`** in [backend/src/services/](backend/src/services/) and [backend/src/routes/](backend/src/routes/), per [`docs/coding-standards/tokio.md`](docs/coding-standards/tokio.md) § 10.
- No behavior change. Routes return the same responses; service logic is unchanged. The PR adds tracing spans so log lines emitted inside services / handlers automatically carry their parent span's fields (`user_id`, `session_id`, etc.).

### Conventions used

- **Services** use `skip(...)` to drop `db` connections, passwords, hashes, and request bodies, with `fields(...)` surfacing the IDs callers search by (`user_id`, `session_id`, `run_id`, `session_race_id`, …).
- **Routes** use `skip_all` because Axum destructures extractor patterns (`State(state)`, `Json(body)`, `Path(id)`) and the `tracing-attributes` macro can't reference those bindings by name. `fields(...)` re-adds the wanted fields explicitly.
- **Leaf helpers** in `services/helpers.rs` and `services/session_context.rs` are tagged `level = "debug"`. Service entry points and route handlers use the default INFO level, so per-request span volume at INFO stays low while `RUST_LOG=…=debug` exposes the drill-down chain.
- **Underscore-prefix removals**: handlers that took the `User` extractor only for auth-gating used `_user: User`; now that `fields(user_id = %user.user_id)` actually uses the binding, the underscore would trigger `clippy::used_underscore_binding`. Renamed to `user`. Same treatment for `_user_id` in `services::sessions::skip_turn`.

### Issue's grep verification

```
$ grep -rL '#\[tracing::instrument' backend/src/services backend/src/routes
backend/src/services/mod.rs
backend/src/routes/mod.rs
```

Both `mod.rs` files contain only re-exports — no public `async fn`s — matching the issue's expected output.

## Test plan

Automated:

- [x] `cd backend && cargo build` — clean compile (proves all `#[tracing::instrument]` invocations are well-formed).
- [x] `cd backend && cargo clippy --all-targets -- -D warnings` — no warnings (catches the underscore-binding renames and any field-expression typos).
- [x] `cd backend && cargo test` — full suite, all 259 tests pass.

Manual span-propagation check (the issue's Verification step):

- [x] Start the backend in one terminal:

  ```bash
  cd backend && RUST_LOG='info,beerio_kart=debug' cargo run
  ```

- [x] In another terminal, register a user and grab the access token:

  ```bash
  TOKEN=$(curl -s -X POST http://localhost:3000/api/v1/auth/register \
    -d '{"username":"trace-test","password":"password123"}' \
    | jq -r .access_token)
  ```

- [x] Hit a handler that traverses two service calls:

  ```bash
  curl -s -X POST http://localhost:3000/api/v1/sessions \
    -H "Authorization: Bearer $TOKEN" \
    -d '{"ruleset":"random"}'
  ```

- [x] Confirm the backend log shows a `create_session` route span containing a `create_session` service span, both carrying `user_id` and `ruleset`. To see a concrete inner-emit line, temporarily add `tracing::debug!("probe")` inside `services::sessions::create_session` — its event will be prefixed with the nested span chain.

Done locally with the probe + `POST /api/v1/sessions {"ruleset":"random"}`:

```
DEBUG create_session{user_id=e252…d2e3 ruleset=random}:
      create_session{user_id=e252…d2e3 ruleset=random}:
      beerio_kart::services::sessions: PROBE: inside services::sessions::create_session
```

Both nested spans carry `user_id` and `ruleset`. ✓


brendanbyrne (commented):
## Verdict

✅ Approve — no Critical or Important findings; four optional 🔵 Suggestions are posted inline.

## What this PR does

Adds `#[tracing::instrument]` to every public `async fn` in `backend/src/services/` and `backend/src/routes/` per [`tokio.md` § 10](docs/coding-standards/tokio.md). No behavior change — log lines emitted inside services/handlers now automatically carry parent-span fields (`user_id`, `session_id`, `ruleset`, …).

## What was verified

- **Coverage.** Re-ran the Issue's verification check (`grep -rL '#\[tracing::instrument' backend/src/services backend/src/routes`) — only the two `mod.rs` re-export files surface, matching the expected output. Hand-walked every `pub async fn` in both directories; every one has the attribute (some with multi-line attribute syntax).
- **Skip-list discipline.** Every `skip(...)` / `skip_all` correctly drops `db`, passwords, hashes, and request bodies. No secret crosses into a span `fields(...)` entry. Matches `tokio.md` § 10 and `rust.md` § 10.
- **`skip_all` on route handlers.** Correctly works around the `tracing-attributes` limitation that the macro can't reference Axum extractor pattern bindings (`State(state)`, `Json(body)`, `Path(id)`). Routes re-add the wanted fields explicitly.
- **Level layering.** Leaf helpers in `services/helpers.rs` and `services/session_context.rs` use `level = "debug"`; service entry points and route handlers use the default INFO. Per-request INFO span volume stays low; `RUST_LOG=…=debug` exposes the drill-down chain. Lines up with the `rust.md` § 10 level guidance.
- **`_user` → `user` renames.** Necessary because newly-referenced bindings in `fields(...)` would trigger `clippy::used_underscore_binding`. Reviewed each rename — no functional change beyond removing the leading underscore.
- **Span-propagation probe in the PR body.** The nested `create_session` route → service span chain carrying `user_id` and `ruleset` confirms the instrumentation actually composes end-to-end, not just compiles.

## Automated checks

- `cargo fmt --check` — clean (only the usual nightly-feature notes for `imports_granularity`/`group_imports`).
- `cargo clippy --all-targets --all-features -- -D warnings` — clean. No `clippy::used_underscore_binding` regression from the `_user` → `user` renames.
- `cargo test` — all **259 tests pass** (lib + 7 integration suites + doc-tests).

## Design-doc accuracy

No `docs/design.md` changes required — this is purely tracing decoration; data model, API surface, and architecture are unchanged. The rule being enforced (`tokio.md` § 10) is already documented in the standards.

## Findings

Four 🔵 Suggestions, posted inline. All optional; none block merge.

backend/src/routes/auth.rs
```diff
@@ -238,6 +240,7 @@ pub async fn login(
 /// validation, the token type is not `refresh`, the user no longer exists,
 /// or the token's `refresh_token_version` doesn't match the DB (revoked via
 /// logout or password change). `Internal` for token-issue or DB failures.
+#[tracing::instrument(skip_all)]
```
brendanbyrne: 🔵 **Suggestion** — `refresh` has `skip_all` with no fields, so the resulting span has no `user_id` until a service span opens. Since the user is derived from the refresh-token cookie inside the body, you could pre-declare it on the attribute and record it once known:

```rust
#[tracing::instrument(skip_all, fields(user_id = tracing::field::Empty))]
pub async fn refresh(
    State(state): State<AppState>,
    headers: HeaderMap,
) -> Result<impl IntoResponse, Error> {
    // ... after decoding the refresh claims ...
    tracing::Span::current().record(
        "user_id",
        tracing::field::display(&claims.sub),
    );
    // ...
}
```

Same trick works for `register` / `login` to surface a stable `user_id` once the row is created/fetched (in addition to the existing `username` field). Not blocking — the parent span on subsequent service calls already gets the right ID once it's known; this just makes search-by-`user_id` uniform across every auth handler.

backend/src/routes/users.rs
```diff
@@ -81,6 +83,10 @@ pub async fn get_user(
 /// Propagates the errors of [`user_service::update_profile`]: `Forbidden`
 /// if a user tries to modify another user, `NotFound` if the target user
 /// doesn't exist, `BadRequest` for invalid race-setup or drink-type IDs.
+#[tracing::instrument(
+    skip_all,
+    fields(actor_user_id = %auth_user.user_id, target_user_id = %id),
```
brendanbyrne: 🔵 **Suggestion** — `update_user` uses `actor_user_id` for the caller while every other route uses plain `user_id`. The naming follows the service-layer convention (`update_profile` has two distinct user IDs to disambiguate), but in the *handler* the caller is the only "user_id" present until the service span opens.

Consider also surfacing under `user_id` so log filters work the same across all route spans:

```rust
#[tracing::instrument(
    skip_all,
    fields(user_id = %auth_user.user_id, target_user_id = %id),
)]
```

Trivial preference — defensible either way. Flagging because uniform field names are the thing that makes structured logs queryable.

backend/src/services/helpers.rs
```diff
@@ -151,6 +163,7 @@ pub async fn touch_session<C: ConnectionTrait>(
 ///
 /// Returns `BadRequest` with a formatted message on miss (e.g., "Invalid
 /// `character_id`"); `Internal` for unexpected DB failures.
+#[tracing::instrument(level = "debug", skip(db, id), fields(entity = %entity_label))]
```
brendanbyrne, line 166: 🔵 **Suggestion** — Skipping `id` is necessary here (it's a fully generic `<<E as EntityTrait>::PrimaryKey as PrimaryKeyTrait>::ValueType` that may not impl `Display`/`Debug` uniformly), so surfacing just `entity = %entity_label` is the right call today.

Worth flagging for the future, though: when debugging "which `character_id` did this race-setup validation reject?" you'd want the actual ID in the span. Not actionable today without tightening the bound; the newtype-ID rollout (when it gets here) makes this trivial — every newtype already impls `Display`, so an added `T: Display` bound on the primary-key value type lets us write `fields(entity = %entity_label, id = %id)` without any caller breakage.

backend/src/routes/auth.rs
```diff
@@ -96,6 +96,7 @@ fn extract_refresh_cookie(headers: &HeaderMap) -> Option<String> {
 /// Returns `BadRequest` if the username is empty / >30 chars or the password
 /// is <8 / >128 chars; `Conflict` if the username is taken; `Internal` for
 /// password-hash, token-issue, or DB failures.
+#[tracing::instrument(skip_all, fields(username = %body.username))]
```
brendanbyrne: 🔵 **Suggestion** — Logging `username = %body.username` is fine: `rust.md` § 10's never-log list is passwords / JWTs / free-text user content, none of which usernames are. On pre-auth handlers it's the only correlator available.

Worth being aware that failed-login probes will pour attacker-supplied strings into the `username` field — typically what you *want* operationally (you can grep "attempts against admin"), but it does mean log-aggregation cardinality grows with attack traffic. If that ever becomes a problem, the standard mitigation is to drop the field for failed-auth events specifically rather than for the whole handler. No change needed now — flagging for awareness.

backend/src/routes/auth.rs
brendanbyrne: Addressed in 327b99d1ac03a565ecd5b1e120610d3874051b90 — pre-declared `user_id = tracing::field::Empty` on `register` / `login` / `refresh` and `Span::current().record(...)`'d it after the user is known (Uuid in register, DB lookup in login/refresh).

backend/src/routes/users.rs
brendanbyrne: Addressed in 327b99d1ac03a565ecd5b1e120610d3874051b90 — renamed `actor_user_id` → `user_id` on the `update_user` route span; `target_user_id` stays for the path-extracted ID. Service layer keeps actor/target framing where both IDs are present.

backend/src/services/helpers.rs
brendanbyrne: Acknowledged and parked for D1 ([#122](https://github.com/brendanbyrne/beerio-kart/issues/122)). Posted a note on that issue ([comment](https://github.com/brendanbyrne/beerio-kart/issues/122#issuecomment-4437275477)) so the `Display`-bound + `fields(... id = %id)` change surfaces in the same PR that introduces the newtype IDs — no caller churn that way.

backend/src/routes/auth.rs
brendanbyrne: Acknowledged. Codified the awareness as a new rule under [rust.md § 10](https://github.com/brendanbyrne/beerio-kart/blob/127/tracing-instrument-audit/docs/coding-standards/rust.md#10-logging-with-tracing) in a7f05f4 — same PR. No code change today (rate-limit middleware [#132](https://github.com/brendanbyrne/beerio-kart/issues/132) caps probes at the door, no log aggregator to bill against); the rule names the standard per-event field-drop mitigation if it ever bites.

backend/src/routes/auth.rs
```diff
@@ -261,6 +272,7 @@ pub async fn refresh(
         .one(&state.db)
         .await?
         .ok_or_else(|| Error::Unauthorized("User not found".into()))?;
+    tracing::Span::current().record("user_id", tracing::field::display(&user.id));
```
brendanbyrne: 🔵 **Suggestion** — Consider recording `user_id` earlier in `refresh`, right after `validate_refresh_token` succeeds (around line 263), rather than waiting for the DB lookup here. The JWT signature is already verified at that point, so `claims.sub` is trustworthy:

```rust
let claims = auth_service::validate_refresh_token(&cookie_value, &state.config)
    .map_err(|_| Error::Unauthorized("Invalid or expired refresh token".into()))?;
tracing::Span::current().record("user_id", tracing::field::display(&claims.sub));

if claims.token_type != "refresh" {
    return Err(Error::Unauthorized("Invalid token type".into()));
}
// ... user lookup, etc. The existing record() at line 275 becomes redundant.
```

This covers two extra failure modes — "wrong token type" and "user no longer exists" — with `user_id` populated, which is exactly the "investigate failed refresh for user X" case. Truly marginal — the current placement is correct, the wins are rare-failure-mode-only.

brendanbyrne (commented):
## Follow-up review (commits `a7f05f4`, `327b99d`)

✅ **Approve.** All four 🔵 suggestions from the prior review addressed:

1. ✅ `register` / `login` / `refresh` now pre-declare `user_id = tracing::field::Empty` and call `Span::current().record(...)` at the right point.
2. ✅ `routes::users::update_user` renamed `actor_user_id` → `user_id` for cross-route uniformity (`target_user_id` retained).
3. (No-op — flagged for the newtype-ID rollout, not this PR.)
4. ✅ Promoted from a PR comment to a documented [`rust.md` § 10 rule](docs/coding-standards/rust.md) with rationale, mitigation, and a doc-history entry. Better than I asked for.

### Recording points are well-chosen

- **register** — records right after the UUID mint. Username-conflict path returns earlier, so the failed-conflict span has just `username` (correct — no `user_id` existed yet).
- **login** — records right after the user is found. User-not-found path returns earlier and leaves `user_id` Empty (correct); wrong-password path runs past the record and carries `user_id` (correct — exactly the info you'd want for that failure mode).
- **refresh** — records after the DB lookup. One marginal nit posted inline.

### Automated checks

- `cargo test` — 259/259 pass.

### Design-doc accuracy

No `docs/design.md` changes required. The `rust.md` standards-doc update carries a doc-history entry per `docs/CLAUDE.md`. ✓

One 🔵 nit posted inline — truly optional.

backend/src/routes/auth.rs
brendanbyrne: Addressed in a03a984128a2a57caa10af0d539d87e9f6ba1803 — moved the `record(user_id, ...)` call to right after `validate_refresh_token` succeeds. JWT signature verification makes `claims.sub` trustworthy at that point. Two more failure modes ("Invalid token type", "User not found") now surface with `user_id` populated; the version-mismatch path benefits too. Agreed it's strictly more useful with no real safety trade.

Issue #136, opened 2026-05-12
Add rstest / insta as dev-dependencies (PR-G2)


brendanbyrne:
## Context

Introduces two test-ergonomics crates the standards doc recommends (`rstest` and `insta`). Demo-only migration in this PR — broad adoption happens organically later as test files get touched.

The third crate `rust.md` § 7 recommends, `proptest`, was originally in this scope but was moved to PR-D1 / PR-D2 so it lands alongside its first real consumer (newtype round-trip invariants). Per `rust.md` § 7, `proptest` is reserved for algebraic invariants; the codebase has none today, and adding the crate with no consumer is the "introduce dep before there's a need" pattern the project pushes back on. See [`docs/compliance-plan.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/compliance-plan.md) § PR-G2 for the lineage.

## Scope

- Add `rstest` and `insta` as `[dev-dependencies]`.
- Migrate one or two existing repetitive test files to `rstest` as a demo.
- Add `insta` for one HTTP integration test as a demo.
- Don't migrate everything in this PR — leave that for organic adoption.

## Standards refs

- [`docs/coding-standards/rust.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/coding-standards/rust.md) § 7

## Dependencies

None.

## Risk

Low.

## Effort

S.

## Verification

- Demo tests pass.
- `cargo insta accept` workflow works (i.e. the snapshot file lands in `tests/snapshots/` and the test passes on a re-run).

## Compliance-plan reference

[`docs/compliance-plan.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/compliance-plan.md) § PR-G2.

_Scope updated 2026-05-12 by Cowork after Claude Code flagged the missing-consumer seam during PR #139 review cleanup. Filed by Cowork on Brendan's behalf._

Issue #114, opened 2026-05-11
Doc-comment audit: 4 missing-doc lints (PR-G3)


brendanbyrne:
## Scope

Remove the four workspace-level `= "allow"` lines for `missing_errors_doc`, `missing_panics_doc`, `doc_markdown`, and `too_long_first_doc_paragraph` from `backend/Cargo.toml`. Fix the **149 sites** the lints surface (counts as of 2026-05-11; will drift before pickup — re-count from `cargo clippy` at start of work).

Count snapshot:

| Lint | Sites (2026-05-11) |
|---|---|
| `clippy::missing_errors_doc` | 72 |
| `clippy::doc_markdown` | 62 |
| `clippy::missing_panics_doc` | 9 |
| `clippy::too_long_first_doc_paragraph` | 6 |

Reproduce the counts:

```bash
cd backend
cargo clean -p beerio-kart
cargo clippy --all-targets -- --cap-lints=warn \
  -W clippy::missing_errors_doc \
  -W clippy::doc_markdown \
  -W clippy::missing_panics_doc \
  -W clippy::too_long_first_doc_paragraph \
  > /tmp/g3.log 2>&1
grep -E "^warning: " /tmp/g3.log | sort | uniq -c | sort -rn
```

## Likely tactic

- `missing_errors_doc` (72) and `missing_panics_doc` (9) need `# Errors` / `# Panics` sections on public fallible / panicking functions.
- `doc_markdown` (62) is mechanical: wrap identifiers in backticks.
- `too_long_first_doc_paragraph` (6) requires light editorial work — break the first paragraph.

## Split-at-pickup option

If the list is still large at pickup, split into:

- **G3-a (semantic):** `missing_errors_doc` + `missing_panics_doc`.
- **G3-b (mechanical):** `doc_markdown` + `too_long_first_doc_paragraph`.

Otherwise: single PR.

## Why this lint group matters

Sharpens `cargo doc` output. The workspace `Cargo.toml` already commits to "raise to warn once docs catch up" in the comment next to `missing_errors_doc = "allow"`. Closing those allows fulfills that commitment for the four lints most directly tied to public-API documentation quality.

## Dependency note

[...]

PR #145, opened 2026-05-13
122: PR-D1 — introduce nutype, define ID newtypes

brendanbyrne:
## Summary

- Wrap every primitive ID column in a `nutype`-backed newtype so the compiler catches "wrong-id-in-this-arg" bugs and UUID-shaped wire input is validated once at the boundary.
- UUID-backed: `UserId`, `SessionId`, `RunId`, `SessionRaceId`, `SessionParticipantId`, `RunFlagId`, `DrinkTypeId` (wrap `uuid::Uuid`).
- Integer-backed: `TrackId`, `CharacterId`, `BodyId`, `WheelId`, `GliderId`, `CupId` (wrap `i32`).
- Entities keep SeaORM-default primitives per `seaorm.md` § 6 — conversion happens at the entity↔service boundary via `Type::from_db(&col_str)` (UUIDs, returns `Internal` on bad UUID) and `Type::new(int)` (integers).
- Wire format unchanged: nutype's transparent `Serialize`/`Deserialize` delegate to the inner type, so UUID newtypes still serialize as the canonical hyphenated string and integer newtypes as JSON numbers.
- `proptest` added as a dev-dep alongside its first consumer (a 256-case `UserId` serde round-trip in `domain::ids::tests`). Lineage: bundled into PR-G2 (#136); moved here per the issue body so the dep lands with real coverage.
- `Config::admin_user_id` is now typed (`Option<UserId>`), so a malformed `ADMIN_USER_ID` env var fails fast at startup with `"ADMIN_USER_ID must be a valid UUID"` instead of silently refusing admin access.

## Behavior change worth noting

A non-UUID path segment on routes like `GET /sessions/:id` now returns **400** from axum's `Path<SessionId>` extractor before the handler runs (previously returned 404 from a DB miss). Two `test_get_nonexistent_*_returns_404` integration tests were updated to use valid-shape-but-not-in-DB UUIDs to exercise the original FK-miss path; the type-extractor 400 is a strict improvement and is captured by the unit test `test_user_id_deserialize_rejects_non_uuid_string` in `domain/ids.rs`.

API contract for valid inputs is unchanged. Verified by running the full test suite (264 tests across lib + 6 integration files, all passing).

## Test plan

- [ ] `cargo test` — all 264 tests pass (187 lib + 27 + 21 + 8 + 1 + 1 + 19 integration).
- [ ] `cargo clippy --all-targets --all-features` — clean.
- [ ] `cargo +nightly fmt --check` — clean.
- [ ] Smoke test the running server: `POST /api/v1/auth/register`, log in, `GET /api/v1/users/me`, create a session, pick a track, submit a run. All payloads should look identical to before this PR.
- [ ] Hit `GET /api/v1/sessions/not-a-uuid` and confirm the response is 400 (was 404). Hit `GET /api/v1/sessions/<random-valid-uuid>` and confirm 404.

## Note on G3 interaction

PR-G3 (#114) deliberately landed ahead of D1; `# Errors` doc-sections that referenced `i32` / `String` IDs have been swept in this PR alongside the function-signature changes.

### DIFF
--- a/backend/src/middleware/auth.rs
+++ b/backend/src/middleware/auth.rs
@@ -1,9 +1,6 @@
 use axum::{extract::FromRequestParts, http::request::Parts};
 
-use crate::{
-    domain::{UserId, ids::parse_db_id},
-    error::Error,
-};
+use crate::{domain::UserId, error::Error};
 
 /// Extractor that validates the JWT from the `Authorization: Bearer <token>`
 /// header and makes the authenticated user's info available to handlers.
@@ -58,7 +55,12 @@ impl FromRequestParts<crate::AppState> for User {
         // serializes a real `UserId`. A non-UUID `sub` here means a forged or
         // foreign-issued token slipped past signature validation — treat it
         // as unauthorized, not internal, so we don't leak corruption details.
-        let user_id = parse_db_id::<UserId>(&claims.sub, "access token sub claim")
+        // Goes through the typed `FromStr` surface (nutype-derived) rather
+        // than the internal `parse_db_id` helper, which is for entity-column
+        // reads where the error context should name the column.
+        let user_id: UserId = claims
+            .sub
+            .parse()
             .map_err(|_| Error::Unauthorized("Invalid user id in token".to_string()))?;
 
         Ok(Self {
--- a/backend/src/middleware/auth.rs
+++ b/backend/src/middleware/auth.rs
@@ -1,6 +1,9 @@
 use axum::{extract::FromRequestParts, http::request::Parts};
 
-use crate::{domain::UserId, error::Error};
+use crate::{
+    domain::{UserId, ids::parse_db_id},
+    error::Error,
+};
 
 /// Extractor that validates the JWT from the `Authorization: Bearer <token>`
 /// header and makes the authenticated user's info available to handlers.
@@ -51,8 +54,15 @@ impl FromRequestParts<crate::AppState> for User {
             return Err(Error::Unauthorized("Invalid token type".to_string()));
         }
 
+        // `sub` is mint-controlled by `create_access_token`, which always
+        // serializes a real `UserId`. A non-UUID `sub` here means a forged or
+        // foreign-issued token slipped past signature validation — treat it
+        // as unauthorized, not internal, so we don't leak corruption details.
+        let user_id = parse_db_id::<UserId>(&claims.sub, "access token sub claim")
+            .map_err(|_| Error::Unauthorized("Invalid user id in token".to_string()))?;
+
         Ok(Self {
-            user_id: UserId::new(claims.sub),
+            user_id,
             username: claims.username,
         })
     }
@@ -86,7 +96,7 @@ impl FromRequestParts<crate::AppState> for AdminUser {
             .as_ref()
             .ok_or_else(|| Error::Forbidden("Admin access not configured".to_string()))?;
 
-        if auth_user.user_id.as_str() != admin_id {
+        if auth_user.user_id != *admin_id {
             return Err(Error::Forbidden("Admin access required".to_string()));
         }
 
--- a/backend/src/routes/auth.rs
+++ b/backend/src/routes/auth.rs
@@ -126,12 +126,12 @@ pub async fn register(
     // Hash password (offloaded to the blocking pool, capped by argon2_limit)
     let password_hash = auth_service::hash_password(&state.argon2_limit, body.password).await?;
 
-    let user_id = uuid::Uuid::new_v4().to_string();
+    let user_id = UserId::new_v4();
     tracing::Span::current().record("user_id", tracing::field::display(&user_id));
 
     // Timestamps are populated by `users::ActiveModelBehavior::before_save`.
     let new_user = users::ActiveModel {
-        id: Set(user_id.clone()),
+        id: Set((&user_id).into()),
         username: Set(username.to_string()),
         email: Set(None),
         password_hash: Set(password_hash),
@@ -158,7 +158,7 @@ pub async fn register(
         Json(Response {
             access_token,
             user: UserInfo {
-                id: UserId::new(user_id),
+                id: user_id,
                 username: username.to_string(),
             },
         }),
@@ -201,7 +201,8 @@ pub async fn login(
         .await;
         return Err(Error::Unauthorized("Invalid username or password".into()));
     };
-    tracing::Span::current().record("user_id", tracing::field::display(&user.id));
+    let user_id = UserId::from_db(&user.id)?;
+    tracing::Span::current().record("user_id", tracing::field::display(&user_id));
 
     // Verify password (offloaded to the blocking pool, capped by argon2_limit)
     let password_ok = auth_service::verify_password(
@@ -215,9 +216,9 @@ pub async fn login(
     }
 
     // Generate tokens
-    let access_token = auth_service::create_access_token(&user.id, &user.username, &state.config)?;
+    let access_token = auth_service::create_access_token(&user_id, &user.username, &state.config)?;
     let refresh_token =
-        auth_service::create_refresh_token(&user.id, user.refresh_token_version, &state.config)?;
+        auth_service::create_refresh_token(&user_id, user.refresh_token_version, &state.config)?;
     let headers = make_refresh_headers(&refresh_token, &state.config)?;
 
     Ok((
@@ -225,7 +226,7 @@ pub async fn login(
         Json(Response {
             access_token,
             user: UserInfo {
-                id: UserId::new(user.id),
+                id: user_id,
                 username: user.username,
             },
         }),
@@ -277,6 +278,7 @@ pub async fn refresh(
         .one(&state.db)
         .await?
         .ok_or_else(|| Error::Unauthorized("User not found".into()))?;
+    let user_id = UserId::from_db(&user.id)?;
 
     // Version mismatch means the token was revoked (logout or password change)
     if claims.refresh_token_version != user.refresh_token_version {
@@ -284,11 +286,11 @@ pub async fn refresh(
     }
 
     // Issue new tokens
-    let access_token = auth_service::create_access_token(&user.id, &user.username, &state.config)?;
+    let access_token = auth_service::create_access_token(&user_id, &user.username, &state.config)?;
 
     // Rotate: issue a fresh refresh token with same version but new expiry
     let new_refresh =
-        auth_service::create_refresh_token(&user.id, user.refresh_token_version, &state.config)?;
+        auth_service::create_refresh_token(&user_id, user.refresh_token_version, &state.config)?;
     let resp_headers = make_refresh_headers(&new_refresh, &state.config)?;
 
     Ok((resp_headers, Json(RefreshResponse { access_token })))
@@ -307,7 +309,7 @@ pub async fn refresh(
 #[tracing::instrument(skip_all, fields(user_id = %user.user_id))]
 pub async fn logout(State(state): State<AppState>, user: User) -> Result<impl IntoResponse, Error> {
     // Look up user to get current version
-    let db_user = users::Entity::find_by_id(&user.user_id)
+    let db_user = users::Entity::find_by_id(String::from(&user.user_id))
         .one(&state.db)
         .await?
         .ok_or_else(|| {
@@ -359,7 +361,7 @@ pub async fn change_password(
     }
 
     // Look up user
-    let db_user = users::Entity::find_by_id(&user.user_id)
+    let db_user = users::Entity::find_by_id(String::from(&user.user_id))
         .one(&state.db)
         .await?
         .ok_or_else(|| Error::NotFound("User not found".into()))?;
@@ -382,7 +384,7 @@ pub async fn change_password(
     // `updated_at` is bumped by `users::ActiveModelBehavior::before_save`.
     let new_version = db_user.refresh_token_version + 1;
     let username = db_user.username.clone();
-    let user_id = db_user.id.clone();
+    let user_id = UserId::from_db(&db_user.id)?;
     let mut active: users::ActiveModel = db_user.into_active_model();
     active.password_hash = Set(new_hash);
     active.refresh_token_version = Set(new_version);
--- a/backend/src/routes/auth.rs
+++ b/backend/src/routes/auth.rs
@@ -309,7 +309,7 @@ pub async fn refresh(
 #[tracing::instrument(skip_all, fields(user_id = %user.user_id))]
 pub async fn logout(State(state): State<AppState>, user: User) -> Result<impl IntoResponse, Error> {
     // Look up user to get current version
-    let db_user = users::Entity::find_by_id(String::from(&user.user_id))
+    let db_user = users::Entity::find_by_id(user.user_id)
         .one(&state.db)
         .await?
         .ok_or_else(|| {
@@ -361,7 +361,7 @@ pub async fn change_password(
     }
 
     // Look up user
-    let db_user = users::Entity::find_by_id(String::from(&user.user_id))
+    let db_user = users::Entity::find_by_id(user.user_id)
         .one(&state.db)
         .await?
         .ok_or_else(|| Error::NotFound("User not found".into()))?;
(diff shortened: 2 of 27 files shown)


backend/src/services/runs.rs
```diff
@@ -437,7 +449,7 @@ pub async fn delete_run(
         .await?
         .ok_or_else(|| Error::NotFound("Run not found".to_string()))?;
 
-    if run.user_id.as_str() != user_id.as_str() {
+    if run.user_id != user_id.to_string() {
```
brendanbyrne: 🔵 **Suggestion** — Mixed comparison style across the entity boundary. After PR-D1, the convention everywhere else is "lift to newtype once, compare typed." Here we stringify the typed `user_id` to match the entity-side `String`. Both work and the entity-side `String` is correct per `seaorm.md` § 6, but the typed form keeps the pattern uniform and surfaces a corruption-in-DB 500 instead of a silent false-negative compare:

```rust
let owner = UserId::from_db(&run.user_id)?;
if owner != *user_id {
    return Err(Error::Forbidden("Only the run's owner can delete it".to_string()));
}
```

Low-impact (FK-protected, can't realistically misfire), so flagging as Suggestion not blocker. Same shape repeats at `services/sessions.rs:960` — happy to roll into one follow-up PR if you'd like.

backend/src/services/sessions.rs
```diff
@@ -949,13 +957,13 @@ pub async fn leave_session(
     active_participant.left_at = Set(Some(now));
     active_participant.update(&txn).await?;
 
-    let is_host_leaving = session.host_id.as_str() == user_id.as_str();
+    let is_host_leaving = session.host_id == user_id.to_string();
```
brendanbyrne: 🔵 **Suggestion** — Same mixed-style note as `services/runs.rs:452`: stringifying the typed `user_id` to compare against the entity's `String` field. Typed form would be:

```rust
let host_id = UserId::from_db(&session.host_id)?;
let is_host_leaving = host_id == *user_id;
```

Non-blocking; uniformity argument only.

brendanbyrne (commented):
## Verdict: Approve

**What this PR does:** Wraps every primitive ID column in a `nutype`-backed newtype (`UserId`, `SessionId`, `RunId`, `SessionRaceId`, `SessionParticipantId`, `RunFlagId`, `DrinkTypeId` over `Uuid`; `TrackId`, `CharacterId`, `BodyId`, `WheelId`, `GliderId`, `CupId` over `i32`). Conversion happens at the entity↔service boundary per `seaorm.md` § 6. Adds `proptest` alongside its first consumer.

**Areas covered:** `rust.md`, `seaorm.md`, `api-contract.md`, `design.md` (no changes required — the newtype boundary is documented in the standards, not the design doc).

- `cargo +nightly fmt --check` — clean.
- `cargo test --lib` — 187/187 passing.

## Findings

Two inline 🔵 Suggestions below — neither blocks merge.

One additional 🔵 Suggestion (file-level, not line-anchorable):

**🔵 `Path<SessionId>` extractor 400s don't carry the `api-contract.md` § 3 error shape.** The PR body correctly notes the behavior change (non-UUID path segment → 400, was 404), and integration tests verify the status code. But the response body comes from axum's `PathRejection`, not `error::Error::IntoResponse`, so it lacks the `{ error, code }` envelope and a `code` field per `api-contract.md` § 3 / § 8. This is a **pre-existing gap** (`Path<Uuid>` had it too — see e.g. `Path<Uuid>` on the original `/sessions/:id` route before this PR), not something introduced here. Worth a follow-up Issue after PR-D2/D4 land more `Path<T>` newtypes so a single registry entry like `invalid_path_param` can cover them all.

## Design-doc accuracy

No `design.md` changes required. The newtype boundary policy lives in `rust.md` § 2 and `seaorm.md` § 6, neither of which were touched. The PR doesn't change schema, API surface, or service architecture in ways `design.md` describes.

## Compliance-plan / lineage

Matches PR-D1 (#122) scope exactly. The `proptest` move from PR-G2 → PR-D1 is documented in the compliance-plan history (2026-05-12 entry) and called out in the PR body and Issue body — full lineage on the record.

## What's good

- Two-macro split (`uuid_id_newtype!` / `i32_id_newtype!`) cleanly handles the UUID vs i32 axis. Both owned and borrowed `From<…> for Value` cover `Column::Foo.eq(id)` and `find_by_id(id)` without ceremony at call sites.
- `parse_db_id` boundary helper names the column in the error context — exactly the diagnostic you want when a corrupted UUID surfaces as a 500.
- `Config::admin_user_id` typed as `Option<UserId>` makes a misconfigured `ADMIN_USER_ID` a startup failure rather than a silent admin-lockout. The `.context("ADMIN_USER_ID must be a valid UUID")` form aligns with `rust.md` § 11.
- The 256-case `proptest` round-trip on `UserId` is a textbook fit for `rust.md` § 7 ("algebraic invariants"). Bonus: it justifies the dep arriving in this PR.
- `test_user_id_deserialize_rejects_non_uuid_string` is the right unit-test for the "parse, don't validate" boundary change — locks down the 400-from-extractor behavior at the unit level so integration tests don't carry the entire spec.
- The doc comment block at the top of `domain/ids.rs` explains the `From<&Self> for Value` and `From<&Self> for String` rationale, including which SeaORM API each is for. Future contributors won't have to reverse-engineer the macro.

backend/src/routes/game_data.rs
```diff
@@ -15,50 +16,113 @@ use crate::{
 // ── Response types ───────────────────────────────────────────────────
 
 #[derive(Serialize)]
-pub struct SimpleItem {
-    pub id: i32,
+pub struct CharacterResponse {
```
brendanbyrne, line 20: Missing `#[serde(rename_all = "snake_case")]` on this and the four sibling response DTOs introduced in this PR (`BodyResponse` L26, `WheelResponse` L33, `GliderResponse` L40, `CupResponse` L47). [`docs/coding-standards/rust.md` § 14 (L390)](https://github.com/brendanbyrne/beerio-kart/blob/43c2df78d47484484be65a4cf73778a5040153ed/docs/coding-standards/rust.md#L389-L392) says: *"Wire format is `snake_case` everywhere. Add `#[serde(rename_all = \"snake_case\")]` to every struct that crosses the wire — request DTOs, response DTOs, event payloads."* Today's fields are single-word so output is unchanged, but the rule exists so a future multi-word field doesn't silently ship as `imagePath`.

backend/src/middleware/auth.rs
```diff
@@ -1,6 +1,9 @@
 use axum::{extract::FromRequestParts, http::request::Parts};
 
-use crate::{domain::UserId, error::Error};
+use crate::{
+    domain::{UserId, ids::parse_db_id},
```
brendanbyrne: Importing `parse_db_id` via the internal `crate::domain::ids::` path instead of going through the typed surface (`UserId::from_db`) is inconsistent with every other call site in this PR. The reason for the direct call is the custom column label (`"access token sub claim"` — see [L61](https://github.com/brendanbyrne/beerio-kart/blob/43c2df78d47484484be65a4cf73778a5040153ed/backend/src/middleware/auth.rs#L60-L62)), which `UserId::from_db` doesn't expose. If that context is important enough to bypass the typed helper, consider widening `from_db` to take an optional label, or adding a `UserId::from_token_sub(...)` constructor so the helper doesn't have to leak out of `domain::ids`.

backend/src/services/runs.rs
brendanbyrne: Regresses to stringly-typed comparison via `user_id.to_string()`. The pattern established elsewhere in this same PR (e.g. [`session_context.rs::require_host`](https://github.com/brendanbyrne/beerio-kart/blob/43c2df78d47484484be65a4cf73778a5040153ed/backend/src/services/session_context.rs#L53-L60)) parses the entity column with `UserId::from_db(&run.user_id)?` and compares typed values. Both sides happen to produce canonical hyphenated UUID strings so it's semantically correct today, but it leaves an entity-layer `String` flowing into a comparison that the rest of the codebase now does in the typed domain.

backend/src/services/sessions.rs
brendanbyrne: Same pattern as `runs.rs:452` — `session.host_id == user_id.to_string()` is a stringly-typed comparison, while the rest of the PR converts at the entity boundary with `UserId::from_db(&session.host_id)?` (see [`session_context.rs::require_host`](https://github.com/brendanbyrne/beerio-kart/blob/43c2df78d47484484be65a4cf73778a5040153ed/backend/src/services/session_context.rs#L53-L60)). Worth aligning with the typed pattern for consistency.

backend/src/services/session_context.rs
```diff
@@ -51,7 +51,8 @@ impl SessionContext {
     ///
     /// Returns `Forbidden` if `user_id` is not the host.
     pub fn require_host(&self, user_id: &UserId) -> Result<(), Error> {
-        if self.session.host_id.as_str() != user_id.as_str() {
+        let host_id = UserId::from_db(&self.session.host_id)?;
```
brendanbyrne: `UserId::from_db(&self.session.host_id)?` re-parses on every `require_host` call. PR [#82](https://github.com/brendanbyrne/beerio-kart/pull/82) established the *cache typed ID once in `SessionContext`* pattern for `session_id` (see [the struct doc at L18-L21](https://github.com/brendanbyrne/beerio-kart/blob/43c2df78d47484484be65a4cf73778a5040153ed/backend/src/services/session_context.rs#L18-L21) — *"the helper methods can borrow `&self.session_id` instead of allocating a fresh wrapper on every call"*). Same fix applies here: add `host_id: UserId` to the struct, populate once in `load_active` alongside `session_id`, borrow in `require_host`. Not a hot path, but inconsistent with the struct's stated rationale.

backend/src/routes/auth.rs
```diff
@@ -307,7 +309,7 @@ pub async fn refresh(
 #[tracing::instrument(skip_all, fields(user_id = %user.user_id))]
 pub async fn logout(State(state): State<AppState>, user: User) -> Result<impl IntoResponse, Error> {
     // Look up user to get current version
-    let db_user = users::Entity::find_by_id(&user.user_id)
+    let db_user = users::Entity::find_by_id(String::from(&user.user_id))
```
brendanbyrne: `find_by_id(String::from(&user.user_id))` is inconsistent with every other `find_by_id` site in this PR, which passes the newtype directly (e.g. `find_by_id(target_user_id)`, `find_by_id(&run.session_race_id)`). The `From<UserId> for String` impl on the newtype is precisely what enables `find_by_id(&user.user_id)` to compile. Same on L364.

brendanbyrne (commented):
### Code review

Found 6 issues (5 inline, 1 below — outside any hunk so it can't be line-anchored):

**`require_exists` tracing follow-up parked for D1 not landed**

[`backend/src/services/helpers.rs#L165-L168`](https://github.com/brendanbyrne/beerio-kart/blob/43c2df78d47484484be65a4cf73778a5040153ed/backend/src/services/helpers.rs#L165-L168) — still `skip(db, id)` with only `fields(entity = %entity_label)`. The PR [#144 discussion](https://github.com/brendanbyrne/beerio-kart/pull/144#discussion_r3231463352) parked this explicitly for D1 / #122 ([the response](https://github.com/brendanbyrne/beerio-kart/pull/144#discussion_r3231522718) confirms it would land *"in the same PR that introduces the newtype IDs — no caller churn that way"*). Newtypes now all derive `Display`, so adding a `Display` bound on the generic and surfacing `fields(entity = %entity_label, id = %id)` is mechanical here.

brendanbyrne (commented):
### Code review

Found 6 issues (5 inline, 1 below — outside any hunk so it can't be line-anchored):

**`require_exists` tracing follow-up parked for D1 not landed**

[`backend/src/services/helpers.rs#L165-L168`](https://github.com/brendanbyrne/beerio-kart/blob/43c2df78d47484484be65a4cf73778a5040153ed/backend/src/services/helpers.rs#L165-L168) — still `skip(db, id)` with only `fields(entity = %entity_label)`. The PR [#144 discussion](https://github.com/brendanbyrne/beerio-kart/pull/144#discussion_r3231463352) parked this explicitly for D1 / #122 ([the response](https://github.com/brendanbyrne/beerio-kart/pull/144#discussion_r3231522718) confirms it would land *"in the same PR that introduces the newtype IDs — no caller churn that way"*). Newtypes now all derive `Display`, so adding a `Display` bound on the generic and surfacing `fields(entity = %entity_label, id = %id)` is mechanical here.

backend/src/routes/game_data.rs
brendanbyrne, line 19: Missing `#[serde(rename_all = "snake_case")]` on this and the four sibling response DTOs introduced in this PR (`BodyResponse` L26, `WheelResponse` L33, `GliderResponse` L40, `CupResponse` L47). [`docs/coding-standards/rust.md` § 14 (L390)](https://github.com/brendanbyrne/beerio-kart/blob/43c2df78d47484484be65a4cf73778a5040153ed/docs/coding-standards/rust.md#L389-L392) says: *"Wire format is `snake_case` everywhere. Add `#[serde(rename_all = \"snake_case\")]` to every struct that crosses the wire — request DTOs, response DTOs, event payloads."* Today's fields are single-word so output is unchanged, but the rule exists so a future multi-word field doesn't silently ship as `imagePath`.

backend/src/middleware/auth.rs
brendanbyrne, line 4: Importing `parse_db_id` via the internal `crate::domain::ids::` path instead of going through the typed surface (`UserId::from_db`) is inconsistent with every other call site in this PR. The reason for the direct call is the custom column label (`"access token sub claim"` — see [L61](https://github.com/brendanbyrne/beerio-kart/blob/43c2df78d47484484be65a4cf73778a5040153ed/backend/src/middleware/auth.rs#L60-L62)), which `UserId::from_db` doesn't expose. If that context is important enough to bypass the typed helper, consider widening `from_db` to take an optional label, or adding a `UserId::from_token_sub(...)` constructor so the helper doesn't have to leak out of `domain::ids`.

backend/src/services/runs.rs
brendanbyrne, line 452: Regresses to stringly-typed comparison via `user_id.to_string()`. The pattern established elsewhere in this same PR (e.g. [`session_context.rs::require_host`](https://github.com/brendanbyrne/beerio-kart/blob/43c2df78d47484484be65a4cf73778a5040153ed/backend/src/services/session_context.rs#L53-L60)) parses the entity column with `UserId::from_db(&run.user_id)?` and compares typed values. Both sides happen to produce canonical hyphenated UUID strings so it's semantically correct today, but it leaves an entity-layer `String` flowing into a comparison that the rest of the codebase now does in the typed domain.

backend/src/services/sessions.rs
brendanbyrne, line 960: Same pattern as `runs.rs:452` — `session.host_id == user_id.to_string()` is a stringly-typed comparison, while the rest of the PR converts at the entity boundary with `UserId::from_db(&session.host_id)?` (see [`session_context.rs::require_host`](https://github.com/brendanbyrne/beerio-kart/blob/43c2df78d47484484be65a4cf73778a5040153ed/backend/src/services/session_context.rs#L53-L60)). Worth aligning with the typed pattern for consistency.

backend/src/services/session_context.rs
brendanbyrne, line 54: `UserId::from_db(&self.session.host_id)?` re-parses on every `require_host` call. PR [#82](https://github.com/brendanbyrne/beerio-kart/pull/82) established the *cache typed ID once in `SessionContext`* pattern for `session_id` (see [the struct doc at L18-L21](https://github.com/brendanbyrne/beerio-kart/blob/43c2df78d47484484be65a4cf73778a5040153ed/backend/src/services/session_context.rs#L18-L21) — *"the helper methods can borrow `&self.session_id` instead of allocating a fresh wrapper on every call"*). Same fix applies here: add `host_id: UserId` to the struct, populate once in `load_active` alongside `session_id`, borrow in `require_host`. Not a hot path, but inconsistent with the struct's stated rationale.

backend/src/routes/auth.rs
brendanbyrne, line 312: `find_by_id(String::from(&user.user_id))` is inconsistent with every other `find_by_id` site in this PR, which passes the newtype directly (e.g. `find_by_id(target_user_id)`, `find_by_id(&run.session_race_id)`). The `From<UserId> for String` impl on the newtype is precisely what enables `find_by_id(&user.user_id)` to compile. Same on L364.

PR #147, opened 2026-05-13
133: PR-D2 — validated string newtypes

### LINKED ISSUE #133: Validated string newtypes via nutype (PR-D2)
## Context

Second PR of Stream D. Moves validation logic (length checks, character-set rules, email shape) out of service bodies and into nutype constructors. After this lands, `Username::try_from(input)?` does the work that an `if username.len() > 30 { return Err(...) }` block currently does.

## Scope

- Define `Username`, `EmailAddress`, `PasswordHash`, `DrinkTypeName`, `ImagePath`, `RunNotes` in `backend/src/domain/strings.rs` using `nutype` with appropriate validation predicates.
- Move existing validation logic out of services into newtype constructors. Where a service had `if username.len() > 30 { return Err(...) }`, replace with `Username::try_from(input)?`.
- Update tests.
- If `proptest` isn't already in `[dev-dependencies]` (e.g. PR-D1 / #122 didn't introduce it), bring it in here alongside the first string-newtype invariant test — valid input round-trips, invalid input rejected. See [`docs/compliance-plan.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/compliance-plan.md) § PR-G2 for the lineage (proptest was originally bundled into PR-G2 / #136 but moved to land alongside its first real consumer).

## Standards refs

- [`docs/coding-standards/rust.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/coding-standards/rust.md) § 2

## Dependencies

PR-D1 (depends on the ID newtypes being established so this PR layers cleanly on top).

## Risk

Medium — moving validation from services to constructors changes which type the parse error materializes from. Existing `AppError::BadRequest` text should be preserved or improved; check that error messages reaching the client don't regress.

## Effort

M.

## Verification

- Tests cover happy + invalid cases for each newtype.
- A representative invalid request (e.g. 31-char username) still returns `400 BadRequest` with a useful message.

## Compliance-plan reference

[`docs/compliance-plan.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/compliance-plan.md) § PR-D2.

brendanbyrne:
Closes #133.

## Summary

- Defines `Username`, `EmailAddress`, `Password`, `PasswordHash`, `DrinkTypeName`, `ImagePath`, `RunNotes` in a new `backend/src/domain/strings.rs`, each with sanitize/validate predicates per [`rust.md` § 2](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/coding-standards/rust.md).
- Moves the inline length checks from `register`, `change_password`, and `create_drink_type` into newtype constructors. Existing `400 BadRequest` messages preserved.
- Threads `PasswordHash` through `hash_password` / `verify_password` and `ImagePath` through every game-data response DTO + `SessionRaceInfo`, with `from_db` helpers that surface a stored-value validation failure as `Internal` (data corruption) rather than `BadRequest`.

## Scope notes

- **`Password` (plaintext) was not on the Issue's 6-newtype list, but added** to absorb the `body.password.len() < 8 || > 128` check. Without it, the "move validation to types" principle applies unevenly — five fields get newtypes and password validation stays inline. `Password` has restricted derives (no `Debug`, `Display`, `Serialize`, `Deserialize`) so it can't accidentally land in logs or wire payloads. Confirmed with Brendan before writing.
- **Length validation now counts characters, not bytes**, for password and drink type name (matches Username's existing `.chars().count()` semantics and the standards example). For ASCII inputs identical; for multi-byte glyphs slightly more permissive. Documented inline.
- **Email validation is intentionally minimal** (`looks_like_email` checks for exactly one `@` and a `.` in the domain part). RFC 5322 isn't regex-checkable; the canonical answer remains "send a confirmation email." Useful as a typo guard; the field is currently always `None` at insert time.

## Test plan

All automated checks pass locally:

```bash
cd backend
cargo build
cargo test                                          # 308 passed
cargo clippy --all-targets -- -D warnings           # clean
cargo +nightly fmt -- --check                       # clean
```

Manual smoke test against a fresh dev server:

```bash
# Start the backend (separate terminal)
cd backend && cargo run

# 1. Register with valid input → 201
  -d '{"username":"alice","password":"correct horse battery staple"}'

# 2. Register with too-short password → 400 "Password must be 8-128 characters"
  -d '{"username":"bob","password":"short"}'

# 3. Register with 31-char username → 400 "Username must be 1-30 characters"
  -d '{"username":"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa","password":"valid_password_123"}'

# 4. Register with whitespace-only username → 400 (trim → empty)
  -d '{"username":"   ","password":"valid_password_123"}'

# 5. Login as alice → 200 with access_token
ACCESS=$(curl -s -X POST http://localhost:3000/api/v1/auth/login \
  -d '{"username":"alice","password":"correct horse battery staple"}' \
  | jq -r .access_token)
echo "access=$ACCESS"

# 6. Change password — wrong current → 401
curl -i -X PUT http://localhost:3000/api/v1/auth/password \
  -H "Authorization: Bearer $ACCESS" -H 'Content-Type: application/json' \
  -d '{"current_password":"wrong","new_password":"newvalid_password_456"}'

# 7. Change password — new password too short → 400
curl -i -X PUT http://localhost:3000/api/v1/auth/password \
  -H "Authorization: Bearer $ACCESS" -H 'Content-Type: application/json' \
  -d '{"current_password":"correct horse battery staple","new_password":"short"}'

# 8. Create drink type — valid → 200/201
curl -i -X POST http://localhost:3000/api/v1/drink-types \
  -H "Authorization: Bearer $ACCESS" -H 'Content-Type: application/json' \
  -d '{"name":"  IPA  ","alcoholic":true}'
# expect: name in response is "IPA" (trimmed)

# 9. Create drink type — empty after trim → 400 "Drink type name must be 1-200 characters"
curl -i -X POST http://localhost:3000/api/v1/drink-types \
  -H "Authorization: Bearer $ACCESS" -H 'Content-Type: application/json' \
  -d '{"name":"   ","alcoholic":false}'

# 10. List characters — image_path renders as a bare string (transparent serde)
curl -s http://localhost:3000/api/v1/characters \
  -H "Authorization: Bearer $ACCESS" | jq '.[0]'
# expect: { "id": ..., "name": "...", "image_path": "/path/to/image.png" }
# image_path must NOT serialize as {"0": "..."} — that would be the
# non-transparent newtype bug

# 11. GET a cup by id — image_path on both cup and nested tracks
curl -s http://localhost:3000/api/v1/cups/1 \
  -H "Authorization: Bearer $ACCESS" | jq '{image_path, first_track: .tracks[0]}'
```

If any of the above 400 responses regress to 422 (axum extractor failure), that's a bug — DTOs accept `String` and the handler runs the boundary parse so the contract stays at 400. Test 10's note on transparent serialization is the canary for "did I lose the wire contract."


backend/src/domain/strings.rs
```diff
@@ -0,0 +1,518 @@
...
+}
+
+fn looks_like_email(s: &str) -> bool {
+    // Intentionally minimal: RFC 5322 is famously not regex-checkable, and
+    // the canonical answer is "send a confirmation email." We enforce the
+    // bare minimum that catches obvious typos: exactly one `@`, with at
+    // least one character on each side. Total length is bounded by the
+    // `len_char_max` on the newtype itself.
+    let parts: Vec<&str> = s.split('@').collect();
+    parts.len() == 2 && !parts[0].is_empty() && !parts[1].is_empty() && parts[1].contains('.')
```
brendanbyrne: 🔵 **Suggestion** — `looks_like_email` allocates a `Vec<&str>` for a shape check. Three-item iterator destructuring would be allocation-free:

```rust
let mut it = s.split('@');
matches!((it.next(), it.next(), it.next()),
    (Some(local), Some(domain), None)
    if !local.is_empty() && !domain.is_empty() && domain.contains('.'))
```

No rule in `rust.md` mandates the change — this is style, and on the email validation path the allocation is one small `Vec` per construction. Not blocking.

backend/src/domain/strings.rs
```diff
@@ -0,0 +1,518 @@
...
+        Eq,
+        Hash,
+        Display,
+        AsRef,
+        Serialize,
+        Deserialize,
+        TryFrom,
+    ),
+)]
+pub struct ImagePath(String);
```
brendanbyrne, line 234: 🔵 **Suggestion** — `ImagePath` is the odd one out among the trim-then-bound newtypes. `DrinkTypeName`/`Username`/`RunNotes` all use `sanitize(trim)` + `len_char_min/max`; `ImagePath` skips `sanitize(trim)` and encodes the trim inside `is_non_empty_path = !s.trim().is_empty()`. Net behavior on construction matches (whitespace-only rejected), but a stored value like `"  /foo  "` would survive un-trimmed because there's no sanitizer running.

Today `ImagePath` is only built from `from_db` against seed rows, so this is moot in practice — worth either adding `sanitize(trim)` for consistency with the other newtypes, or a one-line doc-comment noting the intent ("`ImagePath` is read-only at the boundary; no trim is applied so a corrupt value with surrounding whitespace surfaces verbatim in the response").

backend/src/routes/auth.rs
```diff
@@ -191,26 +192,33 @@ pub async fn login(
...
-    let password_ok = auth_service::verify_password(
-        &state.argon2_limit,
-        body.password,
-        user.password_hash.clone(),
-    )
-    .await?;
+    // Verify password (offloaded to the blocking pool, capped by argon2_limit).
+    // Stored hash is validated at the boundary; a corrupt value surfaces as
+    // Internal (data corruption, not a wrong-password 401).
+    let stored_hash = PasswordHash::from_db(user.password_hash.clone())?;
```
brendanbyrne, line 218: 🔵 **Suggestion** — partial application of the type-driven design principle on `Username`. This line correctly routes the stored hash through `PasswordHash::from_db` (data corruption → `Internal`). The `user.username` two lines down (used at line 227 to build the access-token claim) stays a raw `String` — no `Username::from_db` analogue. Same asymmetry in `change_password` around line 391.

Defensible — the value was validated at insert time, and `seaorm.md` § 6 scopes the boundary conversion to the service layer rather than mandating it at every read site. But if you want full symmetry with the `ImagePath::from_db` / `PasswordHash::from_db` pattern this PR establishes, a `Username::from_db` follow-up would close the loop. Not blocking — flagging so we can decide whether to track it.

brendanbyrne (commented):
**Verdict:** Approve — three 🔵 Suggestions only, no blockers.

**What this PR does:** Defines seven `nutype`-validated string newtypes (`Username`, `EmailAddress`, `Password`, `PasswordHash`, `DrinkTypeName`, `ImagePath`, `RunNotes`), moves length/shape checks out of register/change_password/create_drink_type and into the constructors, and threads `PasswordHash`/`ImagePath` through `services/auth.rs` and the game-data response DTOs.

**Areas reviewed:** `rust.md`, `api-contract.md`, `design.md`.

**Automated checks (all clean):**
- `cargo fmt --check` ✅
- `cargo clippy --all-targets --all-features -- -D warnings` ✅
- `cargo test` ✅ (308 passed)

**Design-doc accuracy:** no changes required. `design.md` does not cover this layer of detail (precedent set by PR-D1 / #122 which introduced the ID newtypes the same way), and the wire format is preserved by `nutype`'s transparent serde — verified by `test_username_serializes_as_bare_string` and the two property tests.

**Scope decisions I checked and chose not to flag**

- **DTOs keep `username: String` rather than `username: Username`.** Intentional per the PR body — switching the DTO field type would surface validation failures as axum's 422 extractor error instead of the existing 400. The current `try_from + map_err(BadRequest)` chain inside the handler preserves the wire contract.
- **Defensive `Password::try_from`/`PasswordHash::try_from` on values that can't actually fail** (the static dummy hash in `login`, and the round-trip after `argon2` produces a fresh hash in `services/auth.rs`). Comments call out the trade-off and explain the `?` over `expect()` choice. Defensive coding for an invariant violation is fine.

**What's good**

- Tight, single-concern scope. Each substitution preserves the existing error surface.
- `from_db` helpers consistently route stored-value failures to `Error::Internal` with a column-naming context — matches the `ids.rs::parse_db_id` pattern and lets the boundary log capture the column on corruption.
- `Password` / `PasswordHash` deliberately omit `Debug`/`Display`/`Serialize`. The doc comments explain *why* (and why `PasswordHash` still gets `Debug` — different leakage profile). The kind of derive-list discipline `rust.md` § 2 was reaching for.
- The `argon2::password_hash::PasswordHash as Argon2Hash` import alias with the rationale-comment above it is exactly the right shape — name collisions hidden by `use` aliases get explained, not silently grepped over.
- The `proptest` round-trip on `Username` and `DrinkTypeName` is the *algebraic invariant* test that `rust.md` § 7 reserves proptest for, and the `prop_assume!(!s.trim().is_empty())` guard correctly handles the trim sanitizer reshape. Good first landing point for the crate.
- The Issue scope listed 5 newtypes plus inline password validation; the PR added `Password` with restricted derives instead — a stronger answer than spec'd, with the rationale documented inline.

PR #148, opened 2026-05-14
120: PR-D3 — DeriveActiveEnum for status / ruleset / drink category / run flag reason

brendanbyrne:
Closes #120.

## Summary

Stream D step 3 (#122 = D1 / IDs, #133 = D2 / strings, this = D3 / enums). Replaces four string-typed enum columns with `sea_orm::DeriveActiveEnum`-backed Rust enums so the entity column type itself enforces the variant set, matches over them are compiler-checked, and call sites stop juggling `.as_str()` / `.to_string()` between Rust and the DB.

### What changed

- New / rewritten enums in `domain/enums.rs`:
  - `SessionStatus` — `Active` / `Closed`. Was already a hand-rolled enum; now goes through `DeriveActiveEnum` so the column carries the type natively. Dropped `as_str` and `FromStr` (no longer used externally — entity round-trips through SeaORM, JSON round-trips through serde).
  - `SessionRuleset` (renamed from `Ruleset`) — `Random`, `Default`, `LeastPlayed`, `RoundRobin`. `FromStr` is preserved for the service-layer parse from the route DTO's `ruleset: String` — see the note about #146 below.
  - `DrinkCategory` — `Alcoholic` / `NonAlcoholic`. Maps `sessions.least_played_drink_category` (TEXT, nullable). The boolean `drink_types.alcoholic` column stays a bool — only two values, no Rust enum benefit there.
  - `RunFlagReason` — six variants (`time_is_incorrect`, `wrong_track`, `wrong_race_setup`, `wrong_drink_type`, `other`, `record_requires_photo_verification`). No write path consumes it yet; defined here so the column is enum-typed end-to-end the moment the flag-creation flow lands.
- Entity columns are now the enum types directly (`entities/sessions.rs`, `entities/run_flags.rs`). Migration is unchanged because SeaORM maps `DeriveActiveEnum` to TEXT at runtime and the columns are already TEXT. **No dev DB reset needed.**
- `SessionDetail` / `SessionSummary` / `SessionSummaryRow` now carry `SessionRuleset` / `SessionStatus` instead of `String`. Wire shape is unchanged thanks to serde's `rename_all` derives matching the `string_value` mappings.
- `test_helpers::insert_session` takes `SessionStatus` instead of `&str`. Every `insert_session(..., \"active\")` / `\"closed\"` callsite now passes the enum variant.

### Behavior change to flag explicitly

[...]

Issue #146, opened 2026-05-13
Typed Path extractor 400s don't match the api-contract error envelope


brendanbyrne:
## Context

When axum's typed `Path<T>` extractor fails to deserialize a URL path segment (e.g. a non-UUID where the route declares `Path<SessionId>`, or a non-integer where the route declares `Path<CupId>`), it returns **400** with a plain-text body from `axum::extract::rejection::PathRejection`. That body doesn't match the JSON envelope required by [`docs/api-contract.md` § 3](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/api-contract.md#3-error-response-contract) (`{ "error": "...", "code": "..." }`), so frontend code can't reliably pattern-match on `code` for this class of failure.

This is **pre-existing**: it was already true for `Path<Uuid>` and `Path<i32>` before [PR #145](https://github.com/brendanbyrne/beerio-kart/pull/145) (PR-D1 / #122). PR-D1 broadens the surface by turning every `Path<String>` / `Path<i32>` into a typed newtype, but the underlying mismatch is the same one and has the same fix.

Surfaced as a 🔵 Suggestion during PR #145 review. Flagged for a follow-up rather than scope creep on D1.

## Scope

- Add a custom rejection handler so `Path<T>` parse failures return the standard envelope. Two reasonable shapes:
  1. A project-local extractor (`extract::Path<T>`) that wraps `axum::extract::Path<T>` and maps `PathRejection` → `error::Error::BadRequest`.
  2. A `WithRejection<Path<T>, ...>` wrapper applied at each route declaration.
  Option 1 is one place to change; option 2 leaves the bare `axum::Path` available. Both work — recommend option 1.
- Add an `invalid_path_param` (or similar) entry to [`api-contract.md` § 8](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/api-contract.md#8-error-code-registry) and wire the new extractor to emit that code.
- Cover Json body deserialization failures in the same pass if the fix shape is general: axum's `JsonRejection` has the same plain-text problem. Probably the same custom-extractor pattern with a sibling `invalid_request_body` code (the existing 422 `unprocessable` may already cover this — re-check before adding a new code).
- Update the two `test_get_nonexistent_*_returns_404` integration tests (currently use a valid-shape UUID to dodge the 400 path) with at least one new test that asserts the **shape** of the 400 envelope.

## Standards refs

[...]

Issue #129, opened 2026-05-12
File-length splits in services/ (PR-G4)


brendanbyrne:
## Context

`services/runs.rs` and `services/sessions.rs` (and possibly others) exceed the 500-line soft cap from `rust.md` § 13. Split by concern (e.g., session lifecycle vs. race orchestration vs. participant management) so files stay browsable and each module's blast radius is local.

## Scope

- Split `services/runs.rs` and `services/sessions.rs` (and any other >500-line non-test files identified in the audit) by concern.
- Tests stay where they are — splitting tests is a separate question.
- Preserve `git mv` history where possible to keep blame useful.

## Standards refs

- [`docs/coding-standards/rust.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/coding-standards/rust.md) § 13

## Dependencies

PR-A2 (audit; landed). The audit confirms which files cross the cap and how — re-read [`docs/designs/2026-04-15-rust-audit.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/designs/2026-04-15-rust-audit.md) before starting to confirm the split boundaries match what the audit suggests.

## Risk

Medium — lots of `pub(crate)` boundary changes; the compiler will catch most issues but ergonomic shifts may surface (e.g., a helper used across the new split needs to move to `services/mod.rs`).

## Effort

M.

## Verification

- Tests pass.
- `git log --follow` on a moved function still produces history (i.e. `git mv` was used, not delete-and-create).
- No file in `backend/src/services/` exceeds 500 non-blank lines after the split.

## Compliance-plan reference

[`docs/compliance-plan.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/compliance-plan.md) § PR-G4.

PR #149, opened 2026-05-14
119: PR-D4 — numeric domain types RaceTimeMs / LapTimeMs

### LINKED ISSUE #119: Numeric domain types: RaceTimeMs / LapTimeMs (PR-D4)
## Context

Fourth and final PR of Stream D. Wraps the four `i32` time columns on `runs` (`track_time`, `lap1_time`, `lap2_time`, `lap3_time`) in domain newtypes so the lap-sum invariant becomes a function over typed values rather than naked integers.

## Scope

- Define `RaceTimeMs(NonZeroI32)` and `LapTimeMs(NonZeroI32)` newtypes via `nutype`.
- Migrate run-recording paths (`services/runs.rs`, run-creation routes) to use the newtypes for `track_time`, `lap1_time`, `lap2_time`, `lap3_time`.
- The "lap times sum to total" invariant becomes a function over typed values, not raw `i32` — e.g. `assert_lap_sum(laps: [LapTimeMs; 3], total: RaceTimeMs) -> Result<(), ValidationError>`.

## Standards refs

- [`docs/coding-standards/rust.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/coding-standards/rust.md) § 2

## Dependencies

PR-D1 (nutype crate is in the workspace; newtype pattern established).

## Risk

Low.

## Effort

S–M.

## Verification

- Existing run-time validation tests pass.
- The function `assert_lap_sum(laps: [LapTimeMs; 3], total: RaceTimeMs)` is the new invariant point — exercised by a unit test that demonstrates both the happy and unhappy paths.

## Compliance-plan reference

[`docs/compliance-plan.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/compliance-plan.md) § PR-D4.

brendanbyrne:
Closes #119. Closes Stream D.

## Summary

Stream D step 4 (final): wrap the four `i32` time columns on `runs` in domain newtypes so the lap-sum invariant becomes a function over typed values rather than naked integers. With this, the recording path can't carry an out-of-range time anywhere downstream — the type proves the value is in `1..=600_000` ms.

### What changed

- New `backend/src/domain/numeric.rs`:
  - `RaceTimeMs(i32)` and `LapTimeMs(i32)` via `nutype`, both with `validate(greater_or_equal = 1, less_or_equal = 600_000)`.
  - `pub const MAX_TIME_MS: i32 = 600_000` next to the types so route DTOs and error messages can spell the cap the same way the constructor enforces it.
  - `assert_lap_sum(laps: [LapTimeMs; 3], total: RaceTimeMs) -> Result<(), Error>` — the single canonical place where the "lap times sum to total" invariant is enforced.
  - 12 unit tests (bounds + lap-sum happy + lap-sum mismatch + off-by-one-ms boundary) plus a proptest that any three lap values in `1..=200_000` round-trip through `assert_lap_sum` cleanly.
- `services/runs.rs::validate_time_fields` rewritten as the boundary between the API's raw `i32`s and the typed values. Returns a `ValidatedRunTimes` struct so `create_run` / `insert_run` don't re-parse — the typed values flow through to the entity `Set(...)` site, where a single `*.as_ref()` per field unwraps back to `i32`.
- Old `const MAX_TRACK_TIME_MS: i32 = 600_000` in `services/runs.rs` is gone; the canonical name is now `domain::MAX_TIME_MS`.
- `domain/mod.rs` re-exports `LapTimeMs`, `MAX_TIME_MS`, `RaceTimeMs`.

### Two intentional deviations from the Issue body

1. **Inner type is `i32`, not `NonZeroI32`.** The Issue sketches `RaceTimeMs(NonZeroI32)`, but `NonZeroI32` buys nothing beyond "non-zero" — we also need an upper bound, which `validate(less_or_equal = 600_000)` covers. Using `i32` lets the newtype match the DB column type with zero conversion at the entity boundary.
2. **Scope is the recording path, not outbound DTOs.** The Issue says "migrate run-recording paths" — done. `RunDetail` / `RaceSubmission` etc. keep `i32` for now, matching the same boundary D2 used for `Username` (the recording side gets typed; outbound conversion is a follow-up if reviewers want it). Keeps the PR S–M as estimated.

### Deferral: CreateRunRequest DTO stays `i32`

Same reason D3 ([#148](https://github.com/brendanbyrne/beerio-kart/pull/148)) kept the ruleset DTO as `String`: switching the DTO to `RaceTimeMs` / `LapTimeMs` would route invalid input through axum's Json extractor and trip the [#146](https://github.com/brendanbyrne/beerio-kart/issues/146) envelope-mismatch. The parse happens at the service boundary instead, preserving the existing 400 + `{ "error": "track_time must be between 1 and 600000 ms" }` shape. Once #146 lands, the typed DTO becomes a one-line follow-up.

### Stream D status after merge

D1 ([#122](https://github.com/brendanbyrne/beerio-kart/issues/122)) + D2 ([#133](https://github.com/brendanbyrne/beerio-kart/issues/133)) + D3 ([#120](https://github.com/brendanbyrne/beerio-kart/issues/120)) + D4 (this PR) = Stream D complete. This unblocks **G4 ([#129](https://github.com/brendanbyrne/beerio-kart/issues/129))** — file-length splits in `services/` — which the handoff said to do after D-stream lands so the split shapes don't churn.

## Test plan

- [ ] `cd backend && cargo check` — passes
- [ ] `cd backend && cargo clippy --all-targets` — passes
- [ ] `cd backend && cargo test` — 252 lib + 19 session_integration + 1 schema_drift + 1 middleware_limits + smaller crates, all green (was 234 before this PR — 18 new tests, of which 12 unit + 1 proptest live in `domain/numeric.rs` and the existing `validate_time_fields` cases continue to pass through the new code path)
- [ ] Spot-check the new module: `cd backend && cargo test domain::numeric::tests` — 12 tests + 1 proptest pass
- [ ] No migration changes; `tests/schema_drift.rs` passes — proves the entity-side stays on `i32` and the column types haven't moved
- [ ] (Manual, optional) Submit a run via the API with `track_time: 0` / `-1` / `600001`: each returns `400 BadRequest` with a message naming the field and the `1..=600000` range — same UX as today, but the bounds are now enforced by `RaceTimeMs::try_from` instead of an open-coded `if`

### DIFF
--- a/backend/src/domain/numeric.rs
+++ b/backend/src/domain/numeric.rs
@@ -1,0 +1,260 @@
+//! Validated numeric newtypes for run times.
+//!
+//! Companion to [`super::strings`]. Where `strings.rs` makes
+//! "unvalidated-string-in-this-arg" a compile error, this module makes
+//! "out-of-range-or-zero-time-in-this-arg" a compile error. Each newtype
+//! wraps an `i32` and runs bound validation in its constructor, so a
+//! downstream caller holding a [`RaceTimeMs`] or [`LapTimeMs`] can trust
+//! the value is in the inclusive range `1..=600_000` (1 ms to ten
+//! minutes) without re-checking. See `coding-standards/rust.md` § 2.
+//!
+//! The `i32` inner type matches the DB column type (`runs.track_time`
+//! and the three `runs.lap*_time` columns are `INTEGER`) and the
+//! entity-stays-primitive boundary rule in
+//! `coding-standards/seaorm.md` § 6. Conversion happens in the service
+//! layer at the boundary — input parsing into typed values via the
+//! `TryFrom<i32>` impl, output unwrap to `i32` via the `AsRef<i32>` impl
+//! at the entity write.
+//!
+//! The lap-sum invariant ("the three lap times must add exactly to the
+//! total race time") is captured by [`assert_lap_sum`], which takes the
+//! typed values and is the single place that contract is enforced.
+
+use nutype::nutype;
+
+use crate::error::Error;
+
+/// Inclusive upper bound (in milliseconds) on a single race time or a
+/// single lap time: 600,000 ms = ten minutes.
+///
+/// Public so route DTOs and error messages can spell the limit out the
+/// same way the constructor enforces it.
+pub const MAX_TIME_MS: i32 = 600_000;
+
+/// Total elapsed time for one full race, in milliseconds.
+///
+/// Constructed via `TryFrom<i32>`: values outside the inclusive range
+/// from `1` up to [`MAX_TIME_MS`] (i.e. zero, negative, or longer than
+/// ten minutes) are rejected by the nutype `validate` machinery.
+#[nutype(
+    validate(greater_or_equal = 1, less_or_equal = 600_000),
+    derive(
+        Debug,
+        Clone,
+        Copy,
+        PartialEq,
+        Eq,
+        PartialOrd,
+        Ord,
+        Hash,
+        Display,
+        AsRef,
+        Serialize,
+        Deserialize,
+        TryFrom,
+    )
+)]
+pub struct RaceTimeMs(i32);
+
+/// Time for a single lap, in milliseconds.
+///
+/// Same bounds as [`RaceTimeMs`]: a single lap can in principle be the
+/// entire race (e.g. the player took the full ten minutes on lap 1), so
+/// the lap upper bound matches the race upper bound. The sum of three
+/// `LapTimeMs` values equals the race's `RaceTimeMs` by construction —
+/// see [`assert_lap_sum`].
+#[nutype(
+    validate(greater_or_equal = 1, less_or_equal = 600_000),
+    derive(
+        Debug,
+        Clone,
+        Copy,
+        PartialEq,
+        Eq,
+        PartialOrd,
+        Ord,
+        Hash,
+        Display,
+        AsRef,
+        Serialize,
+        Deserialize,
+        TryFrom,
+    )
+)]
+pub struct LapTimeMs(i32);
+
+/// Verify that three lap times sum exactly to the total race time.
+///
+/// This is the canonical lap-sum invariant — every run-creation path
+/// funnels through this function so the rule is enforced in exactly one
+/// place. Working over typed values rather than raw `i32`s means an
+/// upstream caller can't accidentally pass garbage: each `LapTimeMs` /
+/// `RaceTimeMs` already lives in `1..=600_000`.
+///
+/// Sum overflow can't happen: three lap times each capped at
+/// [`MAX_TIME_MS`] sum to at most `1_800_000`, comfortably inside
+/// `i32::MAX`.
+///
+/// # Errors
+///
+/// Returns [`Error::BadRequest`] if the laps don't sum to `total`. The
+/// message includes the absolute difference so the client can show the
+/// user which direction to correct.
+pub fn assert_lap_sum(laps: [LapTimeMs; 3], total: RaceTimeMs) -> Result<(), Error> {
+    let sum: i32 = laps.iter().map(|l| *l.as_ref()).sum();
+    let total_inner = *total.as_ref();
+    if sum != total_inner {
+        let diff = (sum - total_inner).abs();
+        return Err(Error::bad_request(format!(
+            "Lap times must add up to total time (off by {diff}ms)"
+        )));
+    }
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use proptest::prelude::*;
+
+    use super::*;
+
+    // ── RaceTimeMs ───────────────────────────────────────────────────
+
+    #[test]
+    fn test_race_time_accepts_typical_two_minute_run() {
+        // 2:00.000 — middle of the road, well inside bounds.
+        assert!(RaceTimeMs::try_from(120_000).is_ok());
+    }
+
+    #[test]
+    fn test_race_time_accepts_one_ms_lower_bound() {
+        // Technically valid; lap-sum check is independent of the bound.
+        assert!(RaceTimeMs::try_from(1).is_ok());
+    }
+
+    #[test]
+    fn test_race_time_accepts_max() {
+        assert!(RaceTimeMs::try_from(MAX_TIME_MS).is_ok());
+    }
+
+    #[test]
+    fn test_race_time_rejects_zero() {
+        assert!(RaceTimeMs::try_from(0).is_err());
+    }
+
+    #[test]
+    fn test_race_time_rejects_negative() {
+        assert!(RaceTimeMs::try_from(-1).is_err());
+    }
+
+    #[test]
+    fn test_race_time_rejects_over_ten_minutes() {
+        assert!(RaceTimeMs::try_from(MAX_TIME_MS + 1).is_err());
+    }
+
+    // ── LapTimeMs ────────────────────────────────────────────────────
+
+    #[test]
+    fn test_lap_time_accepts_typical() {
+        assert!(LapTimeMs::try_from(40_000).is_ok());
+    }
+
+    #[test]
+    fn test_lap_time_rejects_zero() {
+        assert!(LapTimeMs::try_from(0).is_err());
+    }
+
+    #[test]
+    fn test_lap_time_rejects_negative() {
+        assert!(LapTimeMs::try_from(-1).is_err());
+    }
+
+    #[test]
+    fn test_lap_time_rejects_over_max() {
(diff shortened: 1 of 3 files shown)


backend/src/domain/numeric.rs
```diff
@@ -0,0 +1,260 @@
...
+/// Sum overflow can't happen: three lap times each capped at
+/// [`MAX_TIME_MS`] sum to at most `1_800_000`, comfortably inside
+/// `i32::MAX`.
+///
+/// # Errors
+///
+/// Returns [`Error::BadRequest`] if the laps don't sum to `total`. The
+/// message includes the absolute difference so the client can show the
+/// user which direction to correct.
+pub fn assert_lap_sum(laps: [LapTimeMs; 3], total: RaceTimeMs) -> Result<(), Error> {
```
brendanbyrne: 🔵 **Suggestion — visibility.** Per `rust.md` § 3 (default to `pub(crate)` for cross-module items unless they need to cross a crate boundary), this should be `pub(crate)`. It's only used inside the crate — the single import is `crate::domain::numeric::assert_lap_sum` in `services/runs.rs`. The closest sibling — `parse_db_id` in `domain/ids.rs:55` — is `pub(crate)` for the same shape (crate-internal free function in `domain/`). The newtypes correctly stay `pub` because they're re-exported as crate public API; this one isn't.

```rust
pub(crate) fn assert_lap_sum(laps: [LapTimeMs; 3], total: RaceTimeMs) -> Result<(), Error> {
```

backend/src/domain/numeric.rs
```diff
@@ -0,0 +1,260 @@
...
+///
+/// Sum overflow can't happen: three lap times each capped at
+/// [`MAX_TIME_MS`] sum to at most `1_800_000`, comfortably inside
+/// `i32::MAX`.
+///
+/// # Errors
+///
+/// Returns [`Error::BadRequest`] if the laps don't sum to `total`. The
+/// message includes the absolute difference so the client can show the
+/// user which direction to correct.
```
brendanbyrne: 🔵 **Suggestion — docstring vs. behavior.** The docstring says "the message includes the absolute difference so the client can show the user **which direction** to correct." But line 107 computes `let diff = (sum - total_inner).abs();` — `abs()` strips the sign, so the message communicates magnitude only, not direction. The two halves of the sentence disagree.

Two fixes: drop the "which direction to correct" clause (the message really is just magnitude), or compute a signed diff and word the message as "lap sum exceeds total by X" / "is short of total by X". Pre-existing behavior — the `.abs()` was in the old code — but the new docstring makes the gap explicit. Worth resolving while you're touching the lines.

backend/src/services/runs.rs
```diff
@@ -8,8 +8,8 @@ use serde::{Deserialize, Serialize};
 
 use crate::{
     domain::{
-        BodyId, CharacterId, DrinkTypeId, GliderId, RunId, SessionId, SessionRaceId, TrackId,
-        UserId, Username, WheelId,
+        BodyId, CharacterId, DrinkTypeId, GliderId, LapTimeMs, MAX_TIME_MS, RaceTimeMs, RunId,
+        SessionId, SessionRaceId, TrackId, UserId, Username, WheelId, numeric::assert_lap_sum,
```
brendanbyrne: 🔵 **Suggestion — import style.** This block mixes re-exported flat names with one sub-module path (`numeric::assert_lap_sum`). If you adopt the `pub(crate)` suggestion on `assert_lap_sum`, this stays — the sub-module path is the only option. If you keep `pub fn`, re-exporting at the `domain::` level (`pub use numeric::{assert_lap_sum, LapTimeMs, MAX_TIME_MS, RaceTimeMs};` in `domain/mod.rs`) would let the import be flat like the others. Either resolution is fine; the current shape is the awkward middle.

backend/src/services/runs.rs
```diff
@@ -123,33 +120,53 @@ pub struct RunFilters {
...
+///
+/// Parses each `i32` into its typed newtype (enforcing the `1..=MAX_TIME_MS`
+/// bound at construction time), then delegates the lap-sum invariant
+/// (`lap1 + lap2 + lap3 == track_time`) to [`assert_lap_sum`]. The
+/// invariant lives in one place — see `domain/numeric.rs` — and this
+/// function is the boundary that translates `nutype` errors into the
+/// user-facing `BadRequest` messages the API contract expects.
+fn validate_time_fields(body: &CreateRunRequest) -> Result<ValidatedRunTimes, Error> {
+    let track_time = RaceTimeMs::try_from(body.track_time).map_err(|_| {
+        Error::bad_request(format!("track_time must be between 1 and {MAX_TIME_MS} ms"))
```
brendanbyrne: 🔵 **Suggestion — bound symmetry.** The upper bound is wired through `MAX_TIME_MS`; the lower bound `1` is a magic number in the format string (and would silently disagree with the constructor if the `greater_or_equal = 1` on `RaceTimeMs`/`LapTimeMs` ever changed). Cheap to immunize: add `pub const MIN_TIME_MS: i32 = 1;` next to `MAX_TIME_MS` in `domain/numeric.rs` and reference it in both the validators (`validate(greater_or_equal = MIN_TIME_MS, ...)`) and these error messages (lines 151 and 155). Low priority, but it's the symmetric move to what you already did with `MAX_TIME_MS`.

brendanbyrne (commented):
## Review — PR-D4 numeric domain types

**Verdict:** ✅ Approve — four small suggestions; nothing blocks merge.

**What this PR does:** Wraps the four `i32` time columns on `runs` in `RaceTimeMs` / `LapTimeMs` nutype newtypes, moves the lap-sum invariant into a single `assert_lap_sum(laps, total)` function over typed values, and routes the recording path through them via a new `ValidatedRunTimes` boundary struct. Schema and wire format unchanged.

**Areas covered:** `rust.md` (§ 2 type-driven design, § 3 visibility, § 6 documentation, § 7 testing), `seaorm.md` § 6 (entity-stays-primitive boundary), `design.md` / `data-model.md` (no changes required — see below).

### Automated checks

- `cargo fmt --check` — clean (`imports_granularity` / `group_imports` warnings are pre-existing nightly-only config, not file-specific).
- `cargo test domain::numeric` — **17 passed** (12 explicit unit + 1 proptest + 4 auto-generated nutype boundary tests).

### Design-doc accuracy

**No changes required.** Schema unchanged (entity columns stay `i32`, `tests/schema_drift.rs` still passes per the PR plan). Wire format unchanged (`CreateRunRequest.track_time` stays `i32` — the typed DTO is intentionally deferred behind #146 per the PR body and the same call PR-D3 made for `RulesetName`).

`data-model.md:244` describes `track_time: INTEGER (milliseconds, not null, must be positive)` and § Validation lists "track_time must be positive" — the 10-minute upper bound this PR moves into `MAX_TIME_MS` was already enforced by `MAX_TRACK_TIME_MS = 600_000` in `services/runs.rs` before this PR. Data-model.md not mentioning the 10-minute cap is a **pre-existing** gap, not new to this PR. Worth filing a separate small-doc Issue if we want to close it, but it's not blocking here.

### What's good

- **17 tests, well-named.** Names read as behavioral specs (`test_race_time_rejects_zero`, `test_assert_lap_sum_rejects_off_by_one_ms`) per `rust.md` § 7.
- **The off-by-one test** in `numeric.rs` is the kind of regression-anchor § 7 calls for: it exists specifically so a future refactor can't quietly soften equality to `abs(sum-total) <= 1`, and the in-test comment names the threat.
- **The `i32` (not `NonZeroI32`) decision** is justified in the PR body — `validate(greater_or_equal = 1)` already enforces non-zero, and matching the DB column type avoids a conversion at the entity boundary. Deviation-from-Issue properly flagged and reasoned.
- **`ValidatedRunTimes`** is the right shape: validation produces typed values that flow to the entity write, so `insert_run` doesn't re-parse, and the four `Set(*times.X.as_ref())` lines are the entire boundary.
- **Per-lap error labels** ("lap1_time must be between 1 and 600000 ms") are a UX improvement over the old combined messages — the user now knows *which* lap is wrong.
- **`#[allow(clippy::struct_field_names)]`** on `ValidatedRunTimes` carries a load-bearing comment explaining why the `_time` suffix is intentional (aligns with API contract + entity column names).
- **Module-level docstring** at `numeric.rs:1-21` explicitly links to `strings.rs` (the parallel companion), `rust.md` § 2, and `seaorm.md` § 6 — the "why this exists" is one click away.
- **Overflow safety reasoned about** in the `assert_lap_sum` docstring (3 × `MAX_TIME_MS` = 1.8M ≪ `i32::MAX`).

Issue #58, opened 2026-05-06
Race-derived sweeper + 15-min cadence (ADR-0035)


brendanbyrne:
## Status

Reframed 2026-05-16 against [ADR-0035](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/decisions/0035-race-anchored-session-lifetime.md). The Tokio task shape (TaskTracker, `shutdown::supervised`, set-based `Entity::update_many`, `MissedTickBehavior::Skip`) was implemented under PR-F2 of the Hardening compliance plan (PR #152). What's left is the predicate change and cadence adjustment ADR-0035 specifies, plus the removal of `touch_session` (no longer needed once `last_activity_at` is gone).

## Context

ADR-0035 changes the sweeper from "close sessions that have been idle for an hour" to "close sessions whose races have all expired (or that have no race within the bootstrap window)." With the race-derived liveness predicate pushed into every user-facing read path (`get_pending_races` via #51; `check_not_in_any_session` and `get_active_session_id` via #47), the sweeper becomes purely DB hygiene — no user-facing read depends on it having run recently. Cadence relaxes to 15 minutes.

## Acceptance criteria

- [ ] `close_stale_sessions` (`backend/src/services/sessions/lifecycle.rs`): rewrite the predicate per ADR-0035 § Decision outcome. `sessions.status = 'active'` AND `sessions.created_at < NOW() - 1 hour` AND `NOT EXISTS (race in this session created within the last hour)`. Transaction body (mark remaining participants as left, set `status = 'closed'`) unchanged.
- [ ] `helpers::touch_session` (`backend/src/services/helpers.rs`): function deleted. All ~14 call sites across `create_session`, `join_session`, `leave_session`, `next_track`, `skip_turn`, `create_run`, `skip_pending_race` removed.
- [ ] Sweep interval (`backend/src/main.rs`): change from `Duration::from_secs(300)` (5 min) to `Duration::from_secs(900)` (15 min).
- [ ] Tests: drop the `touch_session_updates_last_activity_at` helper test. Add sweeper tests covering: closes when all races expired, keeps session with a recent race, handles the bootstrap case (session created, no races yet — closed after 1 hour).

## Companion work

- #47 — lifecycle changes (schema, `join_session`, the two "are you in a session" check predicates).
- #51 — pending-race derivation (per-race 1-hour expiry).

[...]

PR #152, opened 2026-05-14
124: graceful shutdown via TaskTracker + CancellationToken (PR-F1)

### LINKED ISSUE #124: TaskTracker + CancellationToken + graceful shutdown (PR-F1)
## Context

Foundation for graceful shutdown. Today the server has no shutdown story: SIGTERM kills in-flight requests and orphans any spawned task. This PR introduces the `tokio-util` primitives (`TaskTracker`, `CancellationToken`) and wires them into `axum::serve(...).with_graceful_shutdown(...)`.

## Scope

- Add `tokio-util` to `[dependencies]` (with `task` feature).
- In `main.rs`: create a `CancellationToken` and `TaskTracker` at startup, store both in app state (or pass into background-task spawners).
- Wire `axum::serve(...).with_graceful_shutdown(...)` to a `select!` over `tokio::signal::ctrl_c()` and (on Unix) `SIGTERM`.
- On shutdown: cancel the token, then `tokio::time::timeout(20s, tracker.wait())`.

## Standards refs

- [`docs/coding-standards/tokio.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/coding-standards/tokio.md) § 5, § 13

## Dependencies

None. Unblocks PR-F2.

## Risk

Medium — first time shutdown is exercised end-to-end. Test on Unraid before tagging release.

## Effort

M.

## Verification

- Send SIGTERM during a long-running request; confirm in-flight requests complete (or the 20s timeout triggers cleanly).
- Spawn a no-op tracked task; confirm `tracker.wait()` returns once cancellation propagates.

## Compliance-plan reference

[`docs/compliance-plan.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/compliance-plan.md) § PR-F1.

brendanbyrne:
## Summary

Wire `tokio_util::task::TaskTracker` + `tokio_util::sync::CancellationToken` through `axum::serve(...).with_graceful_shutdown(...)` so SIGTERM (and Ctrl-C in dev) drain in-flight requests **and** background tasks before exit. Closes #124 (PR-F1) and #58 (PR-F2).

Standards refs: [`tokio.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/coding-standards/tokio.md) § 5 (panic-wrapper), § 8 (background-task shape + entry/exit logs), § 13 (signal → cancel → wait); [`seaorm.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/coding-standards/seaorm.md) § 1 (set-based `update_many`).

## Changes

- **`backend/src/shutdown.rs` (new):**
  - `signal(cancel)` — builds the shutdown future for `with_graceful_shutdown`. Races `ctrl_c()` and (on Unix) SIGTERM, then calls `cancel.cancel()`. Returns `io::Result<impl Future<…>>` so a missing SIGTERM-handler capability fails fast at startup instead of from inside the shutdown future.
  - `wait(tracker, timeout)` — `tokio::time::timeout(timeout, tracker.wait())` with clean/timeout logging. Caller closes the tracker before calling (keeps the helper composable).
  - `supervised(name, fut)` — wraps a detached task with `tokio.md` § 8 entry/exit logs and `tokio.md` § 5 `AssertUnwindSafe(...).catch_unwind()` panic capture. Shape is `tracker.spawn(supervised("name", async move { ... }))`.
  - Five unit tests cover empty-tracker, cancel-then-drain, stubborn-task-times-out, supervised-clean-run, supervised-catches-panic paths.
- **`backend/src/main.rs`:** build `CancellationToken` + `TaskTracker` at startup; migrate the two existing inline `tokio::spawn` background tasks (tower-governor cache janitor + stale-session cleanup) to `tracker.spawn(shutdown::supervised(name, async move { ... select! over cancel.cancelled() ... }))`. `tracker.close()` after both spawns. `axum::serve(...).with_graceful_shutdown(shutdown::signal(cancel.clone())?).await?`, then `shutdown::wait(tracker, 20s)`.
- **Stale-session cleanup loop:** switched from `tokio::time::sleep` to `tokio::time::interval` with `MissedTickBehavior::Skip` (`tokio.md` § 8). First tick now fires immediately on startup (cheap on a clean DB; catches up after crash recovery).
- **`backend/src/services/sessions/lifecycle.rs`:** refactored `close_stale_sessions` to the set-based `Entity::update_many()` form (`seaorm.md` § 1 — the exemplar in the doc names this exact cleanup). One `SELECT … id` to capture the stale set, then two `UPDATE` statements (participants `left_at`, sessions `status=closed`) in a single transaction. Replaces the prior SELECT + per-session UPDATE loop. `rows_affected` from the sessions update is the new return value — same semantic as the previous `stale.len() as u64` but without a separate count.
- **`backend/Cargo.toml`:**
  - `tokio-util = { version = "0.7", features = ["rt"] }` — `rt` gates `TaskTracker`; `CancellationToken` is feature-flag-free.
  - `tokio` gains the `signal` feature for `tokio::signal::unix::signal`.
  - `futures-util = { version = "0.3", default-features = false, features = ["std"] }` — for `FutureExt::catch_unwind` in `supervised`. Was a transitive dep already; promoted to direct so the use is visible at the dep graph level.
- **`docs/compliance-plan.md`:** flipped PR-F1 and PR-F2 sign-offs to `[x]`. Refreshed F1's scope text to reflect the actual `rt` feature (Issue body had `task`, which doesn't exist on the crate) and the new `supervised` helper. F2's row notes the only deferred piece of original scope is body extraction into `services::sessions::session_cleanup_loop` (no behavior change — fold into the next sessions/-area PR). Added Document history entries for both.

## Test plan

**Automated (in this PR):**

- [x] `cargo clippy --all-targets` — clean.
- [x] `cargo test` — 327 tests pass, including the five new `shutdown::tests` cases (3 `wait_*` + 2 `supervised_*`). `test_stale_session_cleanup` in `tests/session_integration.rs` confirms the set-based refactor preserves behavior.

**Manual smoke test (verified locally on WSL2):**

```bash
RUST_LOG=beerio_kart=info ./target/debug/beerio-kart &
# wait for "Listening on http://localhost:3000"
kill -TERM <pid>
```

Observed log sequence (no stderr/errors, exit 0):

```
INFO beerio_kart::shutdown: started task="governor cache janitor"
INFO beerio_kart::shutdown: started task="session cleanup task"
INFO beerio_kart: Listening on http://localhost:3000
INFO beerio_kart::shutdown: sigterm received, shutting down
INFO beerio_kart::shutdown: exited cleanly task="governor cache janitor"
INFO beerio_kart::shutdown: exited cleanly task="session cleanup task"
INFO beerio_kart::shutdown: clean shutdown
```

Covers both verification items from #124 (signal arrives via the registered handler, tracked background tasks observe `cancel.cancelled()` and exit, `tracker.wait()` returns within the 20s budget) and the verification from #58 (cleanup tick runs, status flips to `closed`).

**Still pending (Brendan, pre-release):** Unraid end-to-end test under real container-orchestrator SIGTERM. Per Issue #124: "medium risk — first time the shutdown path is exercised end-to-end. Brendan will test on Unraid before tagging a release; merge can happen before that test."

**Reviewer manual run (optional but recommended):**

- [ ] Build: `cd backend && cargo build`
- [ ] Boot a long-running request, e.g. `curl http://localhost:3000/api/v1/hello` with the server backgrounded.
- [ ] `kill -TERM <server-pid>`; confirm the seven-line shutdown log sequence above and exit 0.
- [ ] Repeat with `kill -INT` (Ctrl-C path). Should log `ctrl-c received, shutting down` instead of `sigterm received`.

## Compliance-plan reference

[`docs/compliance-plan.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/compliance-plan.md) § PR-F1 and § PR-F2.

### DIFF
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -14,11 +14,12 @@ use beerio_kart::{
     config::{self, Config},
     db,
     middleware::limits::handle_load_shed_error,
-    routes, services,
+    routes, services, shutdown,
 };
 use migration::{Migrator, MigratorTrait};
 use serde::Serialize;
 use tokio::sync::Semaphore;
+use tokio_util::{sync::CancellationToken, task::TaskTracker};
 use tower::{ServiceBuilder, limit::ConcurrencyLimitLayer, load_shed::LoadShedLayer};
 use tower_governor::{GovernorLayer, governor::GovernorConfigBuilder};
 // `tower::timeout::TimeoutLayer` (cited in tokio.md § 12) produces a service
@@ -95,6 +96,13 @@ async fn main() -> anyhow::Result<()> {
     // is moved into the router.
     let cleanup_db = state.db.clone();
 
+    // Graceful-shutdown plumbing (tokio.md § 13). `cancel` is the propagation
+    // signal — every long-lived background task selects on `cancel.cancelled()`
+    // alongside its work-trigger. `tracker` keeps a handle to those tasks so
+    // we can wait for them to wind down after `axum::serve` returns.
+    let cancel = CancellationToken::new();
+    let tracker = TaskTracker::new();
+
     // Per-peer-IP rate limit. The default `PeerIpKeyExtractor` reads the
     // remote address from the `ConnectInfo<SocketAddr>` extension, so the
     // server below must be served via `into_make_service_with_connect_info`.
@@ -118,12 +126,19 @@ async fn main() -> anyhow::Result<()> {
     // (tokio.md § 8) — same pattern as the stale-session loop below, and
     // a real OS thread is overkill for a once-a-minute janitor.
     let governor_limiter = governor_conf.limiter().clone();
-    tokio::spawn(async move {
-        let mut tick = tokio::time::interval(Duration::from_secs(60));
-        tick.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
-        loop {
-            tick.tick().await;
-            governor_limiter.retain_recent();
+    tracker.spawn({
+        let cancel = cancel.clone();
+        async move {
+            let mut tick = tokio::time::interval(Duration::from_secs(60));
+            tick.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+            loop {
+                tokio::select! {
+                    biased;
+                    () = cancel.cancelled() => break,
+                    _ = tick.tick() => governor_limiter.retain_recent(),
+                }
+            }
+            tracing::info!("governor cache janitor exited");
         }
     });
 
@@ -253,31 +268,62 @@ async fn main() -> anyhow::Result<()> {
         );
 
     // Spawn background task to close stale sessions (no activity for 1 hour).
-    // Runs every 5 minutes.
-    tokio::spawn(async move {
-        loop {
-            tokio::time::sleep(Duration::from_secs(300)).await;
-            match services::sessions::close_stale_sessions(&cleanup_db).await {
-                Ok(0) => {}
-                Ok(n) => tracing::info!("Closed {n} stale session(s)"),
-                Err(_) => tracing::error!("Stale session cleanup failed"),
+    // Runs every 5 minutes. PR-F2 (#58) extracts this body into
+    // `services::sessions::session_cleanup_loop`; for now it lives inline so
+    // F1 stays focused on the shutdown wiring.
+    tracker.spawn({
+        let cancel = cancel.clone();
+        async move {
+            let mut tick = tokio::time::interval(Duration::from_secs(300));
+            tick.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+            loop {
+                tokio::select! {
+                    biased;
+                    () = cancel.cancelled() => break,
+                    _ = tick.tick() => {
+                        match services::sessions::close_stale_sessions(&cleanup_db).await {
+                            Ok(0) => {}
+                            Ok(n) => tracing::info!("Closed {n} stale session(s)"),
+                            Err(_) => tracing::error!("Stale session cleanup failed"),
+                        }
+                    }
+                }
             }
+            tracing::info!("session cleanup task exited");
         }
     });
+
+    // No more tasks are spawned after this; closing lets `tracker.wait()`
+    // return as soon as the spawned tasks finish (tokio.md § 13).
+    tracker.close();
 
     let listener = tokio::net::TcpListener::bind("0.0.0.0:3000")
         .await
         .context("Binding TCP listener on 0.0.0.0:3000")?;
     tracing::info!("Listening on http://localhost:3000");
+
+    // Install signal handlers up front. Any failure surfaces here as an
+    // `io::Error` (typically a missing capability), not from inside the
+    // shutdown future where it would be hard to recover from.
+    let shutdown_signal =
+        shutdown::signal(cancel.clone()).context("Installing shutdown-signal handlers")?;
+
     // `into_make_service_with_connect_info::<SocketAddr>` populates the
     // `ConnectInfo<SocketAddr>` request extension that `tower-governor`'s
     // `PeerIpKeyExtractor` reads to bucket rate-limit counters per IP.
+    //
+    // `with_graceful_shutdown` waits for in-flight requests to finish once
+    // the shutdown future resolves — no new connections accepted, existing
+    // ones drain. Background tasks drain via the cancellation token below.
     axum::serve(
         listener,
         app.into_make_service_with_connect_info::<SocketAddr>(),
     )
+    .with_graceful_shutdown(shutdown_signal)
     .await
     .context("HTTP server returned an error")?;
+
+    shutdown::wait(tracker, Duration::from_secs(20)).await;
 
     Ok(())
 }
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -275,21 +275,18 @@ async fn main() -> anyhow::Result<()> {
         );
 
     // Spawn background task to close stale sessions (no activity for 1 hour).
-    // Runs every 5 minutes. PR-F2 (#58) extracts this body into
-    // `services::sessions::session_cleanup_loop` and wraps it in
-    // `shutdown::supervised` like the governor janitor above; for now it
-    // lives inline (with manual entry/exit logs) so F1 stays focused on
-    // the shutdown wiring.
+    // Runs every 5 minutes. PR-F2 (#58) extracts the closure body into
+    // `services::sessions::session_cleanup_loop`; the `shutdown::supervised`
+    // wrapping stays, the body moves.
     //
     // Shutdown-safe (tokio.md § 13 final rule): `close_stale_sessions` is
     // idempotent — it flips `is_active` to `false` on sessions matching
     // `is_active AND last_activity_at < cutoff`. A future dropped between
     // the read and the write leaves the row in its original active state;
     // the next cycle picks it up. No partial-write window.
-    tracker.spawn({
+    tracker.spawn(shutdown::supervised("session cleanup task", {
         let cancel = cancel.clone();
         async move {
-            tracing::info!("session cleanup task started");
             let mut tick = tokio::time::interval(Duration::from_secs(300));
             tick.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
             loop {
@@ -305,9 +302,8 @@ async fn main() -> anyhow::Result<()> {
                     }
                 }
             }
-            tracing::info!("session cleanup task exited");
         }
-    });
+    }));
 
     // No more tasks are spawned after this; closing lets `tracker.wait()`
     // return as soon as the spawned tasks finish (tokio.md § 13).
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -125,8 +125,16 @@ async fn main() -> anyhow::Result<()> {
     // README. Uses `tokio::time::interval` with `MissedTickBehavior::Skip`
     // (tokio.md § 8) — same pattern as the stale-session loop below, and
     // a real OS thread is overkill for a once-a-minute janitor.
+    //
+    // Shutdown-safe (tokio.md § 13 final rule): `retain_recent()` is pure
+    // in-memory; a future dropped mid-call leaves no externally-visible
+    // state. Worst case on shutdown timeout: the rate-limit map keeps
+    // stale entries for a few extra seconds — irrelevant in a restart.
+    //
+    // Wrapped in `shutdown::supervised` so § 8 entry/exit logs and § 5
+    // panic capture are uniform across detached tasks.
     let governor_limiter = governor_conf.limiter().clone();
-    tracker.spawn({
+    tracker.spawn(shutdown::supervised("governor cache janitor", {
         let cancel = cancel.clone();
         async move {
             let mut tick = tokio::time::interval(Duration::from_secs(60));
@@ -138,9 +146,8 @@ async fn main() -> anyhow::Result<()> {
                     _ = tick.tick() => governor_limiter.retain_recent(),
                 }
             }
-            tracing::info!("governor cache janitor exited");
         }
-    });
+    }));
 
     // STATIC_DIR defaults to ../frontend/dist for local dev (running from backend/).
     // In Docker, set to /app/static where the built frontend is copied.
@@ -269,11 +276,20 @@ async fn main() -> anyhow::Result<()> {
 
     // Spawn background task to close stale sessions (no activity for 1 hour).
     // Runs every 5 minutes. PR-F2 (#58) extracts this body into
-    // `services::sessions::session_cleanup_loop`; for now it lives inline so
-    // F1 stays focused on the shutdown wiring.
+    // `services::sessions::session_cleanup_loop` and wraps it in
+    // `shutdown::supervised` like the governor janitor above; for now it
+    // lives inline (with manual entry/exit logs) so F1 stays focused on
+    // the shutdown wiring.
+    //
+    // Shutdown-safe (tokio.md § 13 final rule): `close_stale_sessions` is
+    // idempotent — it flips `is_active` to `false` on sessions matching
+    // `is_active AND last_activity_at < cutoff`. A future dropped between
+    // the read and the write leaves the row in its original active state;
+    // the next cycle picks it up. No partial-write window.
     tracker.spawn({
         let cancel = cancel.clone();
         async move {
+            tracing::info!("session cleanup task started");
             let mut tick = tokio::time::interval(Duration::from_secs(300));
             tick.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
             loop {
@@ -314,7 +330,8 @@ async fn main() -> anyhow::Result<()> {
     //
     // `with_graceful_shutdown` waits for in-flight requests to finish once
     // the shutdown future resolves — no new connections accepted, existing
-    // ones drain. Background tasks drain via the cancellation token below.
+    // ones drain. Background tasks drain via the cancellation token that
+    // `shutdown_signal` (constructed above) triggers when the signal arrives.
     axum::serve(
         listener,
         app.into_make_service_with_connect_info::<SocketAddr>(),
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -103,6 +103,16 @@ async fn main() -> anyhow::Result<()> {
     let cancel = CancellationToken::new();
     let tracker = TaskTracker::new();
 
+    // Install signal handlers up front. Any failure surfaces here as an
+    // `io::Error` (typically a missing capability) — before any background
+    // task is detached, so `main` bails through the normal error path
+    // without leaving orphans for the runtime to abort. On Unix, the
+    // SIGTERM handler is registered with the OS as soon as
+    // `signal::unix::signal` returns, so a SIGTERM that races startup is
+    // queued and observed once the shutdown future starts polling.
+    let shutdown_signal =
+        shutdown::signal(cancel.clone()).context("Installing shutdown-signal handlers")?;
+
     // Per-peer-IP rate limit. The default `PeerIpKeyExtractor` reads the
     // remote address from the `ConnectInfo<SocketAddr>` extension, so the
     // server below must be served via `into_make_service_with_connect_info`.
@@ -279,11 +289,11 @@ async fn main() -> anyhow::Result<()> {
     // `services::sessions::session_cleanup_loop`; the `shutdown::supervised`
     // wrapping stays, the body moves.
     //
-    // Shutdown-safe (tokio.md § 13 final rule): `close_stale_sessions` is
-    // idempotent — it flips `is_active` to `false` on sessions matching
-    // `is_active AND last_activity_at < cutoff`. A future dropped between
-    // the read and the write leaves the row in its original active state;
-    // the next cycle picks it up. No partial-write window.
+    // Shutdown-safe (tokio.md § 13 final rule): `close_stale_sessions` runs
+    // SELECT + two UPDATEs inside one transaction that only commits at the
+    // end. A future dropped before the commit rolls the txn back, leaving
+    // every row in its original Active state; the next cycle re-finds the
+    // same stale set. No partial-write window.
     tracker.spawn(shutdown::supervised("session cleanup task", {
         let cancel = cancel.clone();
         async move {
@@ -313,12 +323,6 @@ async fn main() -> anyhow::Result<()> {
         .await
         .context("Binding TCP listener on 0.0.0.0:3000")?;
     tracing::info!("Listening on http://localhost:3000");
-
-    // Install signal handlers up front. Any failure surfaces here as an
-    // `io::Error` (typically a missing capability), not from inside the
-    // shutdown future where it would be hard to recover from.
-    let shutdown_signal =
-        shutdown::signal(cancel.clone()).context("Installing shutdown-signal handlers")?;
 
     // `into_make_service_with_connect_info::<SocketAddr>` populates the
     // `ConnectInfo<SocketAddr>` request extension that `tower-governor`'s
--- a/backend/src/shutdown.rs
+++ b/backend/src/shutdown.rs
@@ -1,0 +1,125 @@
+//! Graceful-shutdown wiring for the HTTP server and background tasks.
+//!
+//! See `coding-standards/tokio.md` § 5 (`TaskTracker` for long-lived tasks),
+//! § 8 (background task shape), and § 13 (the canonical
+//! signal → cancel → wait sequence).
+
+use std::time::Duration;
+
+use tokio_util::{sync::CancellationToken, task::TaskTracker};
+
+/// Build the shutdown-signal future for `axum::serve(...).with_graceful_shutdown(...)`.
+///
+/// Resolves when Ctrl-C **or** SIGTERM (Unix) is received, then calls
+/// [`CancellationToken::cancel`] so every tracked background task can wind
+/// down via its own `cancel.cancelled()` branch.
+///
+/// SIGTERM-handler installation can fail; that's surfaced as an `io::Error`
+/// from this constructor rather than from inside the returned future.
+/// `with_graceful_shutdown` only accepts `impl Future<Output = ()>`, so
+/// failing fast at install time keeps `main`'s error path normal.
+///
+/// # Errors
+///
+/// Returns the OS error from `tokio::signal::unix::signal` on Unix if the
+/// SIGTERM handler cannot be installed (typically a missing capability or a
+/// resource limit). On non-Unix platforms this is currently infallible.
+pub fn signal(
+    cancel: CancellationToken,
+) -> std::io::Result<impl std::future::Future<Output = ()> + Send + 'static> {
+    #[cfg(unix)]
+    let mut sigterm = tokio::signal::unix::signal(tokio::signal::unix::SignalKind::terminate())?;
+
+    Ok(async move {
+        #[cfg(unix)]
+        tokio::select! {
+            res = tokio::signal::ctrl_c() => match res {
+                Ok(()) => tracing::info!("ctrl-c received, shutting down"),
+                Err(e) => tracing::error!(?e, "ctrl-c handler error, shutting down anyway"),
+            },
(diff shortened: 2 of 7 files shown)


backend/src/main.rs
```diff
@@ -253,32 +268,63 @@ async fn main() -> anyhow::Result<()> {
...
+    let shutdown_signal =
+        shutdown::signal(cancel.clone()).context("Installing shutdown-signal handlers")?;
+
     // `into_make_service_with_connect_info::<SocketAddr>` populates the
     // `ConnectInfo<SocketAddr>` request extension that `tower-governor`'s
     // `PeerIpKeyExtractor` reads to bucket rate-limit counters per IP.
+    //
+    // `with_graceful_shutdown` waits for in-flight requests to finish once
+    // the shutdown future resolves — no new connections accepted, existing
+    // ones drain. Background tasks drain via the cancellation token below.
```
brendanbyrne: 🔵 **Suggestion** — "Background tasks drain via the cancellation token **below**" is misleading. The cancel is fired inside `shutdown_signal`, which is constructed *above* (line 309). `shutdown::wait` below only `.await`s the tracker — it doesn't cancel. Flip `below` → `above`, or rephrase, e.g.:

```rust
    // `with_graceful_shutdown` waits for in-flight requests to finish once
    // the shutdown future resolves — no new connections accepted, existing
    // ones drain. Background tasks drain via the cancellation token that
    // `shutdown_signal` triggers when the signal arrives.
```

backend/src/main.rs
```diff
@@ -118,12 +126,19 @@ async fn main() -> anyhow::Result<()> {
...
+            loop {
+                tokio::select! {
+                    biased;
+                    () = cancel.cancelled() => break,
+                    _ = tick.tick() => governor_limiter.retain_recent(),
+                }
+            }
+            tracing::info!("governor cache janitor exited");
         }
     });
```
brendanbyrne: 🔵 **Suggestion** — Per `tokio.md` § 13 final rule ("Document, per task, what 'in-flight work' means on shutdown ... whether the work is idempotent or transactional"), this spawn site is missing an inline shutdown-semantics note. `governor_limiter.retain_recent()` is pure in-memory; dropping the future mid-call is a no-op. A one-line comment would close the rule:

```rust
    // Shutdown-safe: `retain_recent()` is pure in-memory; a dropped future
    // leaves no externally-visible state.
```

Same applies to the session-cleanup spawn (see comment on its body); that one's idempotent rather than purely in-memory.

backend/src/main.rs
```diff
@@ -118,12 +126,19 @@ async fn main() -> anyhow::Result<()> {
     // (tokio.md § 8) — same pattern as the stale-session loop below, and
     // a real OS thread is overkill for a once-a-minute janitor.
     let governor_limiter = governor_conf.limiter().clone();
-    tokio::spawn(async move {
-        let mut tick = tokio::time::interval(Duration::from_secs(60));
-        tick.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
-        loop {
-            tick.tick().await;
-            governor_limiter.retain_recent();
+    tracker.spawn({
```
brendanbyrne: 🔵 **Suggestion** — Per `tokio.md` § 5, "When you must detach a task, spawn a wrapper that logs panics and errors." A panic inside this closure (or the session-cleanup closure below) is silently captured by the runtime — `tracker.wait()` returns when the task ends regardless, and you find out at 3 a.m. that the limiter cache has been unbounded for a week.

The session-cleanup body is moving to `services::sessions::session_cleanup_loop` in PR-F2 (#58), so the wrapper is natural to add there. The governor janitor stays inline indefinitely, so it's worth wrapping in this PR. The standard `spawn_supervised` shim from `tokio.md` § 5 (or an inline `AssertUnwindSafe(...).catch_unwind()` around the loop) gets you the missing panic log.

Not a blocker — flagging because § 5 is unambiguous and the original `tokio::spawn` calls also lacked the wrapper, so the gap doesn't get filled if it isn't filled here.

backend/src/main.rs
```diff
@@ -118,12 +126,19 @@ async fn main() -> anyhow::Result<()> {
...
+            let mut tick = tokio::time::interval(Duration::from_secs(60));
+            tick.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+            loop {
+                tokio::select! {
+                    biased;
+                    () = cancel.cancelled() => break,
+                    _ = tick.tick() => governor_limiter.retain_recent(),
+                }
+            }
+            tracing::info!("governor cache janitor exited");
```
brendanbyrne: 🔵 **Suggestion** — `tokio.md` § 8: "Always log on background-task entry and exit, with the task name." The PR adds exit logs (good) but no entry logs. The § 8 example itself only shows the exit log, so the rule and the example don't fully agree — but for strict compliance, add a one-liner at the top of each closure body:

```rust
    tracing::info!("governor cache janitor started");
    let mut tick = ...
```

Same at the start of the session-cleanup closure ("session cleanup task started").

brendanbyrne (commented):
## Review summary

**Verdict:** lgtm — well-aligned with `tokio.md` § 5, § 8, § 13. Four 🔵 suggestions inline; none block merge.

**Areas covered:** `tokio.md` (primary), `rust.md`, `compliance-plan.md`.

**Automated checks (from `backend/`):**
- `cargo test` — all suites pass; includes the 3 new `shutdown::tests` cases. Total count is closer to 335 than the 327 named in the PR body — minor drift, not a finding.

**Design-doc accuracy:** No `design.md` changes required — shutdown semantics are owned by `coding-standards/tokio.md` § 13, which already exists and is what the PR implements.

**What's good:**
- Mirrors the `tokio.md` § 13 canonical shape exactly: `signal → cancel → wait` with the 20s budget, `MissedTickBehavior::Skip` on the new `interval`, `biased;` on the cancel branch, `TaskTracker::close()` before `wait()`.
- Returning `io::Result<impl Future>` from `signal()` to fail fast at startup is the right escape hatch given `with_graceful_shutdown`'s `impl Future<Output = ()>` signature, and the `///` documents *why*.
- `wait()` tests are targeted: empty-tracker, cancel-observed, stubborn-timeout. The stubborn case aborts the dangling handle at the end, keeping the runtime tidy.
- Cargo.toml entries carry rationale comments (`rt` gates `TaskTracker`; `signal` for `tokio::signal::unix`).
- The "PR-F2 (#58) extracts this body" inline comment makes the deferred work explicit and re-litigatable.

**Compliance-plan reference:** `docs/compliance-plan.md` § Stream F → PR-F1 sign-off checkbox flips on merge.

brendanbyrne:
Addressed all four 🔵 suggestions in 77de4ba:

1. **`below` → `above` typo:** rephrased to "the cancellation token that `shutdown_signal` (constructed above) triggers when the signal arrives." — matches the actual call order.
2. **Shutdown-safety notes per § 13 final rule:** inline comment on each `tracker.spawn` site. Governor: pure in-memory, dropped future is a no-op. Session cleanup: idempotent — `UPDATE … WHERE is_active AND last_activity_at < cutoff` is replayable; a drop between read and write leaves the row in the original active state and the next cycle picks it up, no partial-write window.
3. **`spawn_supervised` panic wrapper per § 5:** new `shutdown::supervised(name, fut)` helper. Wraps a detached future with § 8 entry/exit logs and § 5 `AssertUnwindSafe(...).catch_unwind()`. Two new unit tests (clean path + panic path). Governor janitor uses it now. Session cleanup keeps the manual entry/exit logs for this PR; PR-F2 (#58) will move it under `supervised` when the body extracts to `services::sessions::session_cleanup_loop` — `supervised` is shaped for that composition (`tracker.spawn(shutdown::supervised("session cleanup", session_cleanup_loop(db, cancel)))`).
4. **Entry log per § 8:** governor janitor gets it from `supervised` (`started task="governor cache janitor"`); session cleanup gets a manual `"session cleanup task started"` line that PR-F2 will replace with `supervised`.

New dep: `futures-util = { version = "0.3", default-features = false, features = ["std"] }` for `FutureExt::catch_unwind`. Already present transitively (via `tokio-util`, `tower-http`), promoted to direct so the use is visible at the dep graph level.

**Smoke test re-run:** SIGTERM produces

```
INFO beerio_kart::shutdown: started task="governor cache janitor"
INFO beerio_kart: session cleanup task started
INFO beerio_kart: Listening on http://localhost:3000
INFO beerio_kart::shutdown: sigterm received, shutting down
INFO beerio_kart: session cleanup task exited
INFO beerio_kart::shutdown: exited cleanly task="governor cache janitor"
INFO beerio_kart::shutdown: clean shutdown
```

All checks: `cargo fmt --check`, `cargo clippy --all-targets`, `cargo test` (5 shutdown unit tests now, 256 lib tests + the rest of the suite) — clean.

backend/src/shutdown.rs
```diff
@@ -0,0 +1,174 @@
...
+}
+
+/// Wait for all [`TaskTracker`]-spawned tasks to wind down, with a hard timeout.
+///
+/// Call this *after* `axum::serve(...).with_graceful_shutdown(...)` returns —
+/// the cancellation token has been triggered by then, so each tracked
+/// background task should observe its `cancel.cancelled()` branch and exit.
+/// The caller is responsible for `tracker.close()`-ing before calling this;
+/// `wait()` doesn't close so it stays composable.
+/// Wrap a background task with entry/exit logs and panic capture.
```
brendanbyrne, line 70: 🟡 **Important** — doc-comment regression. `supervised`'s `///` block was inserted directly after `wait()`'s existing `///` block with no blank line between them. Per Rust's doc-comment grouping rule, consecutive `///` lines form a single block attached to the next item — so right now:

- The merged 21-line block (lines 52-72) attaches to `pub async fn supervised` at line 73.
- `pub async fn wait` at line 85 has **no doc comment** — violates `rust.md` § 6 first rule ("Every `pub` ... item gets a `///` doc comment whose first line is one sentence ending in a period").
- `cargo doc` renders `supervised`'s summary as *"Wait for all `TaskTracker`-spawned tasks to wind down, with a hard timeout."* — actively misleading.

Not caught by clippy because `missing_docs` isn't enabled in the lint config. `rust.md` § 6 last rule — "When a PR modifies a file, re-review the doc comments on every public item in that file" — is the rule the regression slipped past.

**Fix:** move `supervised` (with its doc block intact) below `wait` so the file shape becomes:

```rust
/// signal(...) ...
pub fn signal(...) -> ... { ... }

/// Wait for all [`TaskTracker`]-spawned tasks to wind down, with a hard timeout.
/// ...
pub async fn wait(tracker: TaskTracker, timeout: Duration) { ... }

/// Wrap a background task with entry/exit logs and panic capture.
/// ...
pub async fn supervised<F>(name: &'static str, fut: F)
where F: Future<Output = ()> + Send + 'static
{ ... }
```

Reorder is cleaner than "add a blank line" because `signal → wait → supervised` reads as primary primitives first, helpers last.

backend/src/main.rs
```diff
@@ -253,32 +275,73 @@ async fn main() -> anyhow::Result<()> {
...
-            tokio::time::sleep(Duration::from_secs(300)).await;
-            match services::sessions::close_stale_sessions(&cleanup_db).await {
-                Ok(0) => {}
-                Ok(n) => tracing::info!("Closed {n} stale session(s)"),
-                Err(_) => tracing::error!("Stale session cleanup failed"),
+    // Runs every 5 minutes. PR-F2 (#58) extracts this body into
+    // `services::sessions::session_cleanup_loop` and wraps it in
+    // `shutdown::supervised` like the governor janitor above; for now it
+    // lives inline (with manual entry/exit logs) so F1 stays focused on
+    // the shutdown wiring.
```
brendanbyrne: 🔵 **Suggestion** — the deferred-to-PR-F2 rationale ("F1 stays focused on the shutdown wiring") was strong when `supervised` didn't exist yet. Now that it does and the governor janitor right above already uses it, wrapping the session-cleanup spawn the same way is a one-line change:

```rust
    tracker.spawn(shutdown::supervised("session cleanup task", {
        let cancel = cancel.clone();
        async move {
            let mut tick = ...
            ...
        }
    }));
```

(Drop the manual `"session cleanup task started"` log and the trailing `"session cleanup task exited"` — `supervised` emits both.)

Gives the session task panic safety today rather than at PR-F2 merge time, and saves PR-F2 from re-doing the wrapping. Either's fine — flagging for awareness, not blocking.

brendanbyrne (commented):
## Re-review (HEAD `77de4ba`)

**Verdict:** lgtm — one 🟡 inline (doc-block regression, clean fix), one optional 🔵. The original 4 🔵 suggestions are fully resolved.

**Automated checks (from `backend/`):**
- `cargo test` — all suites pass; lib count went 256 → 258 (the two new `supervised_*` tests).

**Resolution of prior review:**

| Prior finding | Resolution |
|---|---|
| 🔵 "cancellation token below" wording | ✅ Fixed at `main.rs:333` — now reads "via the cancellation token that `shutdown_signal` (constructed above) triggers" |
| 🔵 Per-task shutdown-safety notes | ✅ Both sites carry the note; governor janitor as pure in-memory, session-cleanup as idempotent set-based UPDATE |
| 🔵 `spawn_supervised` wrapper | ✅ Added `shutdown::supervised` with entry/exit logs + `AssertUnwindSafe(fut).catch_unwind()`. Wraps the governor janitor; session-cleanup explicitly deferred to PR-F2 (#58) per inline comment |
| 🔵 Entry log per task | ✅ Automatic for governor janitor via `supervised`; manual `"session cleanup task started"` log at `main.rs:292` |

The `supervised` helper itself is well-shaped: `pub async fn supervised<F>(name: &'static str, fut: F)` with `Send + 'static` bounds, structured `task = name` fields on every log emission, and two unit tests (clean completion + panic capture) covering both branches.

brendanbyrne:
Addressed both in c0a4f2d:

**🟡 Doc-block regression (`shutdown.rs:52-72`)** — fair catch, real bug. Reordered the file to `signal → wait → supervised` so each function gets its own `///` block again. Primary primitives first, helper at the bottom — matches the natural reading order better than \"add a blank line\" would. Verified with `RUSTDOCFLAGS='-A warnings' cargo doc --no-deps --document-private-items`:

- `target/doc/beerio_kart/shutdown/fn.wait.html` now contains \"Wait for all…\".
- `target/doc/beerio_kart/shutdown/fn.supervised.html` contains \"Wrap a background task…\".

Worth filing a follow-up to enable `missing_docs` in the lint config — it would've caught this. Want me to file that, or are there reasons it's deliberately off?

**🔵 Wrap session cleanup with `supervised`** — agreed, the original \"F1 stays focused\" deferral was load-bearing only until `supervised` itself landed in this PR. With the governor janitor right above it using the helper, leaving the session cleanup unwrapped was asymmetry, not focus. Wrapped it. PR-F2 (#58) now only needs to extract the body to `services::sessions::session_cleanup_loop`; the wrapping stays. Dropped the manual `"session cleanup task started"` / `"session cleanup task exited"` lines — `supervised` emits both.

**Smoke test (HEAD c0a4f2d):**

```
INFO beerio_kart::shutdown: started task=\"governor cache janitor\"
INFO beerio_kart::shutdown: started task=\"session cleanup task\"
INFO beerio_kart: Listening on http://localhost:3000
INFO beerio_kart::shutdown: sigterm received, shutting down
INFO beerio_kart::shutdown: exited cleanly task=\"governor cache janitor\"
INFO beerio_kart::shutdown: exited cleanly task=\"session cleanup task\"
INFO beerio_kart::shutdown: clean shutdown
```

Both tasks now log uniformly. `cargo fmt --check`, `cargo clippy --all-targets`, `cargo test` — clean.

brendanbyrne:
Pushed 9976c75 — addresses your \"might as well\" on F2 bullet 3 and the missing-docs follow-up:

**PR-F2 bullet 3 — `close_stale_sessions` is now set-based.** Replaced the SELECT + per-session UPDATE loop with one SELECT (for ids) + two `Entity::update_many()` statements in one transaction (participants `left_at`, sessions `status`). Returns `rows_affected` instead of a separately-counted `stale.len()`. `seaorm.md` § 1's doc literally names this cleanup as its set-based-update exemplar — implementation now matches the documentation's worked example. `test_stale_session_cleanup` + the two `lifecycle.rs` unit tests confirm behavior is preserved.

**Compliance-plan updates.** PR-F1 and PR-F2 both flipped to `[x]` with PR #152 references. PR-F1's scope text refreshed to reflect the actual `rt` feature (the Issue body said `task`, which doesn't exist on `tokio-util`) and to mention the new `shutdown::supervised` helper. PR-F2's row notes that the only piece of the Issue's original scope deferred is body extraction into `services::sessions::session_cleanup_loop` — pure refactor, no behavior change, folds into the next sessions/-area PR rather than spawning a tracking Issue. Document history entries added for both.

**`missing_docs` follow-up filed:** [#153](https://github.com/brendanbyrne/beerio-kart/issues/153) under the `Hardening: Backend compliance plan` milestone. Body cites this PR's regression as the motivating example.

PR body updated to add `Closes #58` and to describe the new commit.

backend/src/main.rs
```diff
@@ -253,32 +275,69 @@ async fn main() -> anyhow::Result<()> {
...
+    //
+    // Shutdown-safe (tokio.md § 13 final rule): `close_stale_sessions` is
+    // idempotent — it flips `is_active` to `false` on sessions matching
+    // `is_active AND last_activity_at < cutoff`. A future dropped between
+    // the read and the write leaves the row in its original active state;
+    // the next cycle picks it up. No partial-write window.
+    tracker.spawn(shutdown::supervised("session cleanup task", {
+        let cancel = cancel.clone();
+        async move {
+            let mut tick = tokio::time::interval(Duration::from_secs(300));
```
brendanbyrne: 🟡 **Important — stale column reference.** The comment claims `close_stale_sessions` "flips `is_active` to `false` on sessions matching `is_active AND last_activity_at < cutoff`", but the schema has no `is_active` column — `sessions` has `status: SessionStatus` (Active/Closed), and `lifecycle.rs:484` / `lifecycle.rs:510` filter on `Status::Active` and set `Status::Closed`. The idempotence claim itself is still correct (the txn wraps SELECT + both UPDATEs with a single commit, so a dropped future rolls back cleanly) — only the column name is wrong. Per `rust.md` § 6: "Stale docs are worse than no docs — they actively mislead."

Suggested rewrite:
```rust
    // Shutdown-safe (tokio.md § 13 final rule): `close_stale_sessions` runs
    // SELECT + two UPDATEs inside one transaction that only commits at the
    // end. A future dropped before the commit rolls the txn back, leaving
    // every row in its original Active state; the next cycle re-finds the
    // same stale set. No partial-write window.
```

backend/src/shutdown.rs
```diff
@@ -0,0 +1,174 @@
...
+mod tests {
+    use std::sync::{
+        Arc,
+        atomic::{AtomicBool, Ordering},
+    };
+
+    use super::*;
+
+    #[tokio::test]
+    async fn wait_returns_immediately_for_closed_empty_tracker() {
```
brendanbyrne: 🔵 **Suggestion — test naming inconsistency.** The five new tests here (`wait_returns_immediately_for_closed_empty_tracker`, `wait_returns_after_tracked_task_observes_cancel`, `supervised_runs_clean_future_to_completion`, `supervised_catches_panic_and_returns`, `wait_times_out_for_stubborn_task`) drop the `test_` prefix used everywhere else in the repo (e.g. `test_stale_session_cleanup` in `tests/session_integration.rs`, all the `test_*` tests in `lifecycle.rs`).

`rust.md` § 7 doesn't *require* the prefix — the rule is "test names are sentences" — but the rule's own example (`test_login_with_wrong_password_returns_401`) carries it, and every sibling test in the repo does too. Adding `test_` keeps `cargo test test_` grep-discoverable and matches the project convention.

Fix: rename each to `test_<existing_name>`.

backend/src/main.rs
```diff
@@ -253,32 +275,69 @@ async fn main() -> anyhow::Result<()> {
...
     let listener = tokio::net::TcpListener::bind("0.0.0.0:3000")
         .await
         .context("Binding TCP listener on 0.0.0.0:3000")?;
     tracing::info!("Listening on http://localhost:3000");
+
+    // Install signal handlers up front. Any failure surfaces here as an
+    // `io::Error` (typically a missing capability), not from inside the
+    // shutdown future where it would be hard to recover from.
+    let shutdown_signal =
+        shutdown::signal(cancel.clone()).context("Installing shutdown-signal handlers")?;
```
brendanbyrne: 🔵 **Suggestion — install signal handlers before spawning background tasks.** Current order in this file: spawn governor janitor (line 137) → spawn session cleanup (line 287) → `tracker.close()` (line 310) → bind listener → `shutdown::signal()` install (here) → serve.

If `shutdown::signal(...)?` returns the documented `io::Error` from `tokio::signal::unix::signal`, `main` bails out with two tasks already detached. They get aborted on runtime drop, which works, but skips the orderly cancel-then-wait path the rest of this module exists to provide. Moving the install above the spawn sites would fail-fast before anything is detached, mirroring how the file already moves `Config::from_env()?` and `db::connect(...)?` ahead of any spawn points.

Low priority — SIGTERM-handler install failures are vanishingly rare in practice — but it's a free structural tightening if you're already pushing a follow-up commit.

brendanbyrne (commented):
## Verdict: Request Changes

One 🟡, two 🔵 — no logic issues, small doc/structural fixes.

## What this PR does

Wires `tokio_util::TaskTracker` + `CancellationToken` through `axum::serve(...).with_graceful_shutdown(...)` so SIGTERM/Ctrl-C drain in-flight requests and tracked background tasks. Adds `backend/src/shutdown.rs` with `signal()` / `wait()` / `supervised()` helpers; migrates both existing detached tasks (tower-governor janitor, stale-session cleanup) onto the tracker. Refactors `close_stale_sessions` to set-based `update_many()` per `seaorm.md` § 1.

## Areas covered

`tokio.md` (primary), `seaorm.md`, `rust.md`. `api-contract.md` and `design.md` don't apply — no wire-format changes and `design.md` doesn't describe shutdown/lifecycle.

## What was verified

- `cargo test` — 335 passed, 0 failed, 1 ignored.

## Design-doc accuracy

No changes required. `docs/design.md` doesn't describe graceful shutdown / TaskTracker / SIGTERM. `docs/compliance-plan.md` PR-F1 and PR-F2 sign-offs flipped to `[x]`, F1's scope text was refreshed (the Issue body's `task` feature on `tokio-util` was wrong — `rt` is correct), and Document history has matching entries. Accurate.

## What's good

- The `close_stale_sessions` refactor is the exact exemplar named in `seaorm.md` § 1, and the doc-comment cites the rule.
- `supervised()`'s `AssertUnwindSafe(...).catch_unwind()` pattern matches `tokio.md` § 5's prescription verbatim.
- Both background tasks use the canonical `biased; cancel.cancelled() | tick.tick()` shape (`tokio.md` § 6 / § 8) with `MissedTickBehavior::Skip`.
- `signal()` failing fast at *install* time instead of from inside the shutdown future is a thoughtful API choice — `with_graceful_shutdown` only accepts `Future<Output = ()>`, so surfacing `io::Error` in `main`'s normal error path is much easier to debug than a vanished future.
- Five new unit tests cover the obvious paths and use `Duration` budgets that distinguish "returned" from "timed out" (the comment in `wait_returns_immediately_for_closed_empty_tracker` explicitly calls this out).
- `Cargo.toml` comments name the standards rules each new feature serves.

Findings inline below.

brendanbyrne:
Addressed all three in a494a79:

**🟡 Stale `is_active` reference** — you're right, that was fiction. The `sessions` table has `status: SessionStatus`, not `is_active`. Took your suggested rewrite near-verbatim: describes the actual safety mechanism (txn rollback on dropped future) rather than naming columns, so it stays correct under future schema tweaks. The doc-comment on `close_stale_sessions` itself was already correct — only the `main.rs` callsite comment had rotted.

**🔵 Test naming** — fair. Renamed all five:

- `test_wait_returns_immediately_for_closed_empty_tracker`
- `test_wait_returns_after_tracked_task_observes_cancel`
- `test_wait_times_out_for_stubborn_task`
- `test_supervised_runs_clean_future_to_completion`
- `test_supervised_catches_panic_and_returns`

`cargo test test_` now discovers them like everything else.

**🔵 Signal-install order** — agreed, free structural tightening. Moved `shutdown::signal(cancel.clone())?` up to just after `let cancel`/`let tracker`, before either `tracker.spawn` site. Now if the SIGTERM handler can't be installed (the documented `io::Error` case), `main` bails before either task is detached — orphan-free error path matching the placement of `Config::from_env()?` and `db::connect()?`. Bonus from your review I hadn't realized: on Unix, the OS handler is registered as soon as `signal::unix::signal()` returns, so a SIGTERM that races startup queues at the OS level and gets picked up the moment the shutdown future starts polling. Strictly better than the old placement.

Smoke test re-run produces the same seven-line shutdown sequence as before. `cargo fmt --check`, `cargo clippy --all-targets`, `cargo test` — all clean.

brendanbyrne (commented):
## Re-review (commit a494a79c)

**Verdict:** lgtm. All three findings from the prior review are addressed.

### Findings → fixes

- 🟡 **Stale `is_active` reference** (main.rs:286-290) — replaced verbatim with the suggested rewrite. ✓
- 🔵 **Test naming** (shutdown.rs) — all five tests now carry the `test_` prefix (`test_wait_returns_immediately_for_closed_empty_tracker`, etc.). ✓
- 🔵 **Signal-install order** (main.rs) — moved up to lines 106-115, right after `tracker = TaskTracker::new()` and before either `tracker.spawn(...)` call. The added paragraph ("On Unix, the SIGTERM handler is registered with the OS as soon as `signal::unix::signal` returns, so a SIGTERM that races startup is queued and observed once the shutdown future starts polling") is a thoughtful addition — it heads off the natural "but what if SIGTERM lands before the shutdown future is polled?" worry that motivates the late-install instinct. ✓

### What was verified

- `cargo test --lib shutdown` — 5 passed (the renamed `test_*` tests).

### Compliance-plan catch-up (commit 48ab5a5e)

The branch also carries a separate bookkeeping commit flipping 9 already-merged compliance-plan PRs (G1/G2/F3/E1/F5/D1/D2/D3/D4) from `[ ]` to `[x]`, with a new Document history entry explaining the catch-up. Orthogonal to PR-F1/PR-F2's scope but a clean tidy on the same branch — accepting that this lands here.

Ready to merge from my side.

PR #154, opened 2026-05-14
153: Enable rustc missing_docs lint + document ~225 surfaced sites

### LINKED ISSUE #153: Enable clippy/rustdoc missing_docs to catch undocumented pub items
## Context

`rust.md` § 6 requires every `pub` and cross-module `pub(crate)` item to carry a `///` doc comment. PR-G3 (#114) was the bulk pass that brought existing items into compliance, but there's no lint preventing regressions on new code.

PR #152 review surfaced this concretely: a doc-block merge regression in `backend/src/shutdown.rs` left `pub async fn wait` with no doc comment, attached to `pub async fn supervised` instead. Clippy missed it; the regression was caught only by human review. From the line comment:

> Not caught by clippy because `missing_docs` isn't enabled in the lint config.

## Scope

- Audit which lints actually enforce \"every `pub` item has a doc comment\":
  - `rust-by-example` convention is `#![deny(missing_docs)]` (rustc lint, not clippy). Applies to `pub` items; doesn't extend to `pub(crate)`.
  - `clippy::missing_docs_in_private_items` is the broader (and noisier) version covering private items too.
- Decide on one of these (probably `missing_docs` at workspace level, not the private-items version) and add it to `[workspace.lints]` in `backend/Cargo.toml`.
- Fix any sites that surface — likely a handful of trivially-missed ones plus any `pub fn` in places where the doc lives on the module instead.
- Confirm the lint catches the `shutdown.rs` regression class by intentionally removing a `///` block locally and observing the error.

## Standards refs

- `rust.md` § 6 (the doc-comment rules this lint would enforce).
- `rust.md` § 8 (lint configuration block in workspace `Cargo.toml`).

## Effort

S. Lint flip + small mechanical cleanup.

## Dependencies

None.

## Risk

Low. Lint surface change; no behavior change. If the audit reveals many bare-`pub` items (likely in test-helpers or internal-but-`pub` modules), the work scales — but PR-G3 already did the bulk pass, so the residue should be small.

## Verification

- `cargo clippy --all-targets -- -D warnings` is clean after the cleanup.
- Locally remove a `///` block from a `pub` item and confirm clippy flags it.

## Tracking

This Issue is the tracking artifact. Filed during PR #152 review (re-review comment on the `shutdown.rs` doc-block merge).

brendanbyrne:
## Summary

Filed during PR #152 review after a doc-block merge regression left `pub async fn wait` undocumented (attached to a sibling fn's `///` block). The lint that would have caught that — rustc's `missing_docs` — wasn't enabled. This PR turns it on and documents what it surfaces. Closes #153.

## Changes

- **`backend/Cargo.toml`** — `[workspace.lints.rust]` adds `missing_docs = "warn"` with a comment on why we chose this over `clippy::missing_docs_in_private_items` (the clippy variant covers bare-private items, which `rust.md` § 6 doesn't require). `pub(crate)` regressions remain a known gap that review covers; the rustc lint has no `pub`-and-`pub(crate)`-only knob. With `warnings = "deny"` already at the workspace level, `warn` is effectively an error.

- **Two opt-out boundaries:**
  - `#[allow(missing_docs)]` on `pub mod entities;` in `lib.rs` — the 23 entity files under it are mechanical SeaORM column declarations per ADR 0023; the column name IS the documentation. Same justification as the coverage exclusion in `design.md`.
  - `#![allow(missing_docs)]` at the top of `migration/src/m20260101_000001_initial_schema.rs` — the `Iden` enum variants are column-name scaffolding the migration DSL requires.

- **~225 `///` doc comments added across 23 files:**
  - DTO struct + field docs in `routes/auth.rs`, `routes/drink_types.rs`, `routes/game_data.rs`, `routes/runs.rs`, `routes/sessions.rs`, `routes/users.rs`, `services/runs/{read,submission}.rs`, `services/sessions/{detail,types}.rs`, `services/users.rs`.
  - Enum variant docs in `domain/enums.rs` (`SessionStatus::Active`, `SessionRuleset::Random`, etc.).
  - `pub mod` re-export docs in `lib.rs`, `routes/mod.rs`, `middleware/mod.rs`, `domain/mod.rs`, `services/mod.rs`.
  - Crate-level `//!` on `main.rs` and the two integration-test crates that needed one.
  - Struct field docs on a handful of `pub` items in `config.rs`, `error.rs`, `middleware/auth.rs`, `services/session_context.rs`, `services/sessions/lifecycle.rs`, `domain/race_setup.rs`.

- **`docs/project-workflow.md`** — Cowork-authored addition pulled into this PR per request: documents the `<issue_number>: <Title>` PR title convention (mirrors the existing commit-message convention). This PR's title uses it.

## Test plan

**Automated (in this PR):**

- [x] `cargo build --all-targets` — clean (was 392 `missing_docs` errors before this PR).
- [x] `cargo clippy --all-targets --all-features -- -D warnings` — clean.
- [x] `cargo test` — all 335+ tests pass.

**Regression catch verification:**

- [x] Locally stripped `///` from `pub async fn wait` in `shutdown.rs` → `cargo build` produces `error: missing documentation for a function`. Restored the doc; same regression class that slipped past review in PR #152 is now compile-time caught.

**Reviewer manual run:**

- [ ] `cd backend && cargo build --all-targets` — confirm clean.
- [ ] In any `pub fn`/`pub struct`/`pub enum`/field, locally delete the `///` block above it and confirm `cargo build` fails with `error: missing documentation for ...`. (Entity files are intentionally opted out — try a route handler or service DTO instead.)

## Compliance-plan reference

This Issue is its own tracking artifact (filed today during PR #152 review). The work directly enforces `rust.md` § 6 ("Every `pub` and cross-module `pub(crate)` item gets a `///` doc comment").

### DIFF
--- a/backend/src/routes/auth.rs
+++ b/backend/src/routes/auth.rs
@@ -48,7 +48,7 @@ pub struct ChangePasswordRequest {
     pub new_password: String,
 }
 
-/// Body shape for `POST /auth/login` and `POST /auth/register`.
+/// Success-response body for `POST /auth/login` and `POST /auth/register`.
 #[derive(Serialize)]
 pub struct Response {
     /// Short-lived JWT bearer token to send on subsequent requests.
@@ -57,7 +57,7 @@ pub struct Response {
     pub user: UserInfo,
 }
 
-/// Body shape for `POST /auth/refresh`.
+/// Response body for `POST /auth/refresh`.
 #[derive(Serialize)]
 pub struct RefreshResponse {
     /// Newly-minted short-lived JWT bearer token.
--- a/backend/src/routes/auth.rs
+++ b/backend/src/routes/auth.rs
@@ -21,38 +21,55 @@ use crate::{
 
 // ── Request / Response types ────────────────────────────────────────
 
+/// Body shape for `POST /auth/register`.
 #[derive(Deserialize)]
 pub struct RegisterRequest {
+    /// Desired handle. Subject to `Username` validation at the service boundary.
     pub username: String,
+    /// Plaintext password. Hashed server-side; never stored.
     pub password: String,
 }
 
+/// Body shape for `POST /auth/login`.
 #[derive(Deserialize)]
 pub struct LoginRequest {
+    /// Account handle.
     pub username: String,
+    /// Plaintext password. Verified server-side via Argon2.
     pub password: String,
 }
 
+/// Body shape for `PUT /auth/password`.
 #[derive(Deserialize)]
 pub struct ChangePasswordRequest {
+    /// Plaintext current password. Verified server-side before the rotation.
     pub current_password: String,
+    /// New plaintext password. Hashed server-side; replaces the old hash.
     pub new_password: String,
 }
 
+/// Body shape for `POST /auth/login` and `POST /auth/register`.
 #[derive(Serialize)]
 pub struct Response {
+    /// Short-lived JWT bearer token to send on subsequent requests.
     pub access_token: String,
+    /// Public-facing identity bundle.
     pub user: UserInfo,
 }
 
+/// Body shape for `POST /auth/refresh`.
 #[derive(Serialize)]
 pub struct RefreshResponse {
+    /// Newly-minted short-lived JWT bearer token.
     pub access_token: String,
 }
 
+/// Embedded identity bundle on auth responses.
 #[derive(Serialize)]
 pub struct UserInfo {
+    /// User's stable UUID.
     pub id: UserId,
+    /// Display handle.
     pub username: Username,
 }
 
--- a/backend/src/routes/sessions.rs
+++ b/backend/src/routes/sessions.rs
@@ -14,8 +14,12 @@ use crate::{
     services::sessions,
 };
 
+/// Body shape for `POST /sessions`.
 #[derive(Deserialize)]
 pub struct CreateSessionRequest {
+    /// Track-selection ruleset name (e.g. `"random"`, `"least_played"`).
+    /// Validated against [`crate::domain::enums::SessionRuleset`] at the
+    /// service boundary; #146 will move this to the newtype directly.
     pub ruleset: String,
 }
 
@@ -37,8 +41,10 @@ pub async fn create_session(
     Ok((StatusCode::CREATED, Json(detail)))
 }
 
+/// Body shape for `GET /sessions/mine`.
 #[derive(Serialize)]
 pub struct MySessionResponse {
+    /// The user's current active session, or `None` if they're not in one.
     pub session_id: Option<SessionId>,
 }
 
--- a/backend/src/routes/sessions.rs
+++ b/backend/src/routes/sessions.rs
@@ -41,7 +41,7 @@ pub async fn create_session(
     Ok((StatusCode::CREATED, Json(detail)))
 }
 
-/// Body shape for `GET /sessions/mine`.
+/// Response body for `GET /sessions/mine`.
 #[derive(Serialize)]
 pub struct MySessionResponse {
     /// The user's current active session, or `None` if they're not in one.
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -5,6 +5,11 @@ members = ["migration"]
 warnings = "deny"
 unsafe_code = "forbid"
 rust_2018_idioms = { level = "warn", priority = -1 }
+# Enforces `rust.md` § 6 on `pub` items. Rustc's lint only covers `pub`
+# visibility (not `pub(crate)`), so cross-module-`pub(crate)` regressions
+# still rely on review; the broader `clippy::missing_docs_in_private_items`
+# would also force docs on bare-private items, which § 6 doesn't require.
+missing_docs = "warn"
 
 [workspace.lints.clippy]
 pedantic = { level = "warn", priority = -1 }
--- a/backend/migration/src/lib.rs
+++ b/backend/migration/src/lib.rs
@@ -7,6 +7,7 @@
 
 pub use sea_orm_migration::prelude::*;
 
+/// Consolidated initial-schema migration — table, FK, and index DDL.
 pub mod m20260101_000001_initial_schema;
 
 /// Runs the consolidated initial-schema migration.
--- a/backend/migration/src/m20260101_000001_initial_schema.rs
+++ b/backend/migration/src/m20260101_000001_initial_schema.rs
@@ -1,6 +1,10 @@
 // PR-A1 scaffolding: rust-2018 idioms cleanup deferred to Phase H per
 // rust.md § 8 ("config and clearing land in separate PRs").
 #![allow(elided_lifetimes_in_paths)]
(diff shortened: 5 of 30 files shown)


backend/src/routes/auth.rs
```diff
@@ -21,38 +21,55 @@ use crate::{
...
 #[derive(Deserialize)]
 pub struct ChangePasswordRequest {
+    /// Plaintext current password. Verified server-side before the rotation.
     pub current_password: String,
+    /// New plaintext password. Hashed server-side; replaces the old hash.
     pub new_password: String,
 }
 
+/// Body shape for `POST /auth/login` and `POST /auth/register`.
 #[derive(Serialize)]
```
brendanbyrne, line 52: 🟡 **Important** — This struct is `#[derive(Serialize)]` (server-emitted), but the `Body shape for ...` phrasing is identical to the one used on `RegisterRequest`/`LoginRequest`/`ChangePasswordRequest` above. A reader can't tell from the doc alone which direction the type travels — they have to check the derive.

Per `rust.md` § 6 (last rule, the spirit of "stale docs actively mislead"), changing the leading word disambiguates:

```rust
/// Success-response body for `POST /auth/login` and `POST /auth/register`.
#[derive(Serialize)]
pub struct Response {
```

(Aside, not asking for it in this PR: `Response` is unusually generic as a type name — `AuthResponse` or `LoginResponse` would be clearer — but that's pre-existing.)

backend/src/routes/auth.rs
```diff
@@ -21,38 +21,55 @@ use crate::{
...
 pub struct Response {
+    /// Short-lived JWT bearer token to send on subsequent requests.
     pub access_token: String,
+    /// Public-facing identity bundle.
     pub user: UserInfo,
 }
 
+/// Body shape for `POST /auth/refresh`.
```
brendanbyrne: 🟡 **Important** — Same pattern as the `Response` comment above: this is `#[derive(Serialize)]`, so the type travels server→client, but `Body shape for ...` is the same phrasing used on the request DTOs in this file.

Suggested fix:

```rust
/// Response body for `POST /auth/refresh`.
#[derive(Serialize)]
pub struct RefreshResponse {
```

backend/src/routes/sessions.rs
```diff
@@ -37,8 +41,10 @@ pub async fn create_session(
     Ok((StatusCode::CREATED, Json(detail)))
 }
 
+/// Body shape for `GET /sessions/mine`.
```
brendanbyrne: 🟡 **Important** — `GET /sessions/mine` has no request body, so the only sensible reading is "response body". But four lines up, `CreateSessionRequest` uses `/// Body shape for POST /sessions.` for its request body — same phrasing for the opposite direction in the same file.

Suggested fix:

```rust
/// Response body for `GET /sessions/mine`.
#[derive(Serialize)]
pub struct MySessionResponse {
```

backend/src/routes/sessions.rs
```diff
@@ -14,8 +14,12 @@ use crate::{
     services::sessions,
 };
 
+/// Body shape for `POST /sessions`.
 #[derive(Deserialize)]
 pub struct CreateSessionRequest {
+    /// Track-selection ruleset name (e.g. `"random"`, `"least_played"`).
+    /// Validated against [`crate::domain::enums::SessionRuleset`] at the
```
brendanbyrne, line 21: 🔵 **Suggestion** — `` [`crate::domain::enums::SessionRuleset`] `` works fine as an intra-doc link, but the short form `` [`SessionRuleset`] `` is more refactor-friendly if the type is brought into scope at the top of the file (it's referenced again in the `# Errors` block on `create_session` at line 32). Optional — both forms resolve correctly today.

brendanbyrne (commented):
## Code review

**Verdict:** Approve, with 3 minor wording fixes worth considering (inline below).

**What this PR does:** Enables rustc's `missing_docs` lint at the workspace level, adds two well-justified opt-out boundaries (entities module + initial-schema migration file), and adds ~225 `///` doc comments to the surfaced `pub` items. Also adds a PR-title convention to `project-workflow.md`.

**Areas covered:** `rust.md` (§ 6 docs, § 14 serde DTO separation, § 15 Cargo hygiene), `docs/CLAUDE.md` (doc-area conventions).

**Automated checks**

- `cargo fmt --check` — clean
- `cargo test` — passes (exit 0)
- Manually verified the lint is active by confirming `pub async fn wait` in `shutdown.rs` already has `///` docs (the exact regression class from PR #152 review).

**Findings**

All inline below. Three 🟡 on the same wording pattern (response structs documented with the same `Body shape for ...` phrasing as request structs nearby, so the reader can't tell direction from the doc alone) plus one 🔵 on a verbose intra-doc link.

**What's good**

- The `Cargo.toml` comment is genuinely useful — explains why `missing_docs` (rustc) over `clippy::missing_docs_in_private_items` and acknowledges the `pub(crate)` gap reviewers still cover.
- Both opt-outs (`#[allow(missing_docs)]` on `pub mod entities;` and `#![allow(missing_docs)]` in the migration file) are minimal and reference ADR 0023 / the SeaORM `Iden` convention. Opt-out at the module declaration rather than `#![allow]` in 23 entity files is the right scope.
- Regression-catch verification in the test plan (strip the `///` from `wait` → compile fails) demonstrates the lint catches the exact class of bug from PR #152 review.
- The `project-workflow.md` Title section is appropriately concise; its Document history entry follows convention; this PR's own title (`153: Enable rustc...`) follows the new format.
- Crate-level `//!` docs added to `main.rs` and the two integration-test crates — § 6 satisfied.

**Design-doc accuracy**

No changes required. The diff doesn't touch data model, API surface, or architecture.

brendanbyrne:
Pushed 62aa37b. Addressed the three 🟡 wording fixes, declined the 🔵:

**🟡 (×3) — request/response docs were indistinguishable.** You're right — same phrasing on both directions in the same files is exactly the \"stale docs actively mislead\" failure mode. Applied your suggested wording:

- `Response` → \"Success-response body for `POST /auth/login` and `POST /auth/register`.\"
- `RefreshResponse` → \"Response body for `POST /auth/refresh`.\"
- `MySessionResponse` → \"Response body for `GET /sessions/mine`.\"

(Pre-checked the rest of the routes/services tree for `Body shape for` to be sure I hadn't done the same thing on other Serialize types. The remaining three uses are all on request DTOs — correct.)

**🔵 — declined the intra-doc-link shortening.** Tried it first. Adding `SessionRuleset` to the file's `use` block trips `unused_imports`: the type is referenced only in doc comments, not as a Rust item in this file. Your note flagged this as conditional (\"if the type is brought into scope\"); since the condition isn't satisfied cleanly, kept the long-form path. If we wanted the short form in a future change, the cheapest route would be when the file actually starts referring to `SessionRuleset` in code — likely once #146 moves `CreateSessionRequest.ruleset` from `String` to the typed newtype.

**Aside on `Response` as a type name** — agreed, it's unusually generic. Not changing it here per your note, but worth filing if you want it tracked.

PR #155, opened 2026-05-15
123: Per-call DB timeouts (PR-F4)

### LINKED ISSUE #123: Per-call timeouts on DB and external calls (PR-F4)
## Context

Defence in depth for the SQLite-lock-foot-gun and (eventually) outbound HTTP. Wraps every `await` against the database in `tokio::time::timeout` so a stuck query can't pin a worker indefinitely.

## Scope

- Wrap every `await` against SeaORM and any future outbound HTTP in `tokio::time::timeout`.
- Introduce a `db_timeout!` macro (or wrapper helper) to keep call sites tidy. Default 2s for queries, 5s for transactions.
- Map elapsed errors to `AppError::Internal` — **or** a new `AppError::Timeout` variant if we want a distinct status. Discuss before landing; my recommendation is the dedicated variant + 504 status so operators can grep for timeouts independently of "real" 500s.

## Standards refs

- [`docs/coding-standards/tokio.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/coding-standards/tokio.md) § 12

## Dependencies

PR-C1 (#105) — AppError shape stable. **Done.**

If adding `AppError::Timeout`: small ripple on the `IntoResponse` impl from PR-C2 (#107). Mechanical.

## Risk

Medium. Timeouts that fire under load are operational pain; need monitoring before tightening. Default 2s/5s is generous for SQLite-on-Unraid in a single-user Star phase; revisit before opening to friends.

## Effort

L. Touches every service.

## Verification

- Force a SQLite lock (long write transaction in another connection); confirm the blocked query times out cleanly with the configured budget.
- A unit test stubs `Database::execute` to sleep past the budget and asserts the error variant.

## Compliance-plan reference

[`docs/compliance-plan.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/compliance-plan.md) § PR-F4.

brendanbyrne:
Closes #123 (PR-F4 of the [compliance plan](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/compliance-plan.md#pr-f4-per-call-timeouts-on-db-and-external-calls)).

## Summary

Wraps every SeaORM `await` in the backend with a per-call `tokio::time::timeout` budget so a stuck query can no longer pin a worker indefinitely. Elapsed budgets surface as a new `Error::Timeout { budget }` variant mapped to `504 Gateway Timeout`, distinct from `Error::Internal` (500) so operators can chart timeouts independently of generic 500-class failures.

- New helpers in [`backend/src/timeout.rs`](https://github.com/brendanbyrne/beerio-kart/blob/123/per-call-db-timeouts/backend/src/timeout.rs):
  - `db_query(fut).await?` — 2 s, single statements (`Entity::find().one()`, `insert()`, `update()`, `delete()`, `update_many().exec()`, etc.).
  - `db_txn(fut).await?` — 5 s, the transaction-frame statements (`db.begin()`, `txn.commit()`, `txn.rollback()`). The body of the transaction stays per-statement-budgeted via `db_query`.
- New `Error::Timeout` variant on the unified `Error` enum. `#[non_exhaustive]` already, so non-breaking. `IntoResponse` arm logs `budget_ms` via `tracing::warn!` and returns generic `"Request timed out"` to the client (budget never leaks).
- Every leaf `await` against SeaORM in `services/`, `routes/`, and `seed.rs` is wrapped. The campaign is split across 12 small commits, one per file/module, so each is reviewable in isolation.
- `tokio.md` § 12 refreshed: the prior example used the stale `AppError::Internal("db timeout".into())` shape (renamed + reshape since PR-C1/C2); replaced with `db_query` / `db_txn` example, plus subsections covering the function-vs-macro rationale and the 504 mapping.
- Verification test in [`tests/db_timeout.rs`](https://github.com/brendanbyrne/beerio-kart/blob/123/per-call-db-timeouts/backend/tests/db_timeout.rs) forces a real SQLite write lock from one pool and confirms a second writer wrapped in `db_query` returns `Error::Timeout` at ~2 s rather than waiting through the pool's 5 s `busy_timeout`.

## Design call (talked through in chat first)

Brendan's call: Option A — add `Error::Timeout` + 504, not map to `Internal`. Reasons documented on the variant itself ([error.rs](https://github.com/brendanbyrne/beerio-kart/blob/123/per-call-db-timeouts/backend/src/error.rs)). The status code is 504 Gateway Timeout because SQLite is the "upstream" in the 504 sense even though it's in-process; the proxy-flavoured status is semantically apt and gives the operational signal we want.

Functions over a `db_timeout!` macro: tracing spans from PR-F5 already supply call-site context, so there's no need for `stringify!` to capture the expression. Functions compose with `?` naturally and don't fight rustfmt on multi-line builder chains.

## Test plan

Most of this is covered automatically — running `cargo test` from `backend/` exercises all 343 existing tests plus the 8 new ones (3 in `error.rs`, 4 in `timeout.rs`, 1 verification test). To convince yourself by hand:

- [ ] **Build + lint clean.** From `backend/`:
  ```bash
  cargo build
  cargo clippy --all-targets
  cargo test
  ```
  Expect all 343 unit + integration tests green and zero clippy warnings.
- [ ] **`Error::Timeout` variant unit tests.**
  ```bash
  cargo test --manifest-path backend/Cargo.toml --lib error::tests::test_timeout
  ```
  All three pass:
  - `test_timeout_response_status_is_504_gateway_timeout`
  - `test_timeout_response_body_is_generic_and_omits_budget`
  - `test_timeout_emits_warn_with_budget_ms`
- [ ] **Helper unit tests.**
  ```bash
  cargo test --manifest-path backend/Cargo.toml --lib timeout::
  ```
  All five pass — happy path, Elapsed → Timeout via the paused-runtime trick (`start_paused = true`), and `From<DbErr>` passthrough so non-timeout inner errors still convert correctly.
- [ ] **Verification test (the headline behaviour).**
  ```bash
  cargo test --manifest-path backend/Cargo.toml --test db_timeout
  ```
  Should report `1 passed; ... finished in ~2.1s`. The ~2.1 s wall clock is the headline: SQLite's `busy_timeout` is 5 s; if you saw ~5 s here, the helper would be doing nothing.
- [ ] **Smoke-test a live request.** From the repo root with the backend running (`cargo run --manifest-path backend/Cargo.toml`), open a SQLite shell in another terminal:
  ```bash
  sqlite3 data/db/beerio-kart.db
  ```
  Inside the shell, hold a write lock by starting a transaction with a write:
  ```sql
  BEGIN IMMEDIATE;
  UPDATE sessions SET last_activity_at = datetime('now') WHERE 1 = 0;
  -- now leave the shell sitting at the prompt
  ```
  Hit any write endpoint from the running backend (`curl -X POST http://localhost:3000/api/v1/sessions ...` or similar with auth). You should get a `504 Gateway Timeout` response with body `{"error":"Request timed out"}` within ~2.5 s, and the server log should show a `WARN ... budget_ms=2000 Operation timed out` line. Close the SQLite shell with `ROLLBACK;` once verified.
- [ ] **No false positives under normal load.** Restart the backend with no held lock, log in, run a session creation + a couple of next-track calls. Confirm everything completes normally (no timeouts firing on uncontended queries).


backend/src/error.rs
```diff
@@ -121,6 +136,13 @@ impl IntoResponse for Error {
                     "Internal server error".to_string(),
                 )
             }
+            Self::Timeout { budget } => {
+                warn!(
+                    budget_ms = u64::try_from(budget.as_millis()).unwrap_or(u64::MAX),
```
brendanbyrne: 🔵 **Suggestion** — `unwrap_or(u64::MAX)` for `budget_ms`.

`Duration::as_millis()` returns `u128`; the downcast to `u64` is genuinely fallible, but only at budgets > ~584 million years. Using `u64::MAX` as the fallback masks an overflow as a "huge number" in the logs, which is misleading. Practically safe — our budgets are 2 s and 5 s — but the fallback obscures intent.

If you want to keep this defensive *and* unambiguous, this expression cannot overflow `u64`:

```rust
budget_ms = budget.as_secs() * 1000 + u64::from(budget.subsec_millis()),
```

`Duration::as_secs()` already returns `u64`, so the result is a real `u64` within `Duration`'s native range — no `try_from`, no sentinel. Not blocking; minor.

docs/coding-standards/tokio.md
```diff
@@ -363,15 +363,25 @@ The 5-minute "close stale sessions" task is the canonical example for this proje
...
   - **Why:** Without a timeout, a stuck call holds a connection, a permit, and the calling task forever. SeaORM's transaction methods will wait indefinitely if SQLite is locked. Per-call also lets auth verification have a 1s ceiling where a stats aggregation gets 5s — Tower's `TimeoutLayer` is uniform.
+  - **Helper:** `backend/src/timeout.rs` exposes two function wrappers backed by `tokio::time::timeout`:
+    - `db_query(fut).await?` — 2 s budget. Use around every single-statement read or write (`Entity::find().one()`, `Entity::find().all()`, `ActiveModel::insert()`, `update()`, `delete()`, `update_many().exec()`). Per-statement budget applies inside transactions too.
+    - `db_txn(fut).await?` — 5 s budget. Use around `db.begin()`, `txn.commit()`, and `txn.rollback()` — i.e. the transaction-frame statements themselves. The body of the transaction stays per-statement-budgeted via `db_query` on each call inside.
```
brendanbyrne: 🔵 **Suggestion** — the bullet lists `db.begin()`, `txn.commit()`, **and** `txn.rollback()` as the `db_txn`-wrapped frame statements, but the worked example below only shows the commit path. A reader who hits an explicit rollback path won't immediately know whether to wrap it.

In practice the codebase relies on `?`-bail-out drops to roll back (correct per `seaorm.md` § 3), so explicit `rollback` is rare. Two ways to close the gap:

1. Add a one-line aside that dropping the txn without `commit` rolls back automatically — so explicit `db_txn(txn.rollback())` is only for cases where you *want* to await the rollback materialising.
2. Extend the example with a `db_txn(txn.rollback())` branch.

(1) is probably enough given how rare the explicit-rollback case is.

brendanbyrne (commented):
## Verdict — Approve with one fix

## What this PR does

Adds `timeout::db_query` (2 s) and `timeout::db_txn` (5 s) helpers wrapping every SeaORM `await` in the backend. An elapsed budget surfaces as a new `Error::Timeout { budget }` variant that maps to `504 Gateway Timeout`, distinct from the 500-class so operators can chart timeouts separately. Updates `tokio.md` § 12 to document the helpers; signs off PR-F4 in `compliance-plan.md`.

**Areas covered:** `rust.md`, `seaorm.md`, `tokio.md`, `api-contract.md`, `design.md`.

**Automated checks (run locally on the PR HEAD):**
- `cargo +nightly fmt --check` — clean.
- `cargo clippy --all-targets --all-features` — zero warnings.
- `timeout::*` unit tests (5) + `error::*` tests (20) + `tests/db_timeout` (1) — all pass; the verification test ran in ~2.07 s, confirming the helper fires at the 2 s budget, not the pool's 5 s `busy_timeout`.

**Coverage audit:** spot-checked every `.await` against SeaORM in `routes/`, `services/`, `seed.rs`, and `services/helpers.rs`. Every leaf call is wrapped, transaction frames use `db_txn`, helpers that take `&impl ConnectionTrait` wrap their inner statements internally. The "frame uses `db_txn`, body statements use `db_query`" pattern is consistently applied.

---

## Findings

### 🟡 Important — `docs/design.md` not updated to reflect the new variant

`design.md` is the project's single source of truth for architecture, and the "Observability → Error response pattern" section is the canonical table of `error::Error` variants. `design.md` was not touched in this PR; two specific gaps:

1. **Variants table** (`design.md` lines 78–86) lists 8 rows but no `Timeout`. Add:
   ```
   | `Timeout { budget }` | 504 | Generic `"Request timed out"` |
   ```
2. **Stale parenthetical** at line 93 — the existing text reads:
   > `error::Error` is `#[non_exhaustive]`: the compiler enforces a wildcard arm in any external matcher so adding a future variant (e.g., `Timeout`, `RateLimited`) doesn't break callers.

   `Timeout` is no longer hypothetical. Drop it from the example list (leaving `RateLimited`), or restructure the sentence.

Also add a Key behaviors bullet describing the Timeout semantics (504 mapping, `warn!`-logged `budget_ms`, generic body, why-distinct-from-Internal) and a Document history entry for the change. Per `docs/CLAUDE.md` Document history rule, `design.md` is in scope; per the code-review skill, design-doc updates land in the same PR as the code they describe.

### 🔵 Suggestion — `api-contract.md` § 8 error code registry

The registry (`api-contract.md` lines 244–265) has no row for 504. The current `IntoResponse` body still ships only `{"error": "..."}` (the `code` field is deferred per § 3), so this isn't a wire-contract violation today — but the registry already documents codes ahead of implementation (e.g., `lap_times_mismatch`). Adding

```
| 504 | `gateway_timeout` | Per-call timeout budget elapsed; retry is safe. |
```

keeps the registry complete and avoids drift when the `code` field eventually lands.

### 🔵 Suggestion — inline comments

See inline comments on `backend/src/error.rs` and `docs/coding-standards/tokio.md`.

---

## What's good

- Variant design is operationally sharp: `Error::Timeout { budget }` lets each call site declare its own ceiling and surfaces it to operators via structured `warn!`. 504 vs 500 is exactly the kind of decision that pays back at on-call time.
- The generic-on-`From<E>` helper signature lets `?` keep working with both `DbErr` and `Error` at call sites — no boilerplate ripples through every handler.
- Verification test in `backend/tests/db_timeout.rs` is genuinely behavioural: real SQLite file lock, two pools (not two checkouts from one), wall-clock guard `elapsed < QUERY_BUDGET + 1.5s`. The comments document *why* each setup choice matters (file lock vs in-memory, two pools vs two connections). Future readers will thank you.
- `start_paused = true` for the unit tests is a clean pattern — deterministic and instant. Good `Cargo.toml` comment explaining the resolver-3 / no-feature-leak rationale.
- Pre-existing patterns honoured: `#[non_exhaustive]`, structured `warn!` (not `error!`), generic client message, source plumbing — all consistent with the rest of the error layer.

brendanbyrne:
All four items addressed in 95f95c3. Summary:

**🟡 design.md — done.** Added the `Timeout { budget }` row to the variants table, added a Key behaviors bullet covering the variant's semantics (504 mapping, `warn!`-logged `budget_ms`, generic body, why-distinct-from-Internal, SQLite-as-upstream caveat for the 504), dropped `Timeout` from the `#[non_exhaustive]` parenthetical's hypothetical-variant list, and added a Document history entry.

**🔵 api-contract.md § 8 — done.** Added `504 | gateway_timeout`. Document history entry added.

**🔵 tokio.md `txn.rollback()` aside — done with one tweak.** Went with option 1 (the aside). One deviation from the literal suggestion: I avoided cross-referencing `seaorm.md` § 3 in the aside because that section's rule actually prefers the closure form over manual `begin`/`commit`, which conflicts with the manual pattern this codebase uses. Describing the `Drop` behaviour directly seemed cleaner than re-litigating the closure-vs-manual choice in a PR about timeouts. Happy to swap to the explicit cross-reference if you'd rather the consistency.

**🔵 `budget_ms` form — done.** Switched to `budget.as_secs() * 1000 + u64::from(budget.subsec_millis())` and dropped a brief comment explaining why this is direct integer math (no `u128` → `u64` cast, no sentinel masking an overflow as a real-looking 18-quintillion-ms log entry). `test_timeout_emits_warn_with_budget_ms` still passes (the assertion is `logs_contain("budget_ms=2500")` against a 2500ms input — works identically under the new formula).

brendanbyrne (commented):
## Re-review — Approve

Commit `95f95c3` addresses all four findings from the prior review. Verified locally on the new HEAD:

- `cargo +nightly fmt --check` — clean.
- `cargo clippy --all-targets --all-features` — zero warnings.
- `error::tests::test_timeout*` — all 3 pass (the `logs_contain("budget_ms=2500")` assertion still matches under the new computation form, as expected).

### Findings — disposition

**🟡 Important — `design.md` Timeout variant** → **resolved.** Variants table gained the `Timeout { budget }` row (504, generic body). New Key behaviors bullet covers the variant's full semantics: source (`db_query`/`db_txn` only), `budget_ms` logged via `warn!` for operators, why-distinct-from-Internal, why-504-for-in-process-SQLite, cross-reference to `tokio.md` § 12. The `#[non_exhaustive]` parenthetical dropped `Timeout` from the hypothetical list (it's no longer hypothetical) and gained a one-line aside noting the variant landed without breaking existing matchers. Document history entry added. The expanded bullet is substantive — it captures the *operational* rationale (chart timeout rates separately from generic 500s) that was in the variant docstring but not in `design.md`.

**🔵 `api-contract.md` § 8 registry** → **resolved.** Row added: `504 | gateway_timeout | Per-call timeout budget elapsed; retry is safe.` The "retry is safe" phrasing is accurate — `db_query` either runs the statement to completion or drops the future before commit, so a retry sees consistent state. Document history entry added.

**🔵 `tokio.md` § 12 rollback aside** → **resolved.** The `db_txn` bullet now spells out the dominant `?`-bail-out + `Drop`-rollback pattern, and names the specific cases where explicit `db_txn(txn.rollback())` is needed (awaiting the rollback to materialise — e.g., to release the SQLite write lock before continuing in the same scope, or to surface a rollback error rather than letting `Drop` swallow it). Precise and minimal. The commit message notes the author considered and declined a `seaorm.md` § 3 cross-reference because that section's closure-form recommendation conflicts with the codebase's manual-begin/commit pattern — defensible call, and I didn't actually suggest the cross-reference in the inline.

**🔵 `budget_ms` form** → **resolved.** Switched from `u64::try_from(budget.as_millis()).unwrap_or(u64::MAX)` to `budget.as_secs() * 1000 + u64::from(budget.subsec_millis())`. `as_secs()` returns `u64` directly and `subsec_millis()` returns a `u32` < 1000, so this is straight integer math with no fallible cast and no `u64::MAX` sentinel that could appear in logs as a misleading real value. Multiplication can overflow `u64` only beyond ~584 million years of duration, which is fine for 2 s / 5 s budgets. Inline comment in `error.rs` explains the choice for the next reader.

---

Nothing else to flag. Approving from my side; the merge call is Brendan's per project workflow.

PR #158, opened 2026-05-15
157: Wire `code` field across all error responses (subsumes #146)

### LINKED ISSUE #146: Typed Path extractor 400s don't match the api-contract error envelope

### LINKED ISSUE #157: Wire `code` field across all error responses (subsumes #146)
## Context

[`docs/api-contract.md` § 3](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/api-contract.md#3-error-response-contract) specifies the error envelope as `{ \"error\": \"...\", \"code\": \"...\" }`. The current `IntoResponse` impl in [`backend/src/error.rs`](https://github.com/brendanbyrne/beerio-kart/blob/main/backend/src/error.rs) emits only `{ \"error\": \"...\" }` — no `code` field. [§ 8](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/api-contract.md#8-error-code-registry) maintains a forward-looking registry (codes get documented as features ship; the field wiring is deferred). PR-F4 ([#155](https://github.com/brendanbyrne/beerio-kart/pull/155)) review surfaced this gap explicitly when adding the `gateway_timeout` registry entry without a wire-format change.

This Issue tracks delivering the deferred work in one pass: introduce a typed `ErrorCode`, route every `Error` variant through it, and emit the field in the response body.

**Subsumes [#146](https://github.com/brendanbyrne/beerio-kart/issues/146).** The Path/Json extractor fix it asks for needs the same `IntoResponse` / response-envelope surface this Issue touches. Doing them as separate PRs would mean re-touching the same paths twice. The custom extractors land here as one slice.

### 1. New `ErrorCode` enum

Mirrors the [§ 8 registry](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/api-contract.md#8-error-code-registry). Serialized as `snake_case`.

```rust
#[derive(Copy, Clone, Debug, Serialize)]
#[serde(rename_all = \"snake_case\")]
pub enum ErrorCode {
    // 400
    BadRequest, LapTimesMismatch, TrackIdMismatch,
    InvalidPathParam, InvalidRequestBody,
    // 401
    InvalidCredentials, TokenExpired, TokenInvalid,
    // 403
    Forbidden, AdminRequired,
    // 404
    NotFound,
    // 409
    Conflict, UsernameTaken, SessionClosed,
    PendingRacesFirst, OutOfOrderSubmission, RaceNumberConflict,
    // 500
    Internal,
    // 504
    GatewayTimeout,
}
```

422 `unprocessable` stays registered-but-unimplemented for now — the JsonRejection-data-validation path maps to 400 `invalid_request_body` rather than introducing a new `Error::Unprocessable` variant. Splitting 400 vs 422 for JSON failures is a follow-up if/when the distinction earns its keep.

### 2. `Error` variants gain a `code` field where they need one

Variants that already carry exactly one stable code (`NotFound`, `Internal`, `Token`, `Hash`, `Timeout`) stay as-is — their code is determined by the variant. Variants that span multiple codes get a `code` field:

```rust
BadRequest { client, code: ErrorCode, detail },
Unauthorized { msg, code: ErrorCode },
Forbidden { msg, code: ErrorCode },
Conflict { client, code: ErrorCode, detail },
```

A `code()` method on `Error` returns the right `ErrorCode` for every variant.

### 3. `ErrorBody` gains the field, `IntoResponse` emits it

```rust
#[derive(Serialize)]
struct ErrorBody { error: String, code: ErrorCode }
```

### 4. Helper API — hybrid (decided 2026-05-15)

Per-code helpers for the named domain codes:

```rust
Error::lap_times_mismatch(\"...\")
Error::track_id_mismatch(\"...\")
Error::invalid_credentials()              // sentinel msg
Error::token_expired()
Error::token_invalid()
Error::admin_required()
Error::username_taken(\"...\")
Error::session_closed(\"...\")
Error::pending_races_first(\"...\")
Error::out_of_order_submission(\"...\")
Error::race_number_conflict(\"...\")
```

Generic helpers keep their current signatures and default to the generic codes:

```rust
Error::bad_request(\"...\")    // → ErrorCode::BadRequest
Error::conflict(\"...\")       // → ErrorCode::Conflict
Error::forbidden(\"...\")      // → ErrorCode::Forbidden
```

`Error::NotFound(msg)` stays as a tuple variant (single code), as does `Error::Internal(anyhow)` and `Error::Timeout { budget }`.

### 5. Custom Path/Json extractors (the #146 fix)

- Project-local `extract::Path<T>` wrapping `axum::extract::Path<T>`, mapping `PathRejection` → `Error::invalid_path_param(...)`.
- Project-local `extract::Json<T>` wrapping `axum::Json<T>`, mapping `JsonRejection` → `Error::invalid_request_body(...)`. All `axum::Json` import sites in `routes/` switch to the wrapper.
- Custom extractor for `AdminUser` already exists in middleware — gets its code from `Error::admin_required()`.

### 6. `From<DbErr>` impl assigns codes per branch

- `UniqueConstraintViolation` → currently emits generic `Conflict`. Re-classify: still `Error::Conflict { code: ErrorCode::Conflict, ... }` unless service-layer pre-checks (which would have caught the specific case — `UsernameTaken`, etc.) failed. Driver-string in `detail` survives.
- `ForeignKeyConstraintViolation` → `Error::BadRequest { code: ErrorCode::BadRequest, ... }`.
- `RecordNotFound` → `Error::NotFound(msg)`. Single code.
- Else → `Error::Internal(anyhow)`. Single code.

### 7. Tests

- Update `error::tests` to assert the response body contains both `error` and `code` fields.
- Update integration tests that read error response bodies (`auth_integration.rs`, `api_integration.rs`, `session_integration.rs`) to assert on `code` where they currently only assert on status.
- New tests for the custom extractors: malformed UUID in path returns 400 with `code: invalid_path_param`; malformed JSON body returns 400 with `code: invalid_request_body`; deserialize-failure (e.g., wrong type) routes through the same path.
- Verify `gateway_timeout` ([`tests/db_timeout.rs`](https://github.com/brendanbyrne/beerio-kart/blob/main/backend/tests/db_timeout.rs)) — the PR-F4 verification test should now also see `code: gateway_timeout` in the response body. Update assertion accordingly.

### 8. Docs

- [`docs/api-contract.md` § 3](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/api-contract.md#3-error-response-contract) — drop the deferred-code caveat, show the actually-emitted shape.
- [`docs/api-contract.md` § 8](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/api-contract.md#8-error-code-registry) — add `invalid_path_param` and `invalid_request_body` rows. Sweep through and confirm every code maps to an `ErrorCode` variant.
- [`docs/design.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/design.md) Error response pattern — add a `Code` column to the variants table; describe `ErrorCode` and the hybrid helper API in Key behaviors.
- ADR on the per-variant-enum + hybrid-helper design choice. Codes-as-stable-contract is a forever decision; future contributors deserve the reasoning documented.
- `compliance-plan.md` Document history entry referencing this Issue and its PR.

## Standards refs

- [`docs/api-contract.md` § 3](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/api-contract.md#3-error-response-contract), [§ 8](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/api-contract.md#8-error-code-registry).
- [`docs/design.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/design.md) Error response pattern.
- [`docs/coding-standards/rust.md`](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/coding-standards/rust.md) § 1 (error handling).

## Dependencies

None blocking. Best landed before any frontend pattern-matching on `code` ships — once frontend consumes the field, code names become stable contract and renames are breaking.

## Risk

Medium. Mostly mechanical (campaign across every error construction site, mirroring PR-F4's shape), but:

- The registry as forever-contract dimension means the codes chosen here are difficult to rename later. Worth a careful pass during ADR drafting.
- ~60–80 construction sites need code assignment. Easy to miss one in review; the type system gets the variants but the per-code helpers can hide a wrong choice behind the right HTTP status.
- 422 deferred (see Scope §1); a future Issue may revisit if JSON-data-validation surfaces enough to need its own status.

## Effort

M–L. Campaign style: type-level change in one commit, per-module wave updating call sites, custom extractors, then tests + docs + ADR. Comparable in shape to PR-F4 (#155).

## Verification

- Every error response across the integration test suite has both `error` and `code` keys. (Hard to assert programmatically without a snapshot pass; the test updates from Scope §7 cover the bulk.)
- New end-to-end test: malformed-UUID `GET /api/v1/sessions/not-a-uuid` returns `400 { \"error\": \"...\", \"code\": \"invalid_path_param\" }` instead of axum's plain-text default.
- Existing tests' status assertions continue to pass.

## Origin

Surfaced multiple times:

- Filed originally as #146 (bug: typed Path extractor 400s don't match the envelope).
- Re-surfaced during PR-F4 ([#155](https://github.com/brendanbyrne/beerio-kart/pull/155)) review when `gateway_timeout` was added to § 8 ahead of wiring.
- Confirmed during #146 scope-analysis 2026-05-15: doing the Path extractor fix alone would emit a `code` field on one path while every other 4xx/5xx response did not — partial rollout that creates inconsistency without enabling the actual switchboard the `code` field exists for. Decision: bundle into the broader wiring and close #146 as covered.

brendanbyrne:
Closes #157. Also closes #146 — subsumed (custom Path/Json extractors land here as one slice of the broader rollout).

## Summary

[`api-contract.md` § 3](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/api-contract.md#3-error-response-contract) has long specified the error envelope as `{ "error": "...", "code": "..." }`. The implementation only ever emitted `{ "error": "..." }` — the `code` field was deferred from launch and never wired. This PR delivers it across the entire 4xx/5xx surface in one rollout.

- New [`ErrorCode`](https://github.com/brendanbyrne/beerio-kart/blob/157/error-code-field-rollout/backend/src/error.rs) enum mirrors the [§ 8 registry](https://github.com/brendanbyrne/beerio-kart/blob/157/error-code-field-rollout/docs/api-contract.md#8-error-code-registry) one-to-one; `#[serde(rename_all = "snake_case")]` so it serializes as the registry strings. `#[non_exhaustive]`.
- `Error` variants reshape: `BadRequest`, `Conflict`, `Unauthorized`, `Forbidden` carry `code: ErrorCode`. `NotFound`, `Internal`, `Token`, `Hash`, `Timeout` keep their tuple/single shapes — their code comes from `Error::code()`.
- `ErrorBody` gains `code`; `IntoResponse` emits it for every variant including 500s and 504.
- **Hybrid helper API:** per-code helpers for named domain codes (`Error::username_taken`, `Error::session_closed`, `Error::lap_times_mismatch`, `Error::invalid_credentials`, `Error::admin_required`, ...) — typo-resistant by construction. Generic helpers (`Error::bad_request`, `Error::conflict`, `Error::forbidden`) keep their signatures and default to the generic codes for bespoke long-tail errors. 401 has no generic.
- **Custom Path/Json extractors** in [`backend/src/extract.rs`](https://github.com/brendanbyrne/beerio-kart/blob/157/error-code-field-rollout/backend/src/extract.rs) — the #146 fix. `axum::extract::Path` rejections → `invalid_path_param` (400); `axum::Json` rejections (all variants — syntax, data, content-type, bytes) → `invalid_request_body` (400). Axum's default plain-text rejection bodies never surface.
- Token-validation improvement in [`middleware/auth.rs`](https://github.com/brendanbyrne/beerio-kart/blob/157/error-code-field-rollout/backend/src/middleware/auth.rs) and [`routes/auth.rs`](https://github.com/brendanbyrne/beerio-kart/blob/157/error-code-field-rollout/backend/src/routes/auth.rs) discriminates `jsonwebtoken::errors::ErrorKind::ExpiredSignature` → `token_expired()` from other failures → `token_invalid()` — the frontend can react to `token_expired` distinctly (refresh) from `token_invalid` (re-login).
- `From<DbErr>` updated: `UniqueConstraintViolation` → `Error::Conflict { code: ErrorCode::Conflict, ... }`; `ForeignKeyConstraintViolation` → `Error::BadRequest { code: ErrorCode::BadRequest, ... }`. Specific codes like `username_taken` come from service-layer pre-checks, not the DbErr safety net.
- Docs: [`api-contract.md` § 3 + § 8](https://github.com/brendanbyrne/beerio-kart/blob/157/error-code-field-rollout/docs/api-contract.md), [`design.md` Error response pattern](https://github.com/brendanbyrne/beerio-kart/blob/157/error-code-field-rollout/docs/design.md), [ADR 0036](https://github.com/brendanbyrne/beerio-kart/blob/157/error-code-field-rollout/docs/decisions/0036-error-code-rollout.md), `compliance-plan.md` Document history.

## Design decisions

Three substantive calls were made in chat before coding; ADR 0036 captures them in full:

1. **Hybrid helper API** (Option C in the ADR) over four alternatives. Per-code helpers + generic helpers + `ErrorCode` enum, instead of `code: &'static str` per variant (no type safety), `ErrorCode` argument on every helper (call-site bloat, pairing foot-gun), or one variant per code (variant explosion).
2. **422 `unprocessable` deferred.** `JsonRejection` data-validation failures fold into 400 `invalid_request_body` for now. Splitting 400/422 is a follow-up if it earns its keep.
3. **`out_of_order_submission` registered but unused.** The pending-races validation doesn't distinguish "blocking current race" from "submitting pending out of order" — the check is identical for both. Emits `pending_races_first` for both today; the split is a future refinement if it earns its keep.

## Test plan

From repo root; from `backend/` drop `--manifest-path backend/Cargo.toml`.

- [x] **Full suite green.**
  ```bash
  cargo test --manifest-path backend/Cargo.toml
  ```
  Expect 348 tests passing (275 unit + 73 across integration crates). `db::tests` includes the existing FK enforcement check; `extract::tests` adds 7 new tests covering Path/Json rejection envelopes and happy paths; `error::tests` includes one regression-guard per per-code helper.
- [x] **Extractor verification end-to-end.**
  ```bash
  cargo test --manifest-path backend/Cargo.toml --lib extract::tests
  ```
  All 7 pass: Path<i32> with invalid segment (400 + `code: invalid_path_param`), Path<i32> happy path (200), Json malformed body (400 + `code: invalid_request_body`), Json schema mismatch (400 + same code), Json happy path (200), helper code-mapping regression guards.
- [x] **Verify code field across the wire (manual).** From `backend/`:
  ```bash
  cargo run --manifest-path Cargo.toml
  ```
  Then in another terminal:
  ```bash
  # Path extractor rejection
  curl -i http://localhost:3000/api/v1/sessions/not-a-uuid
  # Expect: 400 with body {"error":"...","code":"invalid_path_param"}

  # JSON body rejection
    -H 'content-type: application/json' \
    -d 'not json'
  # Expect: 400 with body {"error":"...","code":"invalid_request_body"}

  # Username conflict
    -H 'content-type: application/json' \
    -d '{"username":"alice","password":"password123"}'
  # ... run twice, second returns: 409 with body {"error":"Username already taken","code":"username_taken"}

  # Login wrong password
  curl -i -X POST http://localhost:3000/api/v1/auth/login \
    -H 'content-type: application/json' \
    -d '{"username":"alice","password":"wrong"}'
  # Expect: 401 with body {"error":"Invalid username or password","code":"invalid_credentials"}
  ```
- [x] **Lint clean.**
  ```bash
  cargo clippy --manifest-path backend/Cargo.toml --all-targets
  ```
  Zero warnings.


backend/src/error.rs
```diff
@@ -441,12 +789,17 @@ mod tests {
             body.contains("Resource already exists"),
             "response missing client text: {body}"
         );
+        assert!(
+            body.contains("\"code\":\"conflict\""),
+            "response missing code field: {body}"
+        );
     }
```
brendanbyrne, line 797: 🔵 **Suggestion** — Wire-shape coverage is asymmetric: `Conflict`, `BadRequest`, and `Timeout` all have tests asserting `"code":"<value>"` appears in the serialized response body. `Internal`, `Token`, and `Hash` don't — they're covered only transitively via the `code()` unit test plus the fact that `IntoResponse` reads `self.code()`. A one-test addition along the lines of:

```rust
#[tokio::test]
async fn test_internal_response_body_includes_internal_code() {
    let err: Error = anyhow::anyhow!("boom").into();
    let resp = err.into_response();
    assert_eq!(resp.status(), StatusCode::INTERNAL_SERVER_ERROR);
    let bytes = axum::body::to_bytes(resp.into_body(), usize::MAX).await.unwrap();
    let body = std::str::from_utf8(&bytes).unwrap();
    assert!(body.contains("\"code\":\"internal\""), "got: {body}");
    assert!(body.contains("Internal server error"));
}
```

would close the gap. Opinion-level — not a rule violation.

brendanbyrne (commented):
## Review — wire the `code` field across all error responses

**Verdict:** Approve. No critical or important findings; two opinion-level suggestions captured inline.

**Areas reviewed:** `docs/coding-standards/rust.md`, `docs/api-contract.md`, `docs/design.md`, ADR 0036.

**Automated checks (run locally on `aae0193`):**
- `cargo clippy --all-targets --all-features -- -D warnings` — zero warnings.
- `cargo test` — 353 passed, 0 failed, 3 ignored (doctests).

### Design-doc accuracy

`docs/design.md` § Error response pattern was the most-affected section, and it's been updated end-to-end:
- Variants table grew the `code` field column.
- `BadRequest`, `Unauthorized`, `Forbidden`, `Conflict` rows reshaped to struct variants with `code: ErrorCode`; tuple form retained only for `NotFound`.
- Response format note updated from `{ error }` to `{ error, code }`.
- New Key behaviors bullets cover the hybrid helper API and the project-local Path/Json extractors.
- Document history entry recorded.

`docs/api-contract.md` § 3 dropped the speculative `Implementation:` block (now real code) and added a pointer to ADR 0036 for the design rationale. § 8 registry grew `invalid_path_param` and `invalid_request_body` rows for the extractor failures. History entry recorded.

ADR 0036 captures four genuine alternatives (static `&'static str`; enum + every-variant argument; enum + hybrid helpers; variant-per-code) with concrete trade-offs each, plus an honest "Negative consequences" section calling out unenforced pairing and deferred items.

**Verdict: design.md needed updates, they were made, accurate.**

### What's good

- **ADR 0036 is exemplary.** Reads as a real decision record — every option has a cost the author engaged with, the chosen option's downsides are spelled out rather than minimized.
- **`test_per_code_helpers_assign_their_codes`** ([`error.rs:715`](https://github.com/brendanbyrne/beerio-kart/blob/157/error-code-field-rollout/backend/src/error.rs#L715-L758)) sweeps every helper. Exactly the regression guard the ADR's "copy-paste bug" negative consequence flags.
- **Token-expired-vs-invalid discrimination** in both [`middleware/auth.rs:53-56`](https://github.com/brendanbyrne/beerio-kart/blob/157/error-code-field-rollout/backend/src/middleware/auth.rs#L53-L56) and [`routes/auth.rs:301-306`](https://github.com/brendanbyrne/beerio-kart/blob/157/error-code-field-rollout/backend/src/routes/auth.rs#L301-L306) is consistent across both validation paths and is a real frontend UX win.
- **`From<DbErr>` stays on the generic codes** ([`error.rs:457-477`](https://github.com/brendanbyrne/beerio-kart/blob/157/error-code-field-rollout/backend/src/error.rs#L457-L477)). The explanatory comment makes "service-layer pre-checks are the path to specific codes" explicit and prevents accidental drift into per-code DB mapping.
- **Integration tests now assert on `code` for high-value cases** — `token_invalid`, `username_taken`, `invalid_credentials` (auth), `pending_races_first`, `conflict` (sessions), `gateway_timeout` end-to-end through `IntoResponse` (db_timeout).
- **`lap_times_mismatch` migration** in [`domain/numeric.rs:121`](https://github.com/brendanbyrne/beerio-kart/blob/157/error-code-field-rollout/backend/src/domain/numeric.rs#L121) is the kind of semantic upgrade per-code helpers are for — the call site now reads as the registry code.
- **Driver-string leak isolation preserved:** the `client`/`detail` split on `BadRequest`/`Conflict` survived the struct-variant refactor with new `code` field added alongside; sanitization tests ([`error.rs:767-823`](https://github.com/brendanbyrne/beerio-kart/blob/157/error-code-field-rollout/backend/src/error.rs#L767-L823)) updated to also assert on `code`.

### Findings

🔵 **Suggestion** — **Asymmetric wire-shape test coverage for 500-class variants.** [`error.rs:767`](https://github.com/brendanbyrne/beerio-kart/blob/157/error-code-field-rollout/backend/src/error.rs#L767-L823), [`error.rs:907-942`](https://github.com/brendanbyrne/beerio-kart/blob/157/error-code-field-rollout/backend/src/error.rs#L907-L942) have wire-body regression tests for `Conflict`, `BadRequest`, and `Timeout` asserting `code` is serialized correctly. There's no equivalent assertion for `Internal`, `Token`, or `Hash` confirming the wire emits `"code":"internal"`. `code()` is unit-tested ([`error.rs:715`](https://github.com/brendanbyrne/beerio-kart/blob/157/error-code-field-rollout/backend/src/error.rs#L715-L758)) and `IntoResponse` always reads `self.code()`, so this is covered transitively — but the explicit boundary guard exists for the other variants, and an `Internal` wire test would round out the coverage. One-test addition; not a rule-cited finding.

🔵 **Suggestion** — **Boundary log structured-field consistency.** [`error.rs:393, 402`](https://github.com/brendanbyrne/beerio-kart/blob/157/error-code-field-rollout/backend/src/error.rs#L393) record `?code` on the warn for `BadRequest`/`Conflict`. The error log for `Internal`/`Token`/`Hash` at [`error.rs:407`](https://github.com/brendanbyrne/beerio-kart/blob/157/error-code-field-rollout/backend/src/error.rs#L407) uses `format_error_chain(&self)` as the only message and doesn't include a structured `code` field. Since those variants always emit `ErrorCode::Internal` this isn't a defect — but a `code = ?ErrorCode::Internal` (or `?code`) field would make log aggregation/filtering uniform across all error responses. Trivial change; opinion-level. Note: line 407 is outside the diff hunks so this finding is recorded in the review body rather than as an inline comment.

### Re-litigation against the standards

Nothing in this PR violates a documented rule. The two suggestions above are opinion-level — they don't map to a `rust.md` / `seaorm.md` / `tokio.md` / `api-contract.md` section. If either becomes a pattern across other handlers (especially the boundary-log structured field), it might be worth a rule in `rust.md` § 10.

brendanbyrne:
Both suggestions addressed in c1b23c6.

**🔵 Wire-shape coverage for 500-class.** Added `test_internal_response_body_includes_internal_code` mirroring the shape of your snippet — asserts `{"code":"internal"}` appears in the body, `"Internal server error"` is the user-facing message, and the underlying anyhow chain doesn't leak to the wire (the `boom` literal). `Internal` is the representative case since `Token` and `Hash` flow through the same `IntoResponse` match arm; one test covers the arm.

**🔵 Boundary log structured-field consistency.** Added `?code` to the 500-class `error!` and the `Timeout` `warn!`. Every error-path log line now carries `code = …` uniformly. The `BadRequest`/`Conflict` warns already had it; this brings the other two into line for log aggregation/filtering. Trivial change.

Both kept inside `error.rs` — no rule violation, no `rust.md` § 10 update worth proposing yet. If we see other handlers adopt or skip the `?code` pattern in future error paths, that's the trigger to lift it into a rule.

Issue #51, opened 2026-05-06
Per-race 1-hour expiry (ADR-0035; replaces 5-minute grace)


brendanbyrne:
## Status

Reframed 2026-05-16 against [ADR-0035](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/decisions/0035-race-anchored-session-lifetime.md). The `session_race_participations` table and the `get_pending_races` query (with the 5-minute-grace semantics) were implemented during Hardening compliance work and live in `backend/src/services/sessions/races.rs`. This Issue now tracks the pending-derivation rewrite ADR-0035 requires.

## Context

ADR-0035 replaces the 5-minute rejoin grace with a per-race 1-hour submission window anchored on `session_races.created_at`. A flaked-out user can rejoin any time within that hour and their pending races are still pending — the race itself hasn't expired. The grace concept and its `joined_at`-reset companion disappear.

## Acceptance criteria

- [ ] `get_pending_races` (`backend/src/services/sessions/races.rs`): rewrite the SQL predicate per ADR-0035 § Decision outcome — drop the grace clause (`sp.left_at IS NULL OR sp.left_at >= grace_cutoff`), drop the `sr.created_at >= sp.joined_at` clause, drop the `session_participants` join, add `sr.created_at >= NOW() - INTERVAL 1 HOUR`.
- [ ] `REJOIN_GRACE_MINUTES` (`backend/src/services/sessions/types.rs`): deleted.
- [ ] Tests: drop the grace-period tests (`pending_*_grace_*`, `rejoin_*_grace*`, `multiple_short_flaps_within_grace_preserve_pending`). Add race-expiry tests: pending excludes races created more than 1 hour ago, pending includes races within the window, pending is unaffected by user `left_at` value (within the race window).

## Companion work

- #47 — lifecycle changes (schema, `join_session`, the two "are you in a session" check queries).
- #58 — sweeper predicate + 15-min cadence + `touch_session` removal.

Per ADR-0035 § Migration, the PR closing this Issue ships together with the PRs closing #47 and #58. Single combined PR with `Closes #47 #51 #58` is the natural form.

## References

- ADR-0035 — race-anchored session lifetime (`docs/decisions/0035-race-anchored-session-lifetime.md`).
- ADR-0013 — pending race cap (unchanged; UI-level concern).
- Superseded design record: `docs/designs/archive/2026-04-19-pending-races-and-grace-period.md` § 3.

---

[...]

PR #162, opened 2026-05-16
Race-anchored session lifetime — ADR-0035 (#47 #51 #58)

### LINKED ISSUE #47: Session lifecycle: race-anchored auto-close (ADR-0035)
## Status

Reframed 2026-05-16 against [ADR-0035](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/decisions/0035-race-anchored-session-lifetime.md) — race-anchored session lifetime. The original Issue scope (lifecycle endpoints: `POST /sessions`, `/join`, `/leave`, host transfer, `last_activity_at` maintenance) was implemented during Hardening compliance work and lives in `backend/src/services/sessions/lifecycle.rs`. This Issue now tracks the lifecycle-side changes ADR-0035 requires.

## Context

ADR-0035 replaces the 1-hour idle session timeout (anchored on `sessions.last_activity_at`) with race-derived liveness (anchored on `session_races.created_at`). This Issue covers the lifecycle-side simplifications: removing the `last_activity_at` column and the rejoin-grace gap math, and pushing the race-derived liveness predicate into the two "are you in a session" check queries so user lockout doesn't depend on sweep timing.

## Acceptance criteria

- [ ] `sessions` table: drop `last_activity_at` column from `backend/migration/src/m20260101_000001_initial_schema.rs` and from `backend/src/entities/sessions.rs`. Drop the supporting index (`idx_sessions_status_last_activity_at`).
- [ ] `join_session` (`backend/src/services/sessions/lifecycle.rs`): remove the gap-comparison branch and the `REJOIN_GRACE_MINUTES` comparison. Rejoin clears `left_at` unconditionally; `joined_at` is set only when the row is first inserted (becomes monotonic).
- [ ] `check_not_in_any_session` and `get_active_session_id` (same file): extend the SQL predicate with the race-derived liveness clause — `sessions.status = 'active'` AND (session has a race within the last hour OR was itself created within the last hour). Decouples user-lockout duration from sweep timing.
- [ ] Tests: existing lifecycle tests pass under the new behavior; new test asserts that a user "in" a stale session (no recent race) is not blocked from creating/joining a fresh session.

## Companion work

- #51 — pending-race derivation (per-race 1-hour expiry replaces 5-minute grace).
- #58 — sweeper predicate (race-derived staleness) + 15-min cadence + `touch_session` removal.

Per ADR-0035 § Migration, the PR closing this Issue ships together with the PRs closing #51 and #58 — schema/service/test changes are co-dependent. Single combined PR with `Closes #47 #51 #58` is the natural form.

## References

- ADR-0035 — race-anchored session lifetime (`docs/decisions/0035-race-anchored-session-lifetime.md`).
- ADR-0014 — session host transfer (cleaner under monotonic `joined_at`; unchanged behavior).
- Data model: `docs/data-model.md` § Sessions, Session Participants.

---

_Issue body rewritten 2026-05-16 by Cowork on Brendan's instruction. Original body remains accessible via the "Edited" link below._

brendanbyrne:
Closes #47 #51 #58

Implements [ADR-0035](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/decisions/0035-race-anchored-session-lifetime.md): replaces the two parallel timeout concepts — the 1-hour idle-session timeout (anchored on `sessions.last_activity_at`) and the 5-minute rejoin grace — with a single **race-anchored** timeout. Each `session_races` row is submittable for 1 hour from its `created_at`; a session is "live" while it has a race in that window, or was itself created within the last hour (the bootstrap case).

Single combined PR because the three issues are co-dependent — the schema drop, the service rewrites, and the test changes can't compile or pass in isolation.

## What changed

**#47 — lifecycle**
- Dropped the `last_activity_at` column and its index from the consolidated migration and `entities/sessions.rs`; added `idx_sessions_status_created_at` to back the new sweeper predicate.
- `join_session`: dropped the rejoin gap-comparison. `joined_at` is now monotonic — set once on first join, never reset. Host succession (ADR-0014) is cleaner under this.
- `check_not_in_any_session` / `get_active_session_id`: gained the race-derived liveness clause, so "are you in a session" no longer waits on the sweeper. The two queries are now consolidated (`check_` delegates to `get_active_session_id`) — no risk of the predicates drifting apart.

**#51 — per-race expiry**
- `get_pending_races`: dropped the grace clause, the `joined_at` clause, and the `session_participants` join; added `sr.created_at >= NOW() - 1h`. A user's `left_at` no longer affects pending access at all.
- Deleted `REJOIN_GRACE_MINUTES`; added `RACE_WINDOW_HOURS` (the single 1-hour constant, shared via `sessions/types.rs`).

**#58 — race-derived sweeper**
- `close_stale_sessions`: predicate rewritten to "session created over an hour ago AND no race created within the last hour" (the `NOT EXISTS` subquery is hand-rolled SQL per seaorm.md § 1).
- Deleted `helpers::touch_session` and all call sites (`create_session`, `join_session`, `leave_session`, `next_track`, `skip_turn`, `create_run`, `delete_run`, `skip_pending_race`) plus `SessionContext::touch`.
- Sweep cadence relaxed 5 min → 15 min.

## Beyond the literal ADR — flagged for review

- **`settle_dangling_participation`.** ADR-0035 says the liveness clause "decouples user lockout from sweep timing." It does — at the application level. But the schema has a partial unique index `idx_session_participants_one_active_session` (`UNIQUE(user_id) WHERE left_at IS NULL`) the ADR doesn't mention: the `INSERT` of a fresh participant row in `create_session`/`join_session` would still collide with a user's dangling row from a now-stale session. So this PR adds `settle_dangling_participation`, called inside the create/join transaction — it marks the user's dangling row as left so the decoupling is real (semantically: you implicitly leave an abandoned session by starting a new one). Without it, lockout would stay sweep-bound, and at 15-min cadence that's *worse* than the old 5-min behavior. See `.agents/handoffs/cowork.md` for the suggestion to record this in the ADR.
- **Wire format.** `SessionDetail` and `SessionSummary` API responses lose their `last_activity_at` field; `frontend/src/api/types.ts` is updated to match. `api-contract.md` § 4's ETag spec still references `session.last_activity_at` and needs a design revision (it was the only ETag input that captured participant joins/leaves) — flagged for Cowork, not touched here.
- **`.editorconfig`.** Added a 2-space override for frontend source extensions. The global `indent_size = 4` is correct for Rust but Prettier (which honors `.editorconfig`) was therefore demanding 4-space for the 2-space frontend codebase. This is a pre-existing latent bug surfaced by touching `types.ts`; fix confirmed with Brendan.

## Out of scope (Cowork's lane)

`docs/data-model.md`, `docs/user-workflows.md` § 1.5, and `docs/api-contract.md` § 4 — see `.agents/handoffs/cowork.md`.

## Testing

`cargo test` — 279 unit + 27/21/8/… integration tests pass; `cargo clippy --tests` clean; `bun run typecheck` clean.

Test changes: dropped the five grace tests (`pending_*_grace_*`, `rejoin_*_grace*`, `multiple_short_flaps_within_grace_*`), the `touch_session` / `SessionContext::touch` tests, and the `last_activity_at`-movement assertions. Added per-race-expiry tests (`get_pending_races` excludes expired races, includes races in-window, is unaffected by `left_at`), sweeper tests (`close_stale_sessions` for the all-races-expired / recent-race / fresh-bootstrap / old-bootstrap cases), a monotonic-`joined_at` rejoin test, and the lockout-decoupling tests (a user in a stale session can create/join a fresh one).

### Manual test steps

The 1-hour expiry and 15-min sweep are timer-driven and covered by unit tests (which backdate timestamps); the steps below cover what's observable live.

1. **Setup:** `cd backend && cargo run` (in another terminal `cd frontend && bun run dev`). The dev SQLite DB is recreated from the migration on boot — if you have an old DB file, delete it first (the schema changed).
2. **Create + join:** Register two users in two browsers. User A creates a session; user B joins it. Both should see two participants.
3. **Leave + rejoin keeps you in:** User B leaves, then rejoins the same session. B should rejoin cleanly with no "already in a session" error, and the participant list should still show both.
4. **One active session at a time:** While B is in A's session, B tries to create a new session → expect a 409 "Already in session …".
5. **Leave frees you:** B leaves A's session, then creates their own → should succeed.
6. **Pending races survive a leave:** A picks a track (`next_track`). B leaves the session *without* submitting. B rejoins → B's pending race for that track should still be listed (no 5-minute grace cutoff anymore).
7. **Host succession unchanged:** In a 3-person session, the host leaves → host role transfers to the earliest-joined remaining participant; session stays active.

### DIFF
--- a/backend/migration/src/m20260101_000001_initial_schema.rs
+++ b/backend/migration/src/m20260101_000001_initial_schema.rs
@@ -263,15 +263,18 @@ impl MigrationTrait for Migration {
                     ruleset TEXT NOT NULL,
                     least_played_drink_category TEXT,
                     status TEXT NOT NULL,
-                    created_at datetime_text NOT NULL,
-                    last_activity_at datetime_text NOT NULL
+                    created_at datetime_text NOT NULL
                 )",
         )
         .await?;
 
+        // Supports the race-derived stale-session sweeper (ADR-0035):
+        // `close_stale_sessions` filters `status = 'active' AND created_at <
+        // cutoff`. The session has no maintained activity column anymore —
+        // liveness is derived from `session_races.created_at`.
         conn.execute_unprepared(
-            "CREATE INDEX idx_sessions_status_last_activity
-                 ON sessions(status, last_activity_at)",
+            "CREATE INDEX idx_sessions_status_created_at
+                 ON sessions(status, created_at)",
         )
         .await?;
 
--- a/backend/migration/src/m20260101_000001_initial_schema.rs
+++ b/backend/migration/src/m20260101_000001_initial_schema.rs
@@ -328,6 +328,19 @@ impl MigrationTrait for Migration {
         )
         .await?;
 
+        // Backs the race-derived liveness subqueries (ADR-0035): the
+        // `EXISTS` in `get_active_session_id` and the `NOT EXISTS` in
+        // `close_stale_sessions` both correlate on
+        // `session_id = ? AND created_at >= ?`. The `UNIQUE(session_id,
+        // race_number)` index above can seek `session_id` but leaves the
+        // `created_at` filter as a residual scan; this index covers both.
+        // Also backs the `list_active_sessions` JOIN on `session_id`.
+        conn.execute_unprepared(
+            "CREATE INDEX idx_session_races_session_created
+                 ON session_races(session_id, created_at)",
+        )
+        .await?;
+
         // ---- Session race participations ----
         //
         // One row per (race, user) for every user present at race-creation
--- a/backend/src/services/sessions/lifecycle.rs
+++ b/backend/src/services/sessions/lifecycle.rs
@@ -119,9 +119,14 @@ pub async fn get_active_session_id(
 /// joining a new one. The stale session's `status` and any other participants'
 /// rows are left for the sweeper (eventual consistency — its remaining job).
 ///
-/// **Precondition:** the caller has already passed `check_not_in_any_session`,
-/// so any dangling row found here belongs to a session with no race-derived
-/// liveness left. Settling it is therefore always correct.
+/// **Why this is safe:** the caller has already passed
+/// `check_not_in_any_session`, so any dangling row found here belonged to a
+/// race-derived-stale session *as of that check*. The check runs before the
+/// transaction, so a concurrent `next_track` from another participant could
+/// in principle revive that session in the gap — but settling the row is
+/// still correct regardless: the user is explicitly starting or joining a
+/// new session, and the one-active-session invariant requires them to leave
+/// the old one either way.
 ///
 /// # Errors
 ///
@@ -554,7 +559,7 @@ struct StaleSessionRow {
 #[tracing::instrument(skip(db))]
 pub async fn close_stale_sessions(db: &DatabaseConnection) -> Result<u64, Error> {
     let now = Utc::now().naive_utc();
-    let window_start = (Utc::now() - chrono::Duration::hours(RACE_WINDOW_HOURS)).naive_utc();
+    let window_start = now - chrono::Duration::hours(RACE_WINDOW_HOURS);
 
     let txn = db_txn(db.begin()).await?;
 
@@ -1137,6 +1142,17 @@ mod tests {
         create_session(&db, &user_id, "random")
             .await
             .expect("user in a stale session may create a fresh one");
+
+        // The dangling row in the stale session was settled — this pins
+        // `settle_dangling_participation`'s direct effect, not just the
+        // side effect that the new-session INSERT didn't collide.
+        assert!(
+            participant_row(&db, &stale_id, &user_id)
+                .await
+                .left_at
+                .is_some(),
+            "the dangling row in the stale session is settled"
+        );
     }
 
     #[tokio::test]
@@ -1157,5 +1173,13 @@ mod tests {
         join_session(&db, &fresh.id, &user_id)
             .await
             .expect("user in a stale session may join a fresh one");
+
+        assert!(
+            participant_row(&db, &stale_id, &user_id)
+                .await
+                .left_at
+                .is_some(),
+            "the dangling row in the stale session is settled"
+        );
     }
 }
--- a/backend/src/services/sessions/lifecycle.rs
+++ b/backend/src/services/sessions/lifecycle.rs
@@ -4,17 +4,17 @@
 //! Detail aggregation lives in [`super::detail`]; race orchestration in
 //! [`super::races`].
 
-use chrono::{DateTime, NaiveDateTime, Utc};
+use chrono::Utc;
 use sea_orm::{
     ActiveModelTrait, ActiveValue::NotSet, ColumnTrait, Condition, ConnectionTrait,
-    DatabaseConnection, EntityTrait, FromQueryResult, PaginatorTrait, QueryFilter, QueryOrder,
-    QuerySelect, Set, TransactionTrait, sea_query::Expr,
+    DatabaseConnection, EntityTrait, FromQueryResult, PaginatorTrait, QueryFilter, QueryOrder, Set,
+    TransactionTrait, sea_query::Expr,
 };
 use uuid::Uuid;
 
 use super::{
     detail::{SessionDetail, get_session_detail},
-    types::REJOIN_GRACE_MINUTES,
+    types::RACE_WINDOW_HOURS,
 };
 use crate::{
     domain::{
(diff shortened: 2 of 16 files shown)


backend/migration/src/m20260101_000001_initial_schema.rs
```diff
@@ -263,15 +263,18 @@ impl MigrationTrait for Migration {
...
 
+        // Supports the race-derived stale-session sweeper (ADR-0035):
+        // `close_stale_sessions` filters `status = 'active' AND created_at <
+        // cutoff`. The session has no maintained activity column anymore —
+        // liveness is derived from `session_races.created_at`.
         conn.execute_unprepared(
-            "CREATE INDEX idx_sessions_status_last_activity
-                 ON sessions(status, last_activity_at)",
+            "CREATE INDEX idx_sessions_status_created_at
+                 ON sessions(status, created_at)",
```
brendanbyrne, line 277: 🔵 **Suggestion** — the two new `session_races` subqueries have no backing index. `get_active_session_id` (`EXISTS`) and `close_stale_sessions` (`NOT EXISTS`) both correlate on `session_races.session_id = ... AND sr.created_at >= ...`, but `session_races` has only its PK index — `session_id` is unindexed, so each subquery scans the table per outer row. This PR adds `idx_sessions_status_created_at` for the sweeper's *outer* query while leaving the subquery side unindexed. Recommend adding, in this same migration edit:

```sql
CREATE INDEX idx_session_races_session_created
    ON session_races(session_id, created_at)
```

It also backs the pre-existing `list_active_sessions` JOIN. Negligible at current scale, but no rule in `seaorm.md` maps to this so it's a suggestion — flagged as a standards gap in the review body.

backend/src/services/sessions/lifecycle.rs
```diff
@@ -488,30 +553,38 @@ pub async fn leave_session(
 /// UPDATE statements that drive the cleanup.
 #[tracing::instrument(skip(db))]
 pub async fn close_stale_sessions(db: &DatabaseConnection) -> Result<u64, Error> {
-    let one_hour_ago = (Utc::now() - chrono::Duration::hours(1)).naive_utc();
     let now = Utc::now().naive_utc();
+    let window_start = (Utc::now() - chrono::Duration::hours(RACE_WINDOW_HOURS)).naive_utc();
```
brendanbyrne: 🔵 **Suggestion** — the clock is read twice (`Utc::now()` on line 556 for `now`, again here for `window_start`). Derive the cutoff from the first read so the `left_at` stamp and the staleness cutoff share one instant:

```rust
let now = Utc::now().naive_utc();
let window_start = now - chrono::Duration::hours(RACE_WINDOW_HOURS);
```

backend/src/services/sessions/lifecycle.rs
```diff
@@ -851,190 +922,240 @@ mod tests {
...
-        backdate_participant(&db, &session.id, &user_b, None, Some(two_min_ago)).await;
-        join_session(&db, &session.id, &user_b).await.unwrap();
+        // Sanity: the stale session is still `active` in the DB.
+        assert_eq!(session_status(&db, &stale_id).await, SessionStatus::Active);
 
-        let pending = get_pending_races(&db, &session.id, &user_b).await.unwrap();
-        assert_eq!(pending.len(), 1, "after first short flap, pending intact");
+        create_session(&db, &user_id, "random")
+            .await
+            .expect("user in a stale session may create a fresh one");
```
brendanbyrne, line 1144: 🔵 **Suggestion** — this test proves `create_session` *succeeds* (so the partial-unique index didn't collide), but never asserts the old stale-session row was actually settled. An explicit assertion would pin `settle_dangling_participation`'s direct effect rather than just its side effect:

```rust
assert!(
    participant_row(&db, &stale_id, &user_id).await.left_at.is_some(),
    "the dangling row in the stale session is settled",
);
```

brendanbyrne (commented):
## Code review — race-anchored session lifetime (ADR-0035)

**Verdict: Approve** (no blocking findings; 3 suggestions inline).

**What this PR does:** Replaces the two-timeout model (1-hour `last_activity_at` idle sweeper + 5-minute rejoin grace) with a single race-anchored 1-hour window. Drops the `last_activity_at` column and seven `touch_session` call sites; rewrites the sweeper and `get_pending_races` around `session_races.created_at`; pushes a race-derived liveness clause into the "are you in a session" read paths.

**Verified:**
- `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings` — clean.
- `cargo test` — 279 unit + 27/21/8/19 integration, 0 failed.
- `bun run typecheck`, `bun run lint` — clean.
- No ETag code exists in the backend yet, so dropping `last_activity_at` breaks no code — the `api-contract.md` § 4 ETag concern is doc-only.

**Design-doc accuracy:** `data-model.md` (sessions table / Pending Race Tracking), `user-workflows.md` § 1.5, and `api-contract.md` § 4 need updates and are deliberately deferred to a doc-only commit per ADR-0035 § Migration step 4, handed to Cowork. This is sanctioned by the signed-off ADR — not a finding — but the deferral leaves those docs stale until that commit lands; worth landing it close to this merge.

**Minor (review-body, not line-anchorable):** `settle_dangling_participation`'s doc comment states "any dangling row found here belongs to a session with no race-derived liveness left … Settling it is therefore always correct." The conclusion holds (the user is explicitly starting a new session, so the one-active-session invariant requires removing them), but the premise doesn't strictly — `check_not_in_any_session` runs pre-transaction, and a concurrent `next_track` from another participant could revive the stale session between the check and the settle. Consider softening the wording so a future reader doesn't build on "the row is provably dead."

**Standards gap:** `seaorm.md` has no rule on indexing query predicates — the missing-index suggestion below had no rule to cite. Worth a future `seaorm.md` addition.

**What's good:** Consolidating `check_not_in_any_session` into `get_active_session_id` eliminates predicate drift between the two. `settle_dangling_participation` is correctly ordered before the INSERT inside the transaction, with an excellent doc comment on the partial-unique-index reasoning. `leave_session` now only writes the session row when a column actually changes. Raw SQL uses bound params throughout (`seaorm.md` § 10). Test coverage is thorough — bootstrap / recent-race / all-expired sweeper cases, monotonic `joined_at`, and both lockout-decoupling paths.

brendanbyrne:
Thanks for the review — all three suggestions plus the doc-wording note addressed in 75ebe4b:

- **`session_races` index** — added `idx_session_races_session_created (session_id, created_at)` in the same migration edit. Backs both liveness subqueries' `created_at` residual and the `list_active_sessions` JOIN.
- **Double clock read** — `close_stale_sessions` now derives `window_start` from the single `now`, so the `left_at` stamp and the staleness cutoff share one instant.
- **Settle assertion** — added the explicit `left_at.is_some()` assertion to *both* lockout-decoupling tests (create and join), pinning `settle_dangling_participation`'s direct effect.
- **Doc wording** — reworded `settle_dangling_participation`'s precondition: the TOCTOU gap (a concurrent `next_track` reviving the stale session between the pre-transaction check and the settle) is now called out, with the note that settling stays correct regardless because the user is leaving that session by acting here.

`cargo test` (279 unit + integration), `clippy`, `fmt` all green.

Issue #164, opened 2026-05-16
Notifications system MVP — backend (ADR-0038)


brendanbyrne:
## Status

New 2026-05-16, filed per [ADR-0038](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/decisions/0038-notifications-system.md) sign-off. Covers the backend half of the notifications MVP: schema, entity, Rust modeling, service helper, API endpoints, and the first consumer trigger (the `PendingRacesDropped` event from [ADR-0037](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/decisions/0037-pending-races-dropped-on-session-close.md)'s session-close path).

## Context

ADR-0038 introduces a notifications system as a general inbox abstraction — per-user persisted event records with kind / payload / read state, delivered via the existing polling pattern (per ADR-0018). The first consumer is ADR-0037's pending-races-dropped event: when a session closes with unresolved pending races, the affected users get a notification on their next sign-in.

The MVP backend is table + helper + endpoints + one trigger site. The other notification variants (`H2hLeadChanged`, `TrackRecordLost`, `LeaderboardRankChanged`) are forward-pointer bullets in the Banana and Leaf scope sections in `docs/roadmap.md` and land with those cups.

## Acceptance criteria

- [ ] `notifications` table added to `backend/migration/src/m20260101_000001_initial_schema.rs` per ADR-0038 § Schema. Columns: `id`, `user_id`, `kind`, `payload`, `created_at`, `read_at`. Indexes: `idx_notifications_user_unread` (partial, `WHERE read_at IS NULL`) and `idx_notifications_user_created`. FK on `user_id` with cascade delete on user removal.
- [ ] Entity `backend/src/entities/notifications.rs` matches the schema.
- [ ] `NotificationPayload` enum (in `backend/src/services/notifications.rs` or similar) carries the `PendingRacesDropped { session_id, dropped_count }` variant only for MVP. Serde tagging: `#[serde(tag = "kind", rename_all = "snake_case")]`. See ADR-0038 § Rust modeling for the full variant list (other variants land later under Banana / Leaf).
- [ ] `services::notifications::record(txn, user_id, payload)` helper that serializes the payload as JSON and inserts a row inside the caller's transaction.
- [ ] API endpoints per ADR-0038 § API surface:
  - `GET /me/notifications` — list unread, newest first; cursor pagination per ADR-0032; optional `?include_read=true`.
[...]

Issue #163, opened 2026-05-16
Notifications UI MVP — bell, dropdown, PendingRacesDropped render (ADR-0038)


brendanbyrne:
## Status

New 2026-05-16, filed per [ADR-0038](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/decisions/0038-notifications-system.md) sign-off. Covers the frontend half of the notifications MVP: the bell icon, the unread badge, the dropdown listing notifications, and the render function for the `PendingRacesDropped` variant.

## Context

ADR-0038 specifies a polling-based notification surface. The home screen polls `GET /me/notifications/unread-count` for the badge — low frequency (every ~30s, or on navigation) — *not* on the 2.5s session-detail polling cadence. The dropdown fetches `GET /me/notifications` on open. Mark-as-read happens via `POST /me/notifications/read-all` on dropdown close (bulk-dismiss-on-view per ADR-0038 § API surface).

The MVP frontend only needs to render the `PendingRacesDropped` variant. Other variants (`H2hLeadChanged`, `TrackRecordLost`, `LeaderboardRankChanged`) land with the Banana / Leaf cups per the roadmap.

## Acceptance criteria

- [ ] Bell icon component on the home and session screens with a badge showing the unread count.
- [ ] `useNotifications()` hook (or equivalent) that:
  - Polls `GET /me/notifications/unread-count` every ~30s. Pauses when the tab is backgrounded (Page Visibility API), matching the existing session-polling behavior.
  - Fetches `GET /me/notifications` on dropdown open.
  - Calls `POST /me/notifications/read-all` on dropdown close.
- [ ] Render function for the `PendingRacesDropped` variant. Copy something like "Session X ended with N submissions you didn't get to." Bulk-dismiss-on-view handles dismissal; no per-row dismiss UI in MVP.
- [ ] TypeScript types in `frontend/src/api/types.ts` for `NotificationPayload` and the `PendingRacesDropped` variant (hand-written, per ADR-0038 § Rust modeling — Frontend type agreement). May be filed-and-merged as part of the backend Issue's PR; tracking here for visibility.
- [ ] Polling lives at the app shell / home-screen level — separate cadence from in-session polling. Don't fold notifications fetching into the existing session-detail poll loop.
- [ ] No real-time push. Per ADR-0018, polling is the chosen pattern.

## Depends on

[...]

PR #165, opened 2026-05-17
164: drop pending races on session close + notifications MVP

### LINKED ISSUE #47: Session lifecycle: race-anchored auto-close (ADR-0035)

brendanbyrne:
## Reviewer notes

Single combined PR — the schema (`dropped_at`), the drop-on-close step, the `get_pending_races` simplification, the sweeper-predicate extension, the `notifications` table, and the `record_pending_drops` consumer are co-dependent; splitting them would force intermediate states that don't compile or don't pass tests.

**One deviation from ADR-0038:** `GET /me/notifications` uses a flat `LIMIT 100` cap instead of ADR-0032 cursor pagination. No keyset-pagination infrastructure exists anywhere in the codebase — `GET /runs`, ADR-0032's own named exemplar, also still uses a flat cap. Building keyset pagination for one endpoint in isolation would make notifications *inconsistent* with the sibling list endpoint. ADR-0032 explicitly sanctions a flat substitute "with a note about the tradeoff." Surfaced in `NOTIFICATIONS_PAGE_LIMIT`'s doc comment and api-contract.md § 1.8. Cowork reviewed the deviation and chose to keep ADR-0032 `accepted` and track project-wide keyset rollout as a follow-up — now filed as **#166**, with an `## Implementation status` section added to ADR-0032 itself.

`#47` is already closed (it tracked the ADR-0035 work that merged); the `Closes` keyword below is a harmless no-op for it. Frontend (#163) is a separate downstream PR — untouched here.

The sweeper now treats a recent join/leave/run/skip as a liveness signal, so three pre-existing sweeper tests had fixtures updated (participants that "joined now" in a backdated session would otherwise keep it alive — correct behavior, unrealistic fixture).

## Summary

Implements [ADR-0037](../blob/main/docs/decisions/0037-pending-races-dropped-on-session-close.md) (pending races dropped on session close) and the [ADR-0038](../blob/main/docs/decisions/0038-notifications-system.md) notifications-system MVP backend. When a session closes — clean last-leave or stale-session sweeper — every unresolved pending race is stamped `dropped_at` and each affected user gets a `pending_races_dropped` notification, all atomic with the close. `get_pending_races` drops its per-race timer and session-status filter (the session is the deadline); the sweeper predicate gains four activity signals.

## Linked work

- Closes #47 #51 #58 #164
- Implements docs/decisions/0037-pending-races-dropped-on-session-close.md
- Implements docs/decisions/0038-notifications-system.md
- Follow-up: #166 — project-wide ADR-0032 keyset pagination (the deliberate `LIMIT 100` deviation; see Reviewer notes)

## How to verify

Backend unit + integration tests (297 lib + integration suites):

```bash
cd backend && cargo test
```
- [x] All tests pass, including the new `services::notifications`, `lifecycle` drop/sweeper/notification-trigger, and `session_integration` notification-flow tests.

Lint + format:

```bash
cd backend && cargo clippy --all-targets && cargo +nightly fmt --check
```
- [x] Clean.

ADR-0037 spec-conformance greps (each should match the stated expectation):

```bash
grep -rn "created_at.*INTERVAL.*HOUR\|created_at.*now()" backend/src/services/sessions/races.rs   # expect: empty
grep -rn "s.status = 'active'" backend/src/services/sessions/races.rs                              # expect: empty
grep -rln "dropped_at" backend/migration backend/src/entities backend/src/services                # expect: migration, entity, helpers, lifecycle, races
```
- [ ] Per-race expiry and session-status clauses gone from `get_pending_races`; `dropped_at` present at the migration/entity/service sites.

End-to-end manual check (drop-on-close → notification inbox). Reset the dev DB first since the schema changed:

```bash
rm -f data/db/beerio-kart.db && cd backend && cargo run
```

Then, against `http://localhost:3000` — each step captures what the next needs (`TOKEN`, `SESSION_ID`) into shell variables via `jq`, so the block runs top-to-bottom as-is (requires `jq`):

```bash
# 1. Register a user — capture the access token
TOKEN=$(curl -s -X POST localhost:3000/api/v1/auth/register \
  -H 'content-type: application/json' \
  -d '{"username":"dropper","password":"testpass123"}' | jq -r .access_token)

# 2. Create a session — capture its id — then pick a track (host has 1 pending race now)
SESSION_ID=$(curl -s -X POST localhost:3000/api/v1/sessions \
  -H "authorization: Bearer $TOKEN" -H 'content-type: application/json' \
  -d '{"ruleset":"random"}' | jq -r .id)
echo "TOKEN=${TOKEN:0:12}…  SESSION_ID=$SESSION_ID"   # sanity-check the captures
curl -s -X POST "localhost:3000/api/v1/sessions/$SESSION_ID/next-track" \
  -H "authorization: Bearer $TOKEN" | jq

# 3. Inbox is empty before any close
curl -s localhost:3000/api/v1/me/notifications -H "authorization: Bearer $TOKEN" | jq

# 4. Leave as the solo participant — session closes, pending race drops
curl -s -X POST "localhost:3000/api/v1/sessions/$SESSION_ID/leave" \
  -H "authorization: Bearer $TOKEN"

# 5. Unread count is now 1
curl -s localhost:3000/api/v1/me/notifications/unread-count \
  -H "authorization: Bearer $TOKEN" | jq

# 6. List shows the pending_races_dropped notification, dropped_count 1
curl -s localhost:3000/api/v1/me/notifications -H "authorization: Bearer $TOKEN" | jq

# 7. Mark all read → 204; unread count back to 0; default list empty;
#    ?include_read=true still returns it with read_at set
curl -s -i -X POST localhost:3000/api/v1/me/notifications/read-all \
  -H "authorization: Bearer $TOKEN"
curl -s localhost:3000/api/v1/me/notifications/unread-count \
  -H "authorization: Bearer $TOKEN" | jq
curl -s "localhost:3000/api/v1/me/notifications?include_read=true" \
  -H "authorization: Bearer $TOKEN" | jq
```
- [ ] Step 3: `[]`. Step 5: `{"count":1}`. Step 6: one item, `payload.kind = "pending_races_dropped"`, `payload.session_id` matches `SESSION_ID`, `payload.dropped_count = 1`, `read_at` null. Step 7: `204`, then `{"count":0}`, then `[]` for the default list and one item with a non-null `read_at` for `?include_read=true`.

## Author checklist

- [x] Tests added or updated for new logic (per `CLAUDE.md`).
- [x] Docs updated (`data-model.md`, `user-workflows.md`, `api-contract.md` — all with `Document history` entries).
- [x] Schema changes: consolidated migration edited; schema-drift test extended; local DB reset to verify.
- [x] Tested locally end-to-end.

### DIFF
--- a/backend/src/routes/notifications.rs
+++ b/backend/src/routes/notifications.rs
@@ -1,0 +1,76 @@
+use axum::{
+    extract::{Query, State},
+    http::StatusCode,
+    response::IntoResponse,
+};
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    AppState, error::Error, extract::Json, middleware::auth::User, services::notifications,
+};
+
+/// Query string for `GET /me/notifications`.
+#[derive(Deserialize)]
+pub struct ListNotificationsQuery {
+    /// When `true`, include already-read notifications. Defaults to `false`
+    /// (unread only) — the home-screen dropdown's day-one shape.
+    #[serde(default)]
+    pub include_read: bool,
+}
+
+/// `GET /me/notifications` — list the authenticated user's notifications,
+/// newest first. Unread-only unless `?include_read=true`.
+///
+/// # Errors
+///
+/// Propagates the errors of [`notifications::list_notifications`] — `Internal`
+/// for unexpected DB failures or a corrupt stored payload.
+#[tracing::instrument(skip_all, fields(user_id = %user.user_id, include_read = query.include_read))]
+pub async fn list_notifications(
+    user: User,
+    State(state): State<AppState>,
+    Query(query): Query<ListNotificationsQuery>,
+) -> Result<Json<Vec<notifications::NotificationView>>, Error> {
+    let list =
+        notifications::list_notifications(&state.db, &user.user_id, query.include_read).await?;
+    Ok(Json(list))
+}
+
+/// Response body for `GET /me/notifications/unread-count`.
+#[derive(Serialize)]
+pub struct UnreadCountResponse {
+    /// Number of unread notifications for the authenticated user.
+    pub count: u64,
+}
+
+/// `GET /me/notifications/unread-count` — cheap unread tally for the
+/// home-screen badge poll.
+///
+/// # Errors
+///
+/// Propagates the errors of [`notifications::unread_count`] — `Internal` for
+/// unexpected DB failures.
+#[tracing::instrument(skip_all, fields(user_id = %user.user_id))]
+pub async fn unread_count(
+    user: User,
+    State(state): State<AppState>,
+) -> Result<Json<UnreadCountResponse>, Error> {
+    let count = notifications::unread_count(&state.db, &user.user_id).await?;
+    Ok(Json(UnreadCountResponse { count }))
+}
+
+/// `POST /me/notifications/read-all` — mark all of the authenticated user's
+/// unread notifications as read. Returns `204 No Content`.
+///
+/// # Errors
+///
+/// Propagates the errors of [`notifications::mark_all_read`] — `Internal` for
+/// unexpected DB failures.
+#[tracing::instrument(skip_all, fields(user_id = %user.user_id))]
+pub async fn mark_all_read(
+    user: User,
+    State(state): State<AppState>,
+) -> Result<impl IntoResponse, Error> {
+    notifications::mark_all_read(&state.db, &user.user_id).await?;
+    Ok(StatusCode::NO_CONTENT)
+}
--- a/backend/src/services/notifications.rs
+++ b/backend/src/services/notifications.rs
@@ -151,10 +151,10 @@ pub async fn record(
 /// # Errors
 ///
 /// Propagates the errors of [`record`].
-#[tracing::instrument(
-    skip(txn),
-    fields(user_id = %user_id, session_id = %session_id, dropped_count),
-)]
+// `dropped_count` is intentionally not named in `fields(...)` — a bare name
+// there declares an empty span field that nothing records. As an unskipped
+// `Copy` argument it is auto-captured into the span by `#[instrument]`.
+#[tracing::instrument(skip(txn), fields(user_id = %user_id, session_id = %session_id))]
 pub async fn record_pending_drops(
     txn: &impl ConnectionTrait,
     user_id: &UserId,
--- a/backend/src/services/notifications.rs
+++ b/backend/src/services/notifications.rs
@@ -1,0 +1,442 @@
+//! Notifications service (ADR-0038).
+//!
+//! A per-user inbox of asynchronous events. Each event materializes one
+//! `notifications` row inside the same transaction as the triggering write —
+//! no worker, no queue, no async dispatch (ADR-0038 § Generation pattern).
+//!
+//! The module exposes:
+//! - [`NotificationPayload`] — the closed, serde-tagged enum of event kinds.
+//! - [`record`] — the generic typed constructor every event site calls.
+//! - [`record_pending_drops`] — the ADR-0037 consumer (pending races dropped
+//!   on session close).
+//! - [`list_notifications`] / [`unread_count`] / [`mark_all_read`] — the read
+//!   and dismiss surface backing the three `/me/notifications` endpoints.
+//!
+//! For MVP the enum carries a single variant (`PendingRacesDropped`); the
+//! other kinds named in ADR-0038 (`H2hLeadChanged`, `TrackRecordLost`,
+//! `LeaderboardRankChanged`) are deliberately deferred to future cups.
+
+use chrono::{DateTime, Utc};
+use sea_orm::{
+    ActiveModelTrait, ActiveValue::NotSet, ColumnTrait, ConnectionTrait, EntityTrait,
+    PaginatorTrait, QueryFilter, QueryOrder, QuerySelect, Set, sea_query::Expr,
+};
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    domain::{NotificationId, SessionId, UserId},
+    entities::notifications,
+    error::Error,
+    timeout::db_query,
+};
+
+/// Hard cap on rows returned by [`list_notifications`].
+///
+/// ADR-0038 § API surface specifies cursor pagination per ADR-0032 for
+/// `GET /me/notifications`. No cursor-pagination infrastructure exists in the
+/// codebase yet — `GET /runs` (the ADR-0032 exemplar) also still uses a flat
+/// `LIMIT 100` cap. This endpoint follows that precedent rather than building
+/// keyset pagination for one endpoint in isolation; revisiting it project-wide
+/// is tracked as a follow-up. At friend-group scale (ADR-0038 estimates
+/// ~18k notification rows/year across ~10 users) a 100-row cap comfortably
+/// covers the unread inbox and a deep history scroll.
+const NOTIFICATIONS_PAGE_LIMIT: u64 = 100;
+
+/// Closed set of notification event kinds, with serde-tagged serialization.
+///
+/// `kind` is the serde tag — lifted into the `notifications.kind` column for
+/// indexing — and the rest of each variant body is the JSON payload stored in
+/// `notifications.payload`. Adding a kind is adding a variant: the compiler
+/// then surfaces every site that pattern-matches.
+///
+/// MVP carries only [`Self::PendingRacesDropped`]. The future variants from
+/// ADR-0038 § Rust modeling land with their respective cups.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(tag = "kind", rename_all = "snake_case")]
+pub enum NotificationPayload {
+    /// The user's session closed while they still had unresolved pending
+    /// races; those races were dropped (ADR-0037).
+    PendingRacesDropped {
+        /// The session that closed.
+        session_id: SessionId,
+        /// How many of the user's pending races were dropped.
+        dropped_count: u32,
+    },
+}
+
+impl NotificationPayload {
+    /// The `snake_case` discriminator stored in the `kind` column.
+    ///
+    /// Mirrors the serde tag — keep the two in lockstep when adding variants.
+    const fn kind_str(&self) -> &'static str {
+        match self {
+            Self::PendingRacesDropped { .. } => "pending_races_dropped",
+        }
+    }
+}
+
+/// Wire view of a single notification — the JSON shape returned by
+/// `GET /me/notifications`.
+///
+/// The stored `kind` column is intentionally not surfaced separately: it is
+/// already present as the serde tag inside `payload`.
+#[derive(Debug, Serialize)]
+pub struct NotificationView {
+    /// Stable UUID of the notification row.
+    pub id: NotificationId,
+    /// When the event was recorded, UTC.
+    pub created_at: DateTime<Utc>,
+    /// When the user dismissed it, or `None` while unread.
+    pub read_at: Option<DateTime<Utc>>,
+    /// Kind-tagged structured payload.
+    pub payload: NotificationPayload,
+}
+
+/// Parse a stored notification row into its wire view.
+///
+/// A `payload` column that doesn't deserialize as a [`NotificationPayload`]
+/// is data corruption (hand-edited row, or a write from a future schema) and
+/// surfaces as `Internal`.
+fn row_into_view(row: &notifications::Model) -> Result<NotificationView, Error> {
+    let payload: NotificationPayload = serde_json::from_str(&row.payload).map_err(|e| {
+        Error::Internal(anyhow::Error::new(e).context("Deserializing notification payload"))
+    })?;
+    Ok(NotificationView {
+        id: NotificationId::from_db(&row.id)?,
+        created_at: row.created_at.and_utc(),
+        read_at: row.read_at.map(|t| t.and_utc()),
+        payload,
+    })
+}
+
+/// Record a notification for `user_id`.
+///
+/// The INSERT runs on whatever connection / transaction handle is passed: an
+/// event site that already holds a transaction passes it so the notification
+/// is atomic with the triggering write (ADR-0038 § Atomicity).
+///
+/// # Errors
+///
+/// Returns `Internal` if the payload fails to serialize (should not happen
+/// for the closed enum) or for unexpected DB failures on the INSERT.
+#[tracing::instrument(skip(txn, payload), fields(user_id = %user_id, kind = payload.kind_str()))]
+pub async fn record(
+    txn: &impl ConnectionTrait,
+    user_id: &UserId,
+    payload: &NotificationPayload,
+) -> Result<(), Error> {
+    let payload_json = serde_json::to_string(payload).map_err(|e| {
+        Error::Internal(anyhow::Error::new(e).context("Serializing notification payload"))
+    })?;
+
+    db_query(
+        notifications::ActiveModel {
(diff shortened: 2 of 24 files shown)


backend/src/services/sessions/lifecycle.rs
```diff
@@ -462,6 +462,111 @@ async fn transfer_host_or_close(
...
+    .await?;
+
+    if drops.is_empty() {
+        // Nothing unresolved — a session can close with every race already
+        // raced or skipped. No drops, no notifications.
+        return Ok(());
+    }
+
+    // Stamp `dropped_at` on every unresolved pending row in this session, in
+    // one set-based UPDATE. The predicate matches the count query above.
```
brendanbyrne: 🔵 **Suggestion** — the `skipped_at IS NULL AND dropped_at IS NULL AND NOT EXISTS(runs…)` predicate is written verbatim here and in the count query above. If they ever drift, a notification's `dropped_count` silently mismatches the rows actually stamped `dropped_at`. The "predicate matches the count query above" comment is the current mitigation; lifting the shared `WHERE` body into a `const &str` fragment that both statements interpolate would make drift impossible. Low severity — both statements are in one function and easy to eyeball today.

brendanbyrne (commented):
## Code Review — ADR-0037 drop-on-close + ADR-0038 notifications MVP

**Verdict: Approve.** High-quality PR; all findings below are suggestions, none block merge.

### Automated checks
- `cargo clippy --all-targets --all-features -D warnings` — clean
- `cargo test` — 377 passed, 0 failed (incl. new per-signal sweeper, drop-on-close, cross-user isolation, full-flow integration tests)
- `cargo fmt --check` — clean (stable-channel warnings for nightly-only options are the known project setup)
- frontend `bun run lint` / `tsc --noEmit` — clean

### Findings

**🔵 Suggestion 1 — the close-path invariant is now load-bearing but unenforced.**
Removing the `sessions.status = 'active'` filter from `get_pending_races` is correct *only because* every session-close path stamps `dropped_at` first. Verified: the two production close paths — `leave_session`'s `SessionClosed` arm and `close_stale_sessions` — both call `close_session_and_drop_pending` before flipping status (`submission.rs:670` is test-only). But the two paths set `status = Closed` independently; nothing structurally forces a future third close path to drop pending first, and a miss resurfaces phantom pending races on a closed session. A single `close_session(txn, id)` choke-point doing drop→notify→status-flip would make the invariant structural rather than convention. `test_pending_excludes_closed_session` guards it today.

**🔵 Suggestion 2 — duplicated raw-SQL predicate in `close_session_and_drop_pending`.** See inline comment.

**🔵 Suggestion 3 — `close_stale_sessions` loops count+UPDATE+inserts per stale session.**
`seaorm.md` § 2 cautions against per-row queries. Here N = stale sessions per 15-min sweep (tiny), and the per-user notification INSERT is inherently per-(session,user), so this isn't a real violation — but the count query and drop UPDATE could be set-based across all stale ids. Fine at current scale; noting for the record.

**🔵 Suggestion 4 — `lifecycle.rs` non-test code is ~772 lines, over `rust.md` § 13's ~500 soft limit.**
Pre-existing (the file was already over before this PR), and `close_session_and_drop_pending` genuinely belongs with lifecycle. But the stale-sweeper + drop-on-close cluster is now a cohesive, sizable concern — a future `services/sessions/sweeper.rs` extraction is the natural split.

### Design-doc accuracy
Verified accurate. `data-model.md` (new `dropped_at` column, four-state status enum, `notifications` table, rewritten liveness/pending sections), `api-contract.md` (§ 1.8 endpoints, § 4 ETag 7→6 inputs), and `user-workflows.md` § 1.5 are all updated and match the code. The ETag § 4 change is doc-only — no ETag code exists yet, so nothing to sync. `design.md` itself carries no pending-race/notification content, so no update is needed there.

### ADR-0032 deviation (flat `LIMIT 100` vs cursor pagination)
Well-justified and surfaced in `NOTIFICATIONS_PAGE_LIMIT`'s doc comment and `api-contract.md` § 1.8. Minor: the reconcile follow-up currently lives only in `.agents/handoffs/cowork.md` (gitignored) — a GitHub Issue would be more durable than a handoff that's deleted on read.

### What's good
Notification atomicity (every INSERT on the close `txn`) implements ADR-0038 § Atomicity exactly. The serde-tagged closed enum with `kind_str()` kept in lockstep is clean and extensible. Defensive `u32::try_from` on the `COUNT(*)` i64. Excellent test coverage — the per-signal sweeper-liveness tests isolate each of the five activity signals; the four-state status enum table in `data-model.md` is genuinely clarifying. Hand-written entity + `_behavior.rs` split follows the established convention.

brendanbyrne:
### Code review

brendanbyrne:
Thanks for the thorough review. Addressed the two code suggestions and filed the ADR-0032 follow-up:

**Suggestion 1 — close-path invariant now structural.** `9eb6da8`. Merged the drop+notify helper and the status flip into a single `close_session(txn, id)` choke-point. Both production close paths — `leave_session`'s last-leave branch and `close_stale_sessions` — route through it, so "every unresolved pending row is stamped `dropped_at` before status flips" is enforced by structure, not convention. A future close path can't flip status without dropping first. This also removed the separate batch status `UPDATE` in the sweeper (now per-session through the choke-point) — N is tiny per sweep, and per Suggestion 3 the loop was already per-(session,user) for notifications anyway.

**Suggestion 2 — shared predicate.** `9eb6da8`. The `skipped_at IS NULL AND dropped_at IS NULL AND NOT EXISTS(runs…)` body is now a single `UNRESOLVED_PENDING_PREDICATE` const that both the count query and the drop `UPDATE` interpolate — drift is impossible. Written with the table named in full (no alias) so it's valid in both the aliased SELECT and the un-aliasable UPDATE target.

**Suggestion 3 — per-stale-session loop.** Left as-is, as you flagged ("fine at current scale; noting for the record"). The Suggestion 1 choke-point makes the sweeper per-session by design.

**Suggestion 4 — `lifecycle.rs` size.** Left for a future `services/sessions/sweeper.rs` extraction — pre-existing and out of scope here.

**ADR-0032 deviation.** Cowork weighed in (Option 1): keep ADR-0032 `accepted`, track project-wide keyset rollout as a follow-up — filed as #166. ADR-0032 gained an `## Implementation status` section (`f95f2d9`) recording that no list endpoint uses keyset pagination yet and pointing at #166. PR description's Linked work updated.

All 297 lib tests + integration suites still pass; clippy + fmt clean. No behavior change from the refactor.

backend/src/services/sessions/races.rs
```diff
@@ -1000,21 +1011,24 @@ mod tests {
...
-    /// Documents intent: no background timer or sweeper task touches
-    /// `session_race_participations`. Pending accessibility is a pure read-time
-    /// derivation from `(session_race_participations, session_races, runs)`
-    /// plus the per-race expiry clock — see `get_pending_races`. Resolved /
-    /// expired rows remain in the DB indefinitely as historical state.
+    /// Documents intent: pending accessibility is a read-time derivation from
+    /// `(session_race_participations, session_races, runs)` — see
+    /// `get_pending_races`. No periodic timer ages a row out on its own
+    /// clock. The one writer of `dropped_at` is the session-close transaction
+    /// (`close_session_and_drop_pending`, ADR-0037), which runs inline on the
```
brendanbyrne: 🟡 **Important** — stale function name. This comment names `close_session_and_drop_pending`, but that function no longer exists: the Suggestion-1 fix merged the helper into the `close_session` choke-point and renamed it. Per `rust.md` § 6, a doc naming a non-existent symbol actively misleads. Fix: `close_session`.

backend/src/services/notifications.rs
```diff
@@ -0,0 +1,442 @@
...
+
+/// Record a [`NotificationPayload::PendingRacesDropped`] for one affected
+/// user. Called once per user inside the session-close transaction (ADR-0037).
+///
+/// # Errors
+///
+/// Propagates the errors of [`record`].
+#[tracing::instrument(
+    skip(txn),
+    fields(user_id = %user_id, session_id = %session_id, dropped_count),
```
brendanbyrne: 🔵 **Suggestion** — `dropped_count` is listed bare in `fields(...)`. A bare name in `#[tracing::instrument]` `fields` declares an *empty* span field meant to be filled later by `Span::record`; nothing records it, so the value never reaches traces. Naming it here also suppresses the automatic arg capture that would otherwise record it. Fix: drop it from `fields()` (it is auto-captured as an unskipped argument) or write `dropped_count = dropped_count`.

backend/src/routes/notifications.rs
```diff
@@ -0,0 +1,76 @@
...
+};
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    AppState, error::Error, extract::Json, middleware::auth::User, services::notifications,
+};
+
+/// Query string for `GET /me/notifications`.
+#[derive(Deserialize)]
+pub struct ListNotificationsQuery {
```
brendanbyrne, line 14: 🔵 **Suggestion** — `ListNotificationsQuery` has no `#[serde(deny_unknown_fields)]`, so a typo'd query param (`?include_reed=true`) is silently dropped rather than rejected. `rust.md` § 14's `deny_unknown_fields` rule is written for body DTOs, so applying it to a query struct is a judgment call — raising it as opinion, not a rule violation.

brendanbyrne (commented):
## Re-review — PR #165 (current HEAD)

Follow-up to the prior review. The author's fixes are verified resolved; this review reports one new finding the rename in those fixes introduced, plus two minor nits.

**Verdict: Approve** — none of the below blocks merge.

### Automated checks (current HEAD)
- `cargo +nightly fmt --check` — clean
- `cargo clippy --all-targets --all-features -D warnings` — clean
- `cargo test` — 297 lib + 80 integration/doc tests, all pass

### Prior findings — verified
- 🔵 1 (close-path invariant) — **resolved**: `close_session` is now the single choke-point; both `leave_session` and `close_stale_sessions` route through it.
- 🔵 2 (duplicated raw-SQL predicate) — **resolved**: lifted to the `UNRESOLVED_PENDING_PREDICATE` const, shared by the count query and the drop UPDATE.
- 🔵 3 (per-stale-session loop) / 🔵 4 (lifecycle.rs length) — unchanged; both were accepted as-is in the prior review. lifecycle.rs is now ~783 non-test lines — a future `services/sessions/sweeper.rs` extraction remains the natural split (`rust.md` § 13).
- ADR-0032 deviation — **resolved**: filed as #166; ADR-0032 gained an `## Implementation status` section.

### New finding
See inline 🟡 on `races.rs` — a doc comment names `close_session_and_drop_pending`, which no longer exists after the Suggestion-1 rename to `close_session`. The prior review couldn't catch this because the function still had the old name at that time.

### Standards-gap note (not a PR finding)
`NotificationView` / `UnreadCountResponse` omit `#[serde(rename_all = "snake_case")]`, which `rust.md` § 14 states for "every struct that crosses the wire." But the whole signed-off codebase (`auth.rs::Response`, `MySessionResponse`, …) omits it on response DTOs whose fields are already snake_case. The new code matches established practice — worth a § 14 wording clarification rather than a finding here.

brendanbyrne:
Re-review addressed — `3ec45bb`.

- **🟡 races.rs stale symbol** — fixed. `test_lazy_check_assertion`'s doc comment named `close_session_and_drop_pending`; the Suggestion-1 rename to `close_session` missed this one reference in `races.rs` (I only swept `lifecycle.rs`). Now `close_session`.

- **🔵 notifications.rs bare tracing field** — fixed. Dropped `dropped_count` from `record_pending_drops`'s `#[instrument(fields(...))]`. You're right that the bare name declared an empty span field and suppressed auto-capture. It's now auto-captured as an unskipped `Copy` arg, with a comment noting why it's deliberately absent from `fields()`.

- **🔵 `deny_unknown_fields` on `ListNotificationsQuery`** — left off, deliberately. The sibling query struct `ListRunsQuery` (`routes/runs.rs`) is also a plain `#[derive(Deserialize)]` with no `deny_unknown_fields`, so the codebase's established pattern for *query* structs is to omit it — and `rust.md` § 14 scopes the rule to body DTOs. Adding it to this one struct alone would be an inconsistency. If fail-loud-on-typo'd-query-param is wanted, it's a project-wide call (every query struct + a § 14 wording update) rather than a one-off here — happy to file that as a separate Issue if you'd like.

- **`rename_all` standards-gap note** — no action, as you flagged it ("not a PR finding"; the new DTOs match established response-DTO practice).

297 lib tests pass; clippy + fmt clean.

brendanbyrne:
Filed the `deny_unknown_fields` query-struct decision as #167 (Backlog) — framed as a decide-then-act ticket: either adopt it across all query structs and update `rust.md` § 14, or add a § 14 note scoping the rule to body DTOs so the omission reads as intentional. Out of scope for this PR.

brendanbyrne:
Filed the `rename_all` standards-gap as #168 (Backlog) — decide-then-act: either reword `rust.md` § 14 so the established "omit when fields are already snake_case" practice is compliant, or keep the rule and sweep the missing attributes in. Out of scope for this PR.

Issue #176, opened 2026-05-18
Set up TanStack Query and migrate static-data hooks


brendanbyrne:
## Context

Fifth PR. Wraps the app in `QueryClientProvider` and migrates the static-data hooks (`useGameData`, `useUserProfile`) to `useQuery`. Drop-in for hooks whose external contract stays the same — keep the legacy names exported as thin wrappers so call sites don't churn.

## Acceptance criteria

- [ ] Wrap `App.tsx` in `QueryClientProvider`. Default config: `refetchOnWindowFocus: true`, `retry: 1`, `staleTime: 30_000`.
- [ ] Add React Query Devtools in dev only.
- [ ] Migrate `useGameData.ts` (the five `useSimpleList` hooks + `useDrinkTypes`) to `useQuery`. Long `staleTime` for static data, no polling.
- [ ] `useDrinkTypes`'s `refresh()` becomes `queryClient.invalidateQueries({ queryKey: ['drink-types'] })`.
- [ ] Migrate `useUserProfile.ts` to `useQuery`; `refresh()` becomes invalidation.
- [ ] Every consumer (DrinkTypeSelector, Profile, Home, Session) still works.
- [ ] Devtools panel shows the queries.

## References

- Compliance plan: PR-C1
- Standards: `docs/coding-standards/react.md` § 4, § 6
- Polling model: `docs/decisions/0018-realtime-via-polling-not-websockets.md`

_Filed by Cowork on Brendan's behalf._

PR #199, opened 2026-05-19
191: introduce Zod schemas + Result type at the API boundary

### LINKED ISSUE #191: Add Zod runtime validation and Result type at the API boundary
## Context

Fourth PR. Replaces `await res.json() as Foo` (an unchecked annotation) with Zod-parsed responses across all 9 affected files. Introduces a `Result<T, ApiError>` shape for expected user-facing failures (4xx) while letting unexpected failures throw and bubble to an error boundary. Also threads `AbortSignal` through every API helper so cancellation works under TanStack Query (PR-C1).

## Acceptance criteria

- [ ] Add Zod schemas for every API response shape in `frontend/src/api/types.ts`. Infer TS types from the schemas (`z.infer<typeof ...>`) — delete the hand-written types.
- [ ] Brand-mint inside the schema via `.transform((s) => s as UserId)` etc., replacing the at-the-call-site brand casts from PR-B1.
- [ ] Replace `await res.json() as Foo` with `Schema.parse(...)` in `api/sessions.ts`, `api/runs.ts`, `api/client.ts`, and the direct `apiFetch` calls in `useAuth.tsx`, `useGameData.ts`, `useUserProfile.ts`, `DrinkTypeSelector.tsx`, `Onboarding.tsx`, `Profile.tsx`.
- [ ] Introduce `Result<T, E>` in `frontend/src/api/result.ts`.
- [ ] Define `ApiError` as a discriminated union on `code` values from `docs/api-contract.md` § 8.
- [ ] Update `getRunDefaults` to return `Result` instead of the silent-fallback hardcoded value.
- [ ] Add `AbortSignal` parameter to every API helper; thread into `fetch`.
- [ ] Happy-path flows work; a forced backend response mismatch produces a clear Zod error.

## References

- Compliance plan: PR-B2
- Audit findings: all 9 `any`-typed `res.json()` sites
- Standards: `docs/coding-standards/typescript.md` § 6, § 7, § 8
- Wire format: `docs/api-contract.md` § 8 (error envelope)

brendanbyrne:
## Reviewer notes

Large diff (18 files / +945 −334) but mostly mechanical: every response DTO became a Zod schema, every helper threads `AbortSignal`, every untyped `await res.json()` became `parseBody(schema, res)`. Two design decisions worth flagging up front:

1. **AC-literal `Result` scope.** Only `getRunDefaults` returns `Result<T, ApiError>` per the compliance plan's PR-B2 scope bullet. Other helpers still `throw` (now an `ApiErrorException` carrying a typed `ApiError`) — PR-C1/C2 migrates them into TanStack Query, which catches throws idiomatically. Going broader now would be churn the C-stream immediately reworks. Confirmed with Brendan before starting.
2. **`ApiError` modeled as `{ code: ApiErrorCode; message: string }`** rather than the literal expanded discriminated union from typescript.md § 6's example. Every variant today would be the identical `{ code; message }`, and an expanded union can't be *constructed* from a dynamically-parsed code without an unchecked cast. `switch (e.code)` still narrows exhaustively over the closed `ApiErrorCode` set. When a code grows code-specific fields, splitting into a true per-code union at that point is one local change. Documented inline in [src/api/result.ts](frontend/src/api/result.ts#L57-L72).

Two smaller things:

- `apiFetch` gained an `ApiFetchOptions` type that widens `signal` to `AbortSignal | undefined` so helpers can build `{ signal }` from an optional param under `exactOptionalPropertyTypes`. The extra `undefined` is stripped before the real `fetch` call.
- Import style is `import * as z from 'zod'`, not `import { z } from 'zod'`. Zod 4's root re-exports `z` via `import * as / export { z }`, which Bun's ESM doesn't honor — `z` ends up undefined at runtime even though typecheck resolves. Namespace import is Zod 4's documented canonical form.

## Summary

PR-B2 of the frontend compliance plan. Adds Zod-parsed boundaries on every API response, a `Result<T, ApiError>` value type for expected 4xx failures, and `AbortSignal` plumbing on every helper. Brands are now minted inside the schema (`.transform(UserId)`), replacing PR-B1's call-site `as` casts.

## Linked work

- Closes #191
- Per [docs/designs/2026-05-16-frontend-compliance-plan.md](docs/designs/2026-05-16-frontend-compliance-plan.md) PR-B2
- Standards: [docs/coding-standards/typescript.md](docs/coding-standards/typescript.md) §§ 6, 7, 8
- Wire format: [docs/api-contract.md](docs/api-contract.md) §§ 3, 8

## How to verify

```bash
# Repo state
cd ~/path/to/beerio-kart   # /mnt/c/Users/obiva/beerio-kart on the WSL2 setup
git fetch origin && git checkout 191/zod-runtime-validation && cd frontend
bun install
```

- [x] **Typecheck, lint, tests pass cleanly.**
  ```bash
  bun run typecheck             # tsc -b — zero errors
  bun run lint 2>&1 | tail -3   # 0 errors, only the warn-down rules
  bun run test 2>&1 | tail -10  # 147 passed | 2 todo
  ```
  Expect `Tests  147 passed | 2 todo` and `0 errors` from lint. New tests live in [frontend/src/api/result.test.ts](frontend/src/api/result.test.ts), with `types.test.ts` / `brand.test.ts` filling in the new schemas and [client.test.ts](frontend/src/api/client.test.ts) / [DrinkTypeSelector.test.tsx](frontend/src/components/DrinkTypeSelector.test.tsx) / [Onboarding.test.tsx](frontend/src/pages/Onboarding.test.tsx) added for the patch-coverage gate.

- [x] **Branded IDs survive parse and round-trip JSON.** Quick repl check — the brand is erased at runtime, so a parsed value is byte-identical to the wire JSON:
  ```bash
  # `bun -e`, not `bun run -e` — `-e`/eval is a runtime flag, and `bun run` is the script runner (it'll print usage if you pass it `-e`). Run from frontend/ so the relative import resolves.
  bun -e "import('./src/api/types.ts').then(({SessionDetailSchema}) => { const p = SessionDetailSchema.parse({id:'s1',host_id:'u1',host_username:'a',ruleset:'random',status:'active',created_at:'2026-05-19T00:00:00.000Z',participants:[],race_number:0,current_race:null,races:[]}); console.log('id:', p.id, 'roundtrip:', JSON.parse(JSON.stringify(p)).id); })"
  ```
  Expect `id: s1 roundtrip: s1`.

- [x] **A backend response shape mismatch fails loud, not silently.** This proves the PR's core claim: `parseBody` rejects a wrong-shaped body at the boundary instead of letting `undefined` leak downstream.

  1. Start the dev stack — backend: `cd backend && cargo run`; frontend (separate terminal): `cd frontend && bun run dev`. Open the printed localhost URL and log in.
  2. Force a wire-shape mismatch with one surgical backend edit. In [backend/src/services/sessions/detail.rs](backend/src/services/sessions/detail.rs) (the `SessionDetail` struct, ~line 114), add a serde rename **above** the `status` field so only the JSON key changes — no other Rust code needs touching (the struct only derives `Serialize`, and all Rust code still references the `status` field name):
     ```rust
         /// Lifecycle state — `Active` or `Closed`.
         #[serde(rename = "state")]   // ← TEMP: makes the wire shape mismatch SessionDetailSchema
         pub status: SessionStatus,
     ```
     (Renaming the Rust field itself would force edits at every construction site and fail to compile — the serde attribute is the one-line way.)
  3. Restart the backend (`Ctrl-C`, `cargo run` again) so the change takes effect.
  4. In the browser, open DevTools → Console, then on the Home screen click **Create session**.
  5. **Expected (after this PR):** the create flow surfaces a red error banner reading **`Response failed schema validation: …`** with the Zod detail naming the missing `status` field. The throw is a typed `ApiErrorException` (`apiError.code: 'response_shape_mismatch'`, `cause` = the `ZodError`), caught by [Home.tsx](frontend/src/pages/Home.tsx)'s create handler. *Before this PR* the body would have parsed to an object with `status: undefined` and broken somewhere unrelated downstream.
  6. *(Optional — inspect the structured error directly.)* Paste into the DevTools Console:
     ```js
     const { createSession } = await import('/src/api/sessions.ts');
     try { await createSession('random'); }
     catch (e) { console.log(e.name, e.apiError, '\ncause:', e.cause); }
     ```
     Expect `ApiErrorException { code: 'response_shape_mismatch', message: … }` and the `ZodError` as `cause`. (Works because Vite serves the same module singleton your logged-in tab already holds, so the auth token is shared.)
  7. **Revert the backend edit** (remove the `#[serde(rename = "state")]` line) and restart the backend.

  > Note: opening an *existing* session page also triggers the throw, but [useSession.ts](frontend/src/hooks/useSession.ts) polls every 2.5 s with no try/catch, so there it shows up as a **recurring unhandled promise rejection** in the console rather than a clean banner — the create-session path above is the cleaner single-shot check.

- [x] **Happy-path flows still work.** With the stack running:
  - [x] Log in with an existing account → land on Home, no errors.
  - [x] Create a new session (host) → see SessionDetail render with branded IDs (look unchanged in the UI; that's the point).
  - [x] Pick the next track → race info renders.
  - [x] Submit a run with a drink type and full race setup → run appears under the race, no console errors.
  - [ ] Open Profile, change preferred character, save → reload Profile, change persisted.
  - [ ] Add a new drink type via the DrinkTypeSelector "+ Add new" flow → new entry appears.

- [ ] **Failed-defaults path leaves the form blank.** Temporarily make `GET /api/v1/runs/defaults` return 500 (e.g. add a forced `return Err(...)` in the handler). Open the run-entry sheet on a race — every field should be blank, no error toast, same end state as the old `source: 'none'` hardcoded fallback. Revert after.

- [ ] **4xx surfaces the backend message.** Try logging in with a wrong password → "Invalid credentials." Try creating an already-existing username → "Username already taken." (Both flow through `parseApiError` now.)

## Author checklist

- [x] Linked to an Issue under "Linked work" above (#191).
- [x] Tests added or updated for new logic (new `result.test.ts`, schema tests in `types.test.ts` / `brand.test.ts`, updated `sessions.test.ts` / `runs.test.ts`).
- [x] Docs updated if applicable — n/a; the compliance plan's PR-B2 sign-off box is Brendan's call to flip post-merge.
- [x] Schema changes: n/a (frontend-only).
- [x] Tested locally end-to-end — `bun run typecheck`, `bun run lint`, `bun run test` all clean; manual stack verification recommended via "How to verify" above.

### DIFF
--- a/frontend/src/api/client.ts
+++ b/frontend/src/api/client.ts
@@ -11,6 +11,9 @@
  * is truly expired and the user needs to log in again.
  */
 
+import { parseBody } from './result';
+import { TokenRefreshSchema } from './types';
+
 let accessToken: string | null = null;
 
 export function setAccessToken(token: string | null) {
@@ -41,7 +44,7 @@ async function tryRefresh(): Promise<boolean> {
       // which means the browser automatically sends the HttpOnly cookie.
     });
     if (!res.ok) return false;
-    const data = await res.json();
+    const data = await parseBody(TokenRefreshSchema, res);
     accessToken = data.access_token;
     return true;
   } catch {
@@ -50,26 +53,46 @@ async function tryRefresh(): Promise<boolean> {
 }
 
 /**
+ * Options accepted by `apiFetch`. Identical to `RequestInit` except `signal`
+ * is widened to explicitly include `undefined`: `exactOptionalPropertyTypes`
+ * otherwise rejects a caller that builds `{ signal }` from an optional
+ * `AbortSignal` parameter — the API-helper pattern in sessions.ts / runs.ts.
+ */
+export type ApiFetchOptions = Omit<RequestInit, 'signal'> & {
+  signal?: AbortSignal | undefined;
+};
+
+/**
  * Wrapper around fetch() that adds the access token and handles 401 refresh.
  */
 export async function apiFetch(
   url: string,
-  options: RequestInit = {},
+  options: ApiFetchOptions = {},
 ): Promise<Response> {
-  const headers = new Headers(options.headers);
+  const { signal, ...rest } = options;
+  const headers = new Headers(rest.headers);
   if (accessToken) {
     headers.set('Authorization', `Bearer ${accessToken}`);
   }
 
-  let res = await fetch(url, { ...options, headers });
+  // Spread `signal` in only when present — `exactOptionalPropertyTypes`
+  // forbids handing `fetch` an explicit `signal: undefined`.
+  const init: RequestInit = {
+    ...rest,
+    headers,
+    ...(signal ? { signal } : {}),
+  };
+
+  let res = await fetch(url, init);
 
   // If we get a 401 and we had a token, try refreshing
   if (res.status === 401 && accessToken) {
     const refreshed = await tryRefresh();
     if (refreshed) {
-      // Retry the original request with the new token
+      // Retry with the new token. The shared `headers` object is mutated in
+      // place, so `init` already points at the updated Authorization header.
       headers.set('Authorization', `Bearer ${accessToken}`);
-      res = await fetch(url, { ...options, headers });
+      res = await fetch(url, init);
     } else {
       // Refresh failed — session is truly expired
       accessToken = null;
--- a/frontend/src/api/client.ts
+++ b/frontend/src/api/client.ts
@@ -11,7 +11,7 @@
  * is truly expired and the user needs to log in again.
  */
 
-import { parseBody } from './result';
+import { logIfResponseShapeMismatch, parseBody } from './result';
 import { TokenRefreshSchema } from './types';
 
 let accessToken: string | null = null;
@@ -47,7 +47,11 @@ async function tryRefresh(): Promise<boolean> {
     const data = await parseBody(TokenRefreshSchema, res);
     accessToken = data.access_token;
     return true;
-  } catch {
+  } catch (e) {
+    // A malformed refresh response (valid 2xx, missing access_token) would
+    // otherwise look identical to "no refresh cookie" and silently log the
+    // user out, masking a backend bug — surface it (§ 8).
+    logIfResponseShapeMismatch(e, '/api/v1/auth/refresh');
     return false;
   }
 }
--- a/frontend/src/api/result.ts
+++ b/frontend/src/api/result.ts
@@ -89,7 +89,7 @@ function isApiErrorCode(code: string): code is ApiErrorCode {
 }
 
 /**
- * The wire error envelope. `code` is optional: § 3 says the backend always
+ * The wire error envelope. `code` is optional: § 2 says the backend always
  * sends it, but a malformed or proxy-generated error response might not, and
  * a missing code degrades to `unknown` rather than throwing.
  */
--- a/frontend/src/api/result.ts
+++ b/frontend/src/api/result.ts
@@ -160,3 +160,24 @@ export async function parseBody<S extends z.ZodType>(
   }
   return parsed.data;
 }
+
+/**
+ * Log a swallowed error if — and only if — it's a `response_shape_mismatch`.
+ *
+ * The legacy `useEffect`-fetch hooks intentionally degrade to an empty result
+ * on a network failure, but § 8 says contract drift must stay visible. This
+ * lets those catch blocks keep their network-error behavior while surfacing a
+ * `parseBody` schema failure to the console. PR-C1 retires these call sites by
+ * moving the hooks to TanStack Query, which propagates errors to a boundary.
+ */
+export function logIfResponseShapeMismatch(
+  error: unknown,
+  context: string,
+): void {
+  if (
+    error instanceof ApiErrorException &&
+    error.apiError.code === 'response_shape_mismatch'
+  ) {
+    console.error(`Response shape mismatch: ${context}`, error);
+  }
+}
--- a/frontend/src/api/result.ts
+++ b/frontend/src/api/result.ts
@@ -1,0 +1,162 @@
+/**
+ * `Result`, the typed `ApiError`, and the runtime-validation helpers that
+ * live at the `fetch` boundary.
+ *
+ * Error-handling split (typescript.md § 6): *expected* domain failures — a
+ * 4xx the caller may want to branch on — are returned as `Result<T, ApiError>`
+ * values; *unexpected* failures (network down, 5xx, a 2xx body that fails its
+ * Zod schema) throw and bubble to an error boundary. `ApiErrorException` is
+ * the thrown carrier — it subclasses `Error` (so existing `catch (e)` /
+ * `e.message` sites keep working) while also exposing the structured
+ * `apiError` for catch sites that want to branch on `.code`.
+ *
+ * PR-B2 (Issue #191).
+ */
+import * as z from 'zod';
+
+/**
+ * Expected-failure-as-value. A helper returns `Result` only when its caller
+ * is expected to handle the failure inline; everything else throws.
+ */
+export type Result<T, E> = { ok: true; value: T } | { ok: false; error: E };
+
+/**
+ * Stable error `code` values. The first block mirrors the wire registry in
+ * api-contract.md § 8 one-for-one — renaming a code there is a breaking wire
+ * change. The last two are client-synthesized:
+ *  - `response_shape_mismatch` — a response parsed as JSON but failed its Zod
+ *    schema (§ 8: contract drift, treated as a programmer error).
+ *  - `unknown` — the backend sent a `code` we don't recognize, or none at all.
+ */
+export const API_ERROR_CODES = [
+  'bad_request',
+  'lap_times_mismatch',
+  'track_id_mismatch',
+  'invalid_path_param',
+  'invalid_request_body',
+  'invalid_credentials',
+  'token_expired',
+  'token_invalid',
+  'forbidden',
+  'admin_required',
+  'not_found',
+  'username_taken',
+  'session_closed',
+  'pending_races_first',
+  'out_of_order_submission',
+  'race_number_conflict',
+  'unprocessable',
+  'internal',
+  'gateway_timeout',
+  'response_shape_mismatch',
+  'unknown',
+] as const;
+
+export type ApiErrorCode = (typeof API_ERROR_CODES)[number];
+
+/**
+ * The frontend's view of the backend error envelope (`{ error, code }`,
+ * api-contract.md § 3). It discriminates on `code`: a `switch (err.code)`
+ * narrows exhaustively against the closed `ApiErrorCode` set.
+ *
+ * Modeled as one shape rather than an expanded `{ code: 'a' } | { code: 'b' }`
+ * union because no code carries a code-specific payload today — every member
+ * would be the identical `{ code; message }`. An expanded union also can't be
+ * *constructed* from a dynamically-parsed `code` without an unchecked cast.
+ * If a future code grows extra fields, split `ApiError` into a real per-code
+ * union at that point.
+ */
+export type ApiError = {
+  code: ApiErrorCode;
+  message: string;
+};
+
+/** Thrown carrier for an `ApiError` — see the file header. */
+export class ApiErrorException extends Error {
+  readonly apiError: ApiError;
+
+  constructor(apiError: ApiError, options?: ErrorOptions) {
+    super(apiError.message, options);
+    this.name = 'ApiErrorException';
+    this.apiError = apiError;
+  }
+}
+
+function isApiErrorCode(code: string): code is ApiErrorCode {
+  // The `as readonly string[]` only widens the literal tuple so `.includes`
+  // accepts an arbitrary string — it is not an unsafe narrowing cast.
+  return (API_ERROR_CODES as readonly string[]).includes(code);
+}
+
+/**
+ * The wire error envelope. `code` is optional: § 3 says the backend always
+ * sends it, but a malformed or proxy-generated error response might not, and
+ * a missing code degrades to `unknown` rather than throwing.
+ */
+const errorEnvelopeSchema = z.object({
+  error: z.string(),
+  code: z.string().optional(),
+});
+
+function httpStatusError(status: number): ApiError {
+  return {
+    code: 'unknown',
+    message: `Request failed (HTTP ${String(status)})`,
+  };
+}
+
+/**
+ * Read a non-2xx `Response` into a typed `ApiError`. Never throws: a body
+ * that isn't the expected envelope (empty, non-JSON, wrong shape) degrades to
+ * an `unknown`-coded error carrying an HTTP-status message.
+ */
+export async function parseApiError(res: Response): Promise<ApiError> {
+  let body: unknown;
+  try {
+    body = await res.json();
+  } catch {
+    return httpStatusError(res.status);
+  }
+  const parsed = errorEnvelopeSchema.safeParse(body);
+  if (!parsed.success) return httpStatusError(res.status);
+  const { error, code } = parsed.data;
+  return {
+    code: code !== undefined && isApiErrorCode(code) ? code : 'unknown',
+    message: error,
+  };
+}
+
+/**
+ * Parse a 2xx `Response` body through a Zod schema. On a JSON or schema
+ * failure throws `ApiErrorException` with `code: 'response_shape_mismatch'`
+ * (§ 8: contract drift fails loud) — the underlying error is preserved as
+ * `cause` so the original stack stays reachable.
+ */
+export async function parseBody<S extends z.ZodType>(
+  schema: S,
+  res: Response,
+): Promise<z.output<S>> {
+  let body: unknown;
+  try {
+    body = await res.json();
+  } catch (cause) {
+    throw new ApiErrorException(
+      {
+        code: 'response_shape_mismatch',
+        message: 'Response body was not valid JSON',
+      },
+      { cause },
+    );
+  }
+  const parsed = schema.safeParse(body);
+  if (!parsed.success) {
+    throw new ApiErrorException(
+      {
+        code: 'response_shape_mismatch',
+        message: `Response failed schema validation: ${parsed.error.message}`,
+      },
+      { cause: parsed.error },
+    );
+  }
+  return parsed.data;
+}
--- a/frontend/src/api/result.ts
+++ b/frontend/src/api/result.ts
@@ -22,7 +22,7 @@ export type Result<T, E> = { ok: true; value: T } | { ok: false; error: E };
 
 /**
  * Stable error `code` values. The first block mirrors the wire registry in
- * api-contract.md § 8 one-for-one — renaming a code there is a breaking wire
+ * api-contract.md § 7 one-for-one — renaming a code there is a breaking wire
  * change. The last two are client-synthesized:
  *  - `response_shape_mismatch` — a response parsed as JSON but failed its Zod
  *    schema (§ 8: contract drift, treated as a programmer error).
@@ -56,7 +56,7 @@ export type ApiErrorCode = (typeof API_ERROR_CODES)[number];
 
 /**
  * The frontend's view of the backend error envelope (`{ error, code }`,
- * api-contract.md § 3). It discriminates on `code`: a `switch (err.code)`
+ * api-contract.md § 2). It discriminates on `code`: a `switch (err.code)`
  * narrows exhaustively against the closed `ApiErrorCode` set.
  *
  * Modeled as one shape rather than an expanded `{ code: 'a' } | { code: 'b' }`
--- a/frontend/src/api/sessions.ts
+++ b/frontend/src/api/sessions.ts
@@ -1,5 +1,14 @@
+import * as z from 'zod';
 import { apiFetch } from './client';
-import { SessionId } from './brand';
+import { SessionIdSchema } from './brand';
+import { ApiErrorException, parseApiError, parseBody } from './result';
+import {
+  RaceInfoSchema,
+  SessionDetailSchema,
+  SessionRaceInfoSchema,
+  SessionSummarySchema,
+} from './types';
+import type { SessionId } from './brand';
 import type {
   RaceInfo,
   SessionDetail,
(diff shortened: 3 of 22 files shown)


frontend/src/hooks/useGameData.ts
```diff
@@ -11,7 +14,7 @@ function useSimpleList(endpoint: string) {
     async function load() {
       try {
         const res = await apiFetch(endpoint);
-        const data = await res.json();
+        const data = await parseBody(z.array(SimpleItemSchema), res);
         setItems(data);
       } catch {
```
brendanbyrne: 🟡 **Important** — Per `typescript.md` § 8, a `response_shape_mismatch` is contract drift and must fail loud, not be silently coerced. With `parseBody` now inside the `try`, this `catch` swallows shape-mismatch errors and leaves `items` empty — indistinguishable from a real network failure.

Fix (minimum, until PR-C1 migrates to TanStack Query):

```ts
} catch (e) {
  if (e instanceof ApiErrorException && e.apiError.code === 'response_shape_mismatch') {
    console.error('Response shape mismatch on', endpoint, e);
  }
  // Items stay empty for network errors.
} finally {
  setLoading(false);
}
```

frontend/src/hooks/useGameData.ts
```diff
@@ -54,7 +57,7 @@ export function useDrinkTypes() {
     async function load() {
       try {
         const res = await apiFetch('/api/v1/drink-types');
-        const data = await res.json();
+        const data = await parseBody(z.array(DrinkTypeSchema), res);
         setItems(data);
       } catch {
```
brendanbyrne: 🟡 **Important** — Same as line 19: `parseBody` is inside the `try`, so a `response_shape_mismatch` here is silently coerced into "empty drink-type list," violating `typescript.md` § 8. Mirror the fix from the other `useSimpleList` catch.

frontend/src/hooks/useUserProfile.ts
```diff
@@ -17,7 +19,7 @@ export function useUserProfile(userId: string | undefined) {
     async function load() {
       try {
         const res = await apiFetch(`/api/v1/users/${userId}`);
-        const data = await res.json();
+        const data = await parseBody(UserDetailProfileSchema, res);
```
brendanbyrne, line 22: 🟡 **Important** — Same pattern as `useGameData.ts:19/62`: a `parseBody` schema failure here is swallowed as a silent "profile is null," indistinguishable from a 404 or a network drop. Per `typescript.md` § 8 a shape mismatch is contract drift and must surface — at minimum, log when the caught error is `ApiErrorException` with `code === 'response_shape_mismatch'`.

frontend/src/api/client.ts
```diff
@@ -41,35 +44,55 @@ async function tryRefresh(): Promise<boolean> {
       // which means the browser automatically sends the HttpOnly cookie.
     });
     if (!res.ok) return false;
-    const data = await res.json();
+    const data = await parseBody(TokenRefreshSchema, res);
     accessToken = data.access_token;
     return true;
   } catch {
```
brendanbyrne: 🟡 **Important** — Subtle one: with `parseBody` now inside the `try`, a malformed `/auth/refresh` response (valid 2xx but missing `access_token`) is swallowed and returns `false`, making it indistinguishable from "no refresh cookie." The user gets logged out as if their cookie expired, masking a real backend bug. Same `typescript.md` § 8 issue as the hook catches. Worth at least logging when the swallowed error is a `response_shape_mismatch`.

frontend/src/hooks/useAuth.tsx
```diff
@@ -45,18 +47,18 @@ export function AuthProvider({ children }: { children: ReactNode }) {
...
-            .split('.')[1]
-            .replace(/-/g, '+')
-            .replace(/_/g, '/');
-          const payload = JSON.parse(atob(base64));
-          setUser({ id: payload.sub, username: payload.username });
+          const payloadSegment = data.access_token.split('.')[1];
+          if (payloadSegment) {
+            const base64 = payloadSegment.replace(/-/g, '+').replace(/_/g, '/');
+            const payload = JSON.parse(atob(base64));
+            setUser({ id: payload.sub, username: payload.username });
```
brendanbyrne, line 65: 🔵 **Suggestion** — The PR Zod-parsed the `TokenRefreshSchema` response, but `JSON.parse(atob(base64))` on the next line still produces `any` and `payload.sub` / `payload.username` are read without validation. The hunk above touched these lines, so closing this loop is in-scope.

Small schema does the job:

```ts
// In types.ts (or co-located here):
export const AccessTokenPayloadSchema = z.object({
  sub: z.string(),
  username: z.string(),
});

// Here:
const payload = AccessTokenPayloadSchema.parse(JSON.parse(atob(base64)));
setUser({ id: payload.sub, username: payload.username });
```

Keeps the JWT-decode path under the same "every untyped boundary goes through Zod" rule (`typescript.md` § 8) as the API responses.

frontend/src/api/sessions.ts
```diff
@@ -8,93 +17,112 @@ import type {
...
-// payload, so it uses the `SessionId` constructor directly.
+// Every response is parsed through its Zod schema (`parseBody`), which is also
+// where the branded IDs are minted — the PR-B1 `as` casts are gone. A non-2xx
+// from a helper that has nothing to fall back on throws `ApiErrorException`
+// carrying the typed error envelope; helpers with a documented empty result
+// (`null` / `[]`) keep returning it. Every helper accepts an optional
+// `AbortSignal` and threads it into `fetch` (typescript.md § 7).
+
+/** `GET /sessions/mine` returns just `{ session_id }` — a tiny local shape. */
+const mySessionSchema = z.object({
```
brendanbyrne: 🔵 **Suggestion** — `mySessionSchema` is the only camelCase schema name; every other in the codebase is `PascalCaseSchema`. Either rename in-place (`MySessionSchema`) or promote to a top-level export in `types.ts` (e.g. `MySessionResponseSchema`) for consistency.

brendanbyrne (commented):
## Code review — PR-B2 (Zod schemas + `Result` at the API boundary)

**Verdict:** Approve with minor changes 🟡

**Standards loaded:** [`typescript.md`](docs/coding-standards/typescript.md) §§ 6, 7, 8 · [`api-contract.md`](docs/api-contract.md) §§ 3, 8 · [`react.md`](docs/coding-standards/react.md) (data fetching).

**Automated checks (run locally on this branch):**
- `bun run typecheck` — clean ✅
- `bun run lint` — 0 errors, 163 warnings (all pre-existing warn-down rules per the [compliance plan](docs/designs/2026-05-16-frontend-compliance-plan.md) PR-A1; this PR introduces no new warnings)
- `bun run test` — 118 passed, 2 todo ✅

**Design-doc accuracy:** No `docs/design.md` updates required — the doc doesn't cover frontend API-helper internals. The compliance plan's PR-B2 sign-off box is deliberately left for Brendan to flip post-merge (called out in the PR description).

---

### 🟡 Important (inline)

1. **Shape-mismatch errors are silently swallowed in three places**, contrary to `typescript.md` § 8 ("Validation failures … treated as a programmer error — logged, error boundary, fail-loud — not silently coerced"). With `parseBody` now inside the `try`, an `ApiErrorException(response_shape_mismatch)` is being silenced — exactly the failure mode § 8 says to surface. Four inline comments cover the sites: `useGameData.ts` (×2), `useUserProfile.ts`, and `client.ts` `tryRefresh`. **PR-C1** will replace these hooks with TanStack Query (which propagates to error boundaries), so this is reasonable to defer there — but at minimum each catch should log when `e instanceof ApiErrorException && e.apiError.code === 'response_shape_mismatch'`.

### 🔵 Suggestions (inline + one for here)

2. **JWT payload decode in `useAuth.tsx`** still bypasses Zod (inline comment on line 60).
3. **`mySessionSchema` casing** is the lone camelCase schema (inline comment on `sessions.ts`).
4. **`typescript.md` § 6's `ApiError` example shows an expanded `{ code: 'a' } | { code: 'b' }` discriminated union; the PR implements a flat `{ code: ApiErrorCode; message }`.** The author's reasoning in [`result.ts:62-68`](frontend/src/api/result.ts#L62-L68) is correct — a dynamically-parsed code can't construct an expanded union without an unchecked cast, and no code carries a per-variant payload today. I'd accept this as-is and suggest a 2-line clarification in `typescript.md` § 6 as a separate docs task ("for codes without payload fields, a flat `{ code; message }` shape is equivalent; `switch (e.code)` narrows the same way"). Not blocking this PR.
5. **Response-schema test coverage is partial vs. `typescript.md` § 12.** Only `SessionDetailSchema`, `DrinkTypeSchema`, and `NotificationPayloadSchema` have per-schema tests. `RunDetailSchema`, `RunDefaultsSchema`, `AuthSessionSchema`, and `TokenRefreshSchema` — the response types the user-facing flows in this PR actually exercise — don't. The brand schemas use `it.each` compactly; the same pattern would let the response schemas earn coverage cheaply (happy path + one rejection + JSON round-trip each, per § 12).

### What's good

- File headers and inline rationale (`result.ts`, `types.ts`, `brand.ts`, the `apiFetch` `signal` comment) are unusually well-written — easy to follow what was done and why.
- `parseApiError` / `parseBody` test coverage is thorough: registered code, unknown code, missing code, non-JSON body, wrong shape body — all five degradation paths covered.
- `Error.cause` preserved through `ApiErrorException` per `typescript.md` § 6.
- `getRunDefaults` `Result` migration is a clean replacement of a silent hardcoded fallback with explicit caller branching.
- Brand minting fully relocated to the schema layer; the PR-B1 `as` casts at call sites are gone.
- `AbortSignal` threaded into every helper, satisfying `typescript.md` § 7. The `ApiFetchOptions` widening trick in `client.ts` to play nice with `exactOptionalPropertyTypes` is clean and well-commented.

brendanbyrne:
Thanks for the review — pushed `5c09a92` addressing the feedback.

**🟡 Important — shape-mismatch swallowed (useGameData ×2, useUserProfile, client.ts tryRefresh).** Fixed. Added `logIfResponseShapeMismatch(error, context)` to `result.ts` — it logs only when the caught error is an `ApiErrorException` with `code === 'response_shape_mismatch'`, so the four catch blocks keep degrading quietly on network errors but surface contract drift per § 8. Went with a shared helper rather than inlining the `instanceof` check four times. `tryRefresh` got an extra comment since its case (malformed refresh → silent logout masking a backend bug) is the subtlest. PR-C1 will retire these sites entirely when the hooks move to TanStack Query.

**🔵 JWT payload bypasses Zod.** Fixed. New `AccessTokenPayloadSchema` (`{ sub, username }`); the silent-refresh decode now does `AccessTokenPayloadSchema.parse(JSON.parse(atob(base64)))`. Side benefit: clears the two `no-unsafe-member-access` warnings that were on that line.

**🔵 `mySessionSchema` casing.** Fixed by promoting it to a top-level `MySessionResponseSchema` export in `types.ts` (your "promote" option), matching the `PascalCaseSchema` convention. `sessions.ts` no longer carries the local schema or the now-unused `SessionIdSchema` import.

**🔵 Partial response-schema test coverage.** Fixed. Added `it.each` coverage for `RunDetailSchema`, `RunDefaultsSchema`, `AuthSessionSchema`, `TokenRefreshSchema`, and the new `AccessTokenPayloadSchema` — happy path + JSON round-trip + one targeted rejection each. Plus three tests for the new logging helper. **136 tests pass** (was 118), typecheck + lint clean (0 errors).

**🔵 ApiError shape / § 6 docs clarification.** Agreed it's a separate docs task — I've written it up in a handoff to Cowork (`.agents/handoffs/cowork.md`) bundled with a broader `api-contract.md` § 2 reconciliation Brendan flagged (§ 2 reads as a firm "generate the client" decision that contradicts the hand-rolled reality + the codegen research note). Not touching the standard in this PR.

**Codecov 76.85% patch.** The uncovered lines are the form-bearing component paths (`Profile.tsx`, `Onboarding.tsx`, `DrinkTypeSelector.tsx`) and the JWT branch — those get their RTL tests in PR-E1 (forms → `useActionState`). The frontend project gate is informational until PR-H1's backfill, so non-blocking.

docs/coding-standards/typescript.md
```diff
@@ -371,8 +371,8 @@ Each row in this table maps to an ESLint rule listed in § 9.
...
-  - **Why:** Every tool that parses Rust source directly (typeshare, ts-rs, specta) fails on `nutype`-generated structs because `nutype` rewrites the struct during macro expansion. `schemars` works because `nutype` has a first-class `schemars08` feature flag. Hand-rolling preserves optionality and keeps the contract legible to both Cowork and Claude Code in the meantime.
+- **Rule:** Type-sync between Rust DTOs and TS DTOs is currently hand-maintained, with the Zod schemas (per § 8) as the source of truth and TS types derived via `z.infer<typeof Schema>`. CI does not yet enforce drift detection. See [`../research/rust-to-ts-codegen.md`](../research/rust-to-ts-codegen.md) for the evaluation of automated options (typeshare, ts-rs, specta, schemars, utoipa); the 2026-05-21 addendum names the at-threshold path as `schemars` + [`json-schema-to-zod`](https://www.npmjs.com/package/json-schema-to-zod) + a thin brand-mint overlay (generates the runtime Zod schemas, with [`frontend/src/api/brand.ts`](../../frontend/src/api/brand.ts) swapping in the branded `*IdSchema` for ID fields). The trigger is reframed from "DTO count ~30" to "Zod-maintenance friction," with DTO count as a proxy.
```
brendanbyrne: 🔵 **Suggestion** — This commit substantively edited § 11 (source-of-truth statement + recommended codegen path), but `typescript.md`'s `## Document history` section didn't get a matching entry. Per [`docs/CLAUDE.md`](docs/CLAUDE.md) § Document history rule, `coding-standards/*` files are in-scope. The companion `rust-to-ts-codegen.md` *did* get its 2026-05-21 entry — `typescript.md` just needs a one-line bullet to match. Trivial; fold into this PR or a docs follow-up.

brendanbyrne (commented):
## Re-review — `5c09a92` (address PR-B2 review feedback)

**Verdict:** Approve ✅ — all five prior findings addressed cleanly; one trivial docs-hygiene nit (inline).

**Automated checks (re-run on this commit):**
- `bun run typecheck` — clean ✅
- `bun run test` — 136 passed (+18 from the new tests), 2 todo ✅
- `bun run lint` — 0 errors; touched files clean (the two `import/no-duplicates` warnings are pre-existing in `RunEntrySheet.tsx`, untouched here)

### Findings resolution

| # | Prior finding | Status |
|---|---|---|
| 1 🟡 | Shape-mismatch swallowed in 4 sites | ✅ **Resolved.** New `logIfResponseShapeMismatch(error, context)` logs only when the caught error is an `ApiErrorException` with `code === 'response_shape_mismatch'`; network errors still degrade quietly. Applied at all four sites (`useGameData` ×2, `useUserProfile`, `client.ts` `tryRefresh`). Tested 3 ways. Exactly the surgical § 8 fix, with PR-C1 still owning the eventual retirement. |
| 2 🔵 | JWT payload decode bypassed Zod | ✅ **Resolved.** `AccessTokenPayloadSchema` added; `const raw: unknown = JSON.parse(atob(...))` then `.parse(raw)`. Closes the last `any` boundary and clears the two latent `no-unsafe-member-access` warnings. |
| 3 🔵 | `mySessionSchema` casing | ✅ **Resolved.** Promoted to a top-level `MySessionResponseSchema` export in `types.ts`. |
| 4 🔵 | `typescript.md` § 6 `ApiError`-shape clarification | ✅ **Deferred correctly** — flagged to Cowork as a separate docs task. Bonus: § 11 and the codegen research doc were refined to record Zod as the type-sync source of truth and `json-schema-to-zod` as the at-threshold target — accurate and well-reasoned. |
| 5 🔵 | Partial response-schema test coverage | ✅ **Resolved.** `RunDetailSchema`, `RunDefaultsSchema`, `AuthSessionSchema`, `TokenRefreshSchema`, `AccessTokenPayloadSchema` each get happy-path + JSON round-trip + one rejection, via `it.each`. |

Nice work — the `logIfResponseShapeMismatch` helper is a cleaner answer than the per-catch inline check I suggested, and the `it.each` coverage closes the § 12 gap economically. Approving once the one docs nit below is folded in (or split to a follow-up).

brendanbyrne:
Pushed \`5b5736d\` — two things, both housekeeping from the \`api-contract.md\` renumber that landed on main as part of ADR 0039:

1. **Fixed this PR's own stale section refs.** \`api-contract.md\` § 2 (API client generation) was deleted and §§ 3–11 renumbered to §§ 2–10, so the doc-comment cross-refs I'd written here were off by one: \`result.ts\` error-registry § 8 → § 7 and error-envelope § 3 → § 2; \`types.ts\` auth-refresh § 5 → § 4.

2. **Removed two docs that didn't belong in this PR.** A \`git add -A\` back in \`5c09a92\` accidentally swept in two of Cowork's in-progress doc edits (\`coding-standards/typescript.md\`, \`research/rust-to-ts-codegen.md\`) — that's the typescript.md edit you spotted in the re-review. Those (plus the missing § 11 history entry you flagged) have landed on **main directly** as a docs commit (\`953c9bc\`), alongside the new [ADR 0039](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/decisions/0039-api-client-generation.md). This PR's net diff is now **frontend-only** — confirmed \`git diff <merge-base> HEAD\` touches no docs.

So the re-review nit (typescript.md history entry) is resolved on main, not here. Backend + \`.claude/skills\` cross-reference fixes for the same renumber are coming as a separate small sweep PR. Nothing else changed in the code.

brendanbyrne:
**Correction + fix: the frontend patch-coverage gate was blocking, and was failing.**

Earlier in this thread I said frontend coverage was "informational, so non-blocking." That was wrong — I conflated two separate Codecov statuses in `codecov.yml`:

- `project/frontend` *is* `informational: true` (non-blocking until the PR-H1 backfill).
- `patch/frontend` is **`informational: false` — blocking** (80% of new/changed lines), and it was failing at **77.78%**.

The `api/` boundary this PR introduces was already 100% patch-covered; the gap was the consumer-side call sites it touched. Fix (latest commit) adds behavior tests at the fetch boundary (MSW):

- `client.test.ts` — the 401 → refresh → retry flow (transparent token refresh, auth-failure on a dead refresh cookie, malformed-refresh handling). Previously 0% covered and the largest changed block.
- `DrinkTypeSelector.test.tsx` — add-custom-drink success/failure.
- `Onboarding.test.tsx` — a save failure surfaces the backend error and doesn't advance the wizard.

Patch coverage **77.78% → 82.54%**; `codecov/patch/frontend` now passes. The remaining uncovered changed lines are in the direct-`apiFetch` data hooks (`useGameData`/`useUserProfile`) that PR-C1 (#176) rewrites to TanStack Query — they get their tests there rather than throwaway ones here.

frontend/src/api/result.ts
```diff
@@ -0,0 +1,183 @@
...
+}
+
+function isApiErrorCode(code: string): code is ApiErrorCode {
+  // The `as readonly string[]` only widens the literal tuple so `.includes`
+  // accepts an arbitrary string — it is not an unsafe narrowing cast.
+  return (API_ERROR_CODES as readonly string[]).includes(code);
+}
+
+/**
+ * The wire error envelope. `code` is optional: § 3 says the backend always
```
brendanbyrne: 🟡 **Stale section ref** — this bare `§ 3` refers to api-contract.md's Error response contract (the section stating the backend always sends `code`) — the *same* section `5b5736d` renumbered to § 2 on line 59. This bare ref was missed by that cleanup. On main (post-ADR-0039), § 3 is now "Polling & conditional GETs," so the reference is wrong. Should be `§ 2`:

```
 * The wire error envelope. `code` is optional: § 2 says the backend always
```

brendanbyrne (commented):
## Re-review — `5b5736d` (ref fix + docs revert) and `2221159` (patch-coverage tests)

**Verdict:** Approve ✅ — both commits are solid; one trivial stale section-ref the renumber cleanup missed (inline).

**Automated checks (on `2221159`):**
- `bun run typecheck` — clean ✅
- `bun run test` — **147 passed** (up from 136), 2 todo ✅
- `bun run lint` — 0 errors, 161 warnings (the JWT `no-unsafe-member-access` pair is now gone; the two "Unused eslint-disable" warnings are in gitignored `frontend/coverage/` artifacts, not the PR)

### `5b5736d` — section-ref fix + docs revert
- ✅ **Reverted the accidentally-swept-in `typescript.md` § 11 + research-addendum edits**, making the PR frontend-only as intended. This also cleanly moots my prior re-review's Document-history nit by removing the edit entirely — those docs land on main directly. Good call.
- ✅ **Updated api-contract.md refs in code comments to match main's post-ADR-0039 renumbering** (main deleted the old § 2 "API client generation" and shifted the rest): error registry § 8→§ 7, error envelope § 3→§ 2 ([`result.ts:59`](frontend/src/api/result.ts#L59)), auth refresh § 5→§ 4 ([`types.ts:333`](frontend/src/api/types.ts#L333)). All three correct against main; forward-referencing main's numbering is the right move since this branch predates the renumber.
- 🟡 One bare ref missed — inline below.

### `2221159` — patch-coverage tests
- ✅ Excellent, behavior-focused additions. [`client.test.ts`](frontend/src/api/client.test.ts) covers the full 401→refresh→retry flow, auth-failure on a dead cookie, malformed-refresh handling (exercises the `logIfResponseShapeMismatch` path and asserts `console.error` fired — verifying the § 8 fail-loud fix), `AbortSignal` forwarding, and the no-token case. `DrinkTypeSelector` and `Onboarding` tests cover add-drink and save-failure paths. Patch coverage 77.78% → 82.54%; the remaining gaps are the `useGameData`/`useUserProfile` hooks that PR-C1 (#176) rewrites to TanStack Query.

Nothing blocking from my side beyond the one stale ref. Sign-off and merge remain yours.

frontend/src/api/result.ts
brendanbyrne: Fixed in `0f80f37`. Same section as the line-59 ref `5b5736d` caught — that cleanup updated the discriminated-union comment but missed this bare `§ 3` in the `errorEnvelopeSchema` doc comment, which points at the same Error-response section. Now `§ 2`. Swept the rest of `src/api/` while here: the remaining `§ 7`/`§ 4` refs (registry / auth-refresh) are correct against main, and the `§ 8` refs are typescript.md (fail-loud), not api-contract — so this was the last stale one.

Issue #186, opened 2026-05-18
Migrate polling hooks (useSession, useSessions) to TanStack Query


brendanbyrne:
## Context

Sixth PR. Replaces the hand-rolled polling/visibility-API/`cancelled`-flag machinery in `useSession` and `useSessions` with TanStack Query's `refetchInterval` + `refetchIntervalInBackground: false`. Also migrates `useAuth` mutations, `BottomNav`'s session-check `useEffect`, and `RunEntrySheet`'s defaults-loading `useEffect`. This is the largest deletion of code in the compliance plan.

## Acceptance criteria

- [ ] Migrate `useSession.ts`: `useQuery({ refetchInterval: (q) => (q.state.data?.ended_at ? false : 2500), refetchIntervalInBackground: false })`. Delete the manual polling/visibility/`endedRef` code.
- [ ] Migrate `useSessions.ts`: same pattern with 5000 ms interval.
- [ ] Replace `useAuth.tsx`'s `login` / `register` / `logout` / `changePassword` direct `fetch` calls with `useMutation` (or plain async called from the context if mutation state isn't needed).
- [ ] Replace `BottomNav.tsx`'s `useEffect`-on-pathname for `getMySession()` with `useQuery({ queryKey: ['my-session'] })` invalidated by the relevant session mutations.
- [ ] Replace `RunEntrySheet.tsx`'s defaults-loading `useEffect` (lines 60-75) with `useQuery({ queryKey: ['run-defaults'] })`.
- [ ] Tab-switch test: open a session, switch tabs 30 s, return — polling resumes without a backfill spike.
- [ ] Cross-device test: submit a run on one device, confirm it appears on another within ~5 s.
- [ ] No `setState on unmounted component` warnings in dev.

## References

- Compliance plan: PR-C2
- Audit findings on every `useEffect`-driven fetch hook
- Standards: `docs/coding-standards/react.md` § 4, § 6
- Polling model: `docs/decisions/0018-realtime-via-polling-not-websockets.md`

_Filed by Cowork on Brendan's behalf._

Issue #190, opened 2026-05-18
Upgrade router to createBrowserRouter with error boundaries and lazy routes


brendanbyrne:
## Context

Twelfth PR. Migrates `App.tsx` from the legacy `BrowserRouter` + `<Routes>` form to react-router 7's `createBrowserRouter` + `RouterProvider`. Adds an app-level error boundary, per-route `errorElement`, and `lazy()`-loads every page. Also fixes `BottomNav.tsx`'s `<button onClick={navigate}>` → `<NavLink>` regression. This is also where the strictest lint rules flip from warn to error.

## Acceptance criteria

- [ ] Migrate `App.tsx` from `BrowserRouter` + `<Routes>` to `createBrowserRouter` + `RouterProvider`.
- [ ] Add an app-level `<ErrorBoundary>` from `react-error-boundary` with a generic fallback ("Something went wrong. Reload?").
- [ ] Add per-route `errorElement` for each top-level route.
- [ ] `lazy()`-load the six page imports; wrap each in `<Suspense fallback={<PageSkeleton />}>`.
- [ ] Replace `BottomNav.tsx`'s `<button onClick={navigate}>` with `<NavLink>`; set `aria-current="page"` via NavLink's className function.
- [ ] Flip `no-explicit-any` and `no-non-null-assertion` from warn to error in `eslint.config.js`.
- [ ] Every existing route navigates correctly.
- [ ] Triggering a render error in one route (temporary `throw new Error('test')` in a page) shows the route-scoped fallback, not the global one.
- [ ] Bundle analyzer shows separate chunks per route.

## References

- Compliance plan: PR-F1
- Standards: `docs/coding-standards/react.md` § 9, § 11

_Filed by Cowork on Brendan's behalf._

PR #203, opened 2026-05-22
176: TanStack Query setup + migrate static-data hooks (PR-C1)

### LINKED ISSUE #176: Set up TanStack Query and migrate static-data hooks

brendanbyrne:
## What & why

PR-C1 in the [frontend compliance plan](docs/designs/2026-05-16-frontend-compliance-plan.md) (Issue #176). Sets up TanStack Query for the app and migrates the **static-data** hooks off hand-rolled `useEffect` fetching. Polling hooks (`useSession`/`useSessions`/`useAuth`/etc.) are **out of scope** — that's PR-C2 (#186).

Standards: react.md § 4 (data fetching), § 6 (effects), § 13 (testing).

### Changes

- **`App.tsx`** — wraps the tree in `<QueryClientProvider>` with the plan's default config (`refetchOnWindowFocus: true`, `retry: 1`, `staleTime: 30_000`). The client is created at module scope so it survives re-renders. React Query Devtools mounted dev-only (`import.meta.env.DEV`).
- **`useGameData.ts`** — the five `useSimpleList` hooks (`useCharacters`/`useBodies`/`useWheels`/`useGliders`) + `useDrinkTypes` now use `useQuery` with a 1-hour `staleTime` (static reference data, no polling). `useDrinkTypes`' `refresh()` now calls `queryClient.invalidateQueries({ queryKey: ['drink-types'] })` instead of bumping a version counter. The legacy `{ items, loading }` / `{ items, loading, refresh }` return shapes are preserved, so **no call site changed** (`DrinkTypeSelector`, `RaceSetupPicker`, `RunEntrySheet`, `Profile`, `Home`).
- **`useUserProfile.ts`** — `useQuery` keyed on `['user-profile', userId]`, disabled until `userId` is known; `refresh()` invalidates that key. Same `{ profile, loading, refresh }` shape.
- **Tests** — fills the two `it.todo` hook-test placeholders B2 left in `useGameData.test.tsx`; adds `useUserProfile.test.tsx`, an app-shell smoke test (`__tests__/App.test.tsx`), and wraps the existing `DrinkTypeSelector.test.tsx` render in a `QueryClientProvider` (required now that `useDrinkTypes` is a Query hook). Pattern per react.md § 13: `renderHook` under a provider, network mocked with **MSW** at the fetch boundary.

### Behavior note

Both hooks keep the old "degrade to an empty list/null on failure" behavior: TanStack Query holds the error in its own state (no `throwOnError`), so components render empty rather than crashing — same external contract as before. Loud error surfacing comes later with the error-boundary work.

## How to verify

Run from `frontend/`. Each command below was run on this branch before opening the PR.

```bash
cd frontend

# 1. Types compile (tsc -b; ~18-20s on this checkout)
bun run typecheck        # → "$ tsc -b" then exits 0, no errors

# 2. Lint: 0 errors (pre-existing warn-down warnings remain)
bun run lint             # → "✖ N problems (0 errors, N warnings)"

# 3. Full test suite, including the new hook tests
bun run test             # → "Test Files 13 passed (13)  Tests 158 passed (158)"

# 4. Coverage — frontend patch gate is 80% of changed lines
bun run test:coverage    # → useGameData.ts / useUserProfile.ts at 100% lines;
                         #   open frontend/coverage/index.html for the report
```

### Manual smoke test (dev server)

```bash
cd frontend
bun run dev              # Vite serves on http://localhost:5173 (proxies /api to the backend)
```

Then, with the backend running, in the browser:

1. **Devtools panel** — a floating TanStack Query logo appears bottom-right (dev only). Click it; you should see queries like `['characters']`, `['drink-types']`, `['user-profile', <id>]` populate as you navigate.
2. **Pick-lists still work** — open the run-entry sheet / race-setup picker: characters, bodies, wheels, gliders, and drink types all load as before.
3. **`refresh()` → invalidation** — in `DrinkTypeSelector`, add a new custom drink type. It appears in the list immediately (the `invalidateQueries` refetch). In Devtools you'll see `['drink-types']` go stale → fetching → fresh.
4. **Profile loads** — visit Profile/Home; your profile data renders. Editing and saving still reflects (Profile calls `refresh()`).

## Checklist

- [x] Tests added (hook tests + app-shell smoke test; react.md § 13 MSW pattern)
- [x] `bun run typecheck` clean
- [x] `bun run lint` 0 errors
- [x] `bun run test` all green (158 passed)
- [x] `bun run test:coverage` — changed hook files at 100% lines
- [x] Out-of-scope work (polling hooks) deferred to PR-C2 (#186)

Closes #176

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -1,6 +1,7 @@
-import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { QueryClientProvider } from '@tanstack/react-query';
 import { ReactQueryDevtools } from '@tanstack/react-query-devtools';
 import { BrowserRouter, Navigate, Route, Routes } from 'react-router-dom';
+import { createQueryClient } from './api/queryClient';
 import { AuthProvider, useAuth } from './hooks/useAuth';
 import Login from './pages/Login';
 import Register from './pages/Register';
@@ -9,23 +10,10 @@ import Home from './pages/Home';
 import Profile from './pages/Profile';
 import Session from './pages/Session';
 
-/**
- * One QueryClient for the whole app, created at module scope so it survives
- * re-renders (a client recreated in render would drop the cache every time).
- * Defaults per the compliance plan (react.md § 4): revalidate on tab focus,
- * retry a failed fetch once, and treat data as fresh for 30s. Per-query
- * overrides (e.g. the long staleTime on static reference data) live in the
- * hooks themselves.
- */
-const queryClient = new QueryClient({
-  defaultOptions: {
-    queries: {
-      refetchOnWindowFocus: true,
-      retry: 1,
-      staleTime: 30_000,
-    },
-  },
-});
+// One QueryClient for the whole app, created at module scope so it survives
+// re-renders (a client recreated in render would drop the cache every time).
+// Config + the contract-drift logging live in createQueryClient (api layer).
+const queryClient = createQueryClient();
 
 /** Shows a loading spinner while the initial silent refresh is in progress. */
 function AuthGate({ children }: { children: React.ReactNode }) {
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -1,3 +1,5 @@
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { ReactQueryDevtools } from '@tanstack/react-query-devtools';
 import { BrowserRouter, Navigate, Route, Routes } from 'react-router-dom';
 import { AuthProvider, useAuth } from './hooks/useAuth';
 import Login from './pages/Login';
@@ -6,6 +8,24 @@ import Onboarding from './pages/Onboarding';
 import Home from './pages/Home';
 import Profile from './pages/Profile';
 import Session from './pages/Session';
+
+/**
+ * One QueryClient for the whole app, created at module scope so it survives
+ * re-renders (a client recreated in render would drop the cache every time).
+ * Defaults per the compliance plan (react.md § 4): revalidate on tab focus,
+ * retry a failed fetch once, and treat data as fresh for 30s. Per-query
+ * overrides (e.g. the long staleTime on static reference data) live in the
+ * hooks themselves.
+ */
+const queryClient = new QueryClient({
+  defaultOptions: {
+    queries: {
+      refetchOnWindowFocus: true,
+      retry: 1,
+      staleTime: 30_000,
+    },
+  },
+});
 
 /** Shows a loading spinner while the initial silent refresh is in progress. */
 function AuthGate({ children }: { children: React.ReactNode }) {
@@ -38,55 +58,58 @@ function GuestOnly({ children }: { children: React.ReactNode }) {
 
 function App() {
   return (
-    <BrowserRouter>
-      <AuthProvider>
-        <AuthGate>
-          <Routes>
-            <Route
-              path="/login"
-              element={
-                <GuestOnly>
-                  <Login />
-                </GuestOnly>
-              }
-            />
-            <Route path="/register" element={<Register />} />
-            <Route
-              path="/onboarding"
-              element={
-                <RequireAuth>
-                  <Onboarding />
-                </RequireAuth>
-              }
-            />
-            <Route
-              path="/profile"
-              element={
-                <RequireAuth>
-                  <Profile />
-                </RequireAuth>
-              }
-            />
-            <Route
-              path="/session/:id"
-              element={
-                <RequireAuth>
-                  <Session />
-                </RequireAuth>
-              }
-            />
-            <Route
-              path="/"
-              element={
-                <RequireAuth>
-                  <Home />
-                </RequireAuth>
-              }
-            />
-          </Routes>
-        </AuthGate>
-      </AuthProvider>
-    </BrowserRouter>
+    <QueryClientProvider client={queryClient}>
+      <BrowserRouter>
+        <AuthProvider>
+          <AuthGate>
+            <Routes>
+              <Route
+                path="/login"
+                element={
+                  <GuestOnly>
+                    <Login />
+                  </GuestOnly>
+                }
+              />
+              <Route path="/register" element={<Register />} />
+              <Route
+                path="/onboarding"
+                element={
+                  <RequireAuth>
+                    <Onboarding />
+                  </RequireAuth>
+                }
+              />
+              <Route
+                path="/profile"
+                element={
+                  <RequireAuth>
+                    <Profile />
+                  </RequireAuth>
+                }
+              />
+              <Route
+                path="/session/:id"
+                element={
+                  <RequireAuth>
+                    <Session />
+                  </RequireAuth>
+                }
+              />
+              <Route
+                path="/"
+                element={
+                  <RequireAuth>
+                    <Home />
+                  </RequireAuth>
+                }
+              />
+            </Routes>
+          </AuthGate>
+        </AuthProvider>
+      </BrowserRouter>
+      {import.meta.env.DEV && <ReactQueryDevtools initialIsOpen={false} />}
+    </QueryClientProvider>
   );
 }
 
--- a/frontend/src/hooks/useGameData.test.tsx
+++ b/frontend/src/hooks/useGameData.test.tsx
@@ -1,13 +1,150 @@
-import { describe, it } from 'vitest';
+import { http, HttpResponse } from 'msw';
+import { describe, expect, it } from 'vitest';
+import { act, renderHook, waitFor } from '@testing-library/react';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import type { ReactNode } from 'react';
+import { server } from '../mocks/server';
+import {
+  useBodies,
+  useCharacters,
+  useDrinkTypes,
+  useGliders,
+  useWheels,
+} from './useGameData';
 
-// Placeholder — the hook-test pattern.
-//
-// PR-C1 (Issue #176) migrates the static-data hooks in useGameData.ts to
-// TanStack Query. When it lands, fill these in: per
-// docs/coding-standards/react.md § 13, hook tests use `renderHook` wrapped in
-// QueryClientProvider and assert rendered output / side effects (network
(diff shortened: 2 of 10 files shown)


frontend/src/App.tsx
```diff
@@ -7,6 +9,24 @@ import Home from './pages/Home';
...
 
+/**
+ * One QueryClient for the whole app, created at module scope so it survives
+ * re-renders (a client recreated in render would drop the cache every time).
+ * Defaults per the compliance plan (react.md § 4): revalidate on tab focus,
+ * retry a failed fetch once, and treat data as fresh for 30s. Per-query
+ * overrides (e.g. the long staleTime on static reference data) live in the
+ * hooks themselves.
+ */
+const queryClient = new QueryClient({
```
brendanbyrne: 🟡 **Important** — contract-drift visibility regresses to silent (`typescript.md` § 8).

PR-B2 deliberately made `parseBody` *fail loud* on schema mismatch (`response_shape_mismatch`), and the legacy hooks honored that by calling `logIfResponseShapeMismatch(...)` — § 8 says contract drift "must stay visible." This PR drops those calls. The thrown `ApiErrorException` now lands in TanStack Query's `query.error`, but nothing reads it: the hooks don't expose `error`, there's no `QueryCache` `onError`, `throwOnError` isn't set, and error boundaries don't arrive until PR-F1 (#190). TanStack Query v5 removed its default error logger, so a backend shape change on `/characters`, `/drink-types`, `/users/:id`, etc. now degrades to empty/null with **zero console output** — the exact silent drift § 8 exists to prevent.

The user-facing degrade-to-empty contract is preserved and tested, so this isn't Critical — but B2's loud-on-failure is removed with nothing replacing it until F1. Cheap fix that keeps § 8 satisfied in the interim (`logIfResponseShapeMismatch` is still live — `client.ts` uses it):

```ts
import { QueryCache } from '@tanstack/react-query';
import { logIfResponseShapeMismatch } from './api/result';

const queryClient = new QueryClient({
  queryCache: new QueryCache({
    onError: (error, query) => logIfResponseShapeMismatch(error, query.queryHash),
  }),
  defaultOptions: { queries: { refetchOnWindowFocus: true, retry: 1, staleTime: 30_000 } },
});
```

Note `result.ts:169-171` already claims PR-C1 "propagates errors to a boundary" — not true in this end state (no boundary, no `throwOnError`), which is what makes the gap easy to miss. Either add the `onError` above, or, if you'd rather accept a deliberate silent window until F1, say so explicitly so it's on the record.

frontend/src/hooks/useGameData.test.tsx
```diff
@@ -1,13 +1,150 @@
...
+  });
+
+  // The four pick-list hooks are thin wrappers over the same useSimpleList
+  // body, each pointed at its own endpoint. Verify each hits its endpoint and
+  // surfaces the result.
+  it.each([
+    [useBodies, '/api/v1/bodies'],
+    [useWheels, '/api/v1/wheels'],
+    [useGliders, '/api/v1/gliders'],
+  ])('fetches %o from its endpoint', async (hook, endpoint) => {
```
brendanbyrne: 🔵 **Suggestion** — `%o` formats the hook *function*, so these cases render as `fetches [Function: useBodies] from its endpoint`. `frontend/CLAUDE.md` § Testing asks for sentence-style names. Pass an explicit label as a third tuple element and interpolate `%s`:

```ts
it.each([
  [useBodies, '/api/v1/bodies', 'useBodies'],
  [useWheels, '/api/v1/wheels', 'useWheels'],
  [useGliders, '/api/v1/gliders', 'useGliders'],
])('%s fetches its list from its endpoint', async (hook, endpoint) => {
```

brendanbyrne (commented):
## Review — PR-C1: TanStack Query setup + static-data hooks

**Verdict:** Approve with one discussion point (no blockers). Approve/request-changes is reserved for Brendan; this review is `COMMENT`.

**Areas covered:** `react.md` (§ 4 data fetching, § 6 effects, § 13 testing), `typescript.md` (§ 8 contract drift), the frontend compliance plan (PR-C1 scope), `design.md`.

**Automated checks (all pass):**
- `bun run typecheck` → clean (`tsc -b`)
- `bun run lint` → **0 errors**, 157 warnings (all pre-existing warn-down rules from PR-A1; none in the changed files)
- `bun run test` → **158 passed (13 files)**

**Design-doc accuracy:** No changes required — `design.md` doesn't describe the frontend data-fetching layer (that lives in `react.md` + the compliance plan), and the API surface and data model are untouched.

### What's good
- **Faithful contract preservation** — explicit return-type annotations plus `query.data ?? []` / `?? null` defaults keep every call site untouched. Using `isPending` for `loading` is the right call: it correctly reports loading for the *disabled* `useUserProfile` query (matching the legacy "no id → stays loading, never fetches" behavior), which `isLoading` would not.
- **`AbortSignal` threaded** into every `queryFn` via `{ signal }` → `apiFetch`.
- **Tests target the public contract, not internals** — comments call this out explicitly, the `refresh()` test asserts the actual fetch count, and the disabled-query test proves no request fires. Exactly the § 13 renderHook pattern.

Detailed findings inline.

frontend/src/App.tsx
brendanbyrne: Agreed — and you're right that deferring to F1 was the wrong justification: react.md § 9 says error boundaries catch *render* errors, not query errors, so F1 would never have covered this. Fixed in fd11318.

I went slightly beyond the inline snippet: extracted the client into a tested `api/queryClient.ts` factory rather than inlining the `QueryCache` in `App.tsx`, so the § 8 behavior is actually unit-testable. `queryClient.test.ts` drives a query to error through the real client and asserts (a) a `response_shape_mismatch` logs with the resource name in the context, and (b) a plain network error stays silent (preserving the degrade-to-empty contract). `onError` calls `logIfResponseShapeMismatch(error, query.queryHash)` exactly as you suggested.

Also fixed the stale `result.ts` docstring you flagged — it now describes the cache `onError` routing instead of the non-existent boundary.

frontend/src/hooks/useGameData.test.tsx
brendanbyrne: Done in fd11318 — added an explicit label as the third tuple element and switched to `%s`, so the cases now read `useBodies fetches its list from its endpoint`, etc.

brendanbyrne:
Pushed fd11318 addressing the review:

- **🟡 Contract-drift visibility** — restored via a `QueryCache.onError` in a new tested `api/queryClient.ts` factory (replies inline). § 8 is satisfied in the interim again.
- **🔵 `it.each` naming** — explicit `%s` labels.
- **Bonus (also clears the Codecov partial):** `useUserProfile` now uses `skipToken` instead of `enabled` + a defensive `throw`. `skipToken` narrows `userId` to `string` in the populated branch, so the unreachable guard — the one line Codecov flagged as a partial on that file — is gone.

Re-verified: `bun run typecheck` clean · `bun run lint` 0 errors · `bun run test` **160 passed (14 files)** · `bun run test:coverage` — `useGameData.ts` / `useUserProfile.ts` / `queryClient.ts` all 100% lines. The only remaining patch partial is App.tsx's dev-only `import.meta.env.DEV && <ReactQueryDevtools/>` branch (the smoke test stubs `DEV=false` to keep the devtools panel out of jsdom); leaving it as dev-tooling unless you'd prefer it covered.

brendanbyrne (commented):
## Re-review @ `fd11318` — both findings resolved

**Verdict:** Approve. (Approve/request-changes stays with Brendan; this is `COMMENT`.)

**Checks (all pass):** `typecheck` clean · `lint` **0 errors** (157 warnings, unchanged — the two new files added zero) · `test` **160 passed** (+2 from the new `queryClient.test.ts`).

### Finding resolution

**🟡 Contract-drift visibility (`typescript.md` § 8) — resolved, and better than suggested.** Instead of inlining the `QueryCache` in `App.tsx`, this extracts a `createQueryClient()` factory (`api/queryClient.ts`) with the `onError` → `logIfResponseShapeMismatch` wiring. Three things done well:
- **It's tested** — `queryClient.test.ts` drives a query to error through the *real* client and asserts both halves: `response_shape_mismatch` logs (resource name via `queryHash`), network errors stay silent. The § 8 boundary is now regression-locked.
- **The stale `result.ts` comment is fixed**, and the new `queryClient.ts` doc comment correctly notes error boundaries (PR-F1) catch *render* errors, not query errors (`react.md` § 9) — so the cache `onError` is the right permanent home, not a stopgap.
- **Factory-for-testability** is a cleaner seam than a module-scope const.

**🔵 `it.each` naming — resolved.** Now `'%s fetches its list from its endpoint'` with explicit hook-name labels.

### Bonus improvement (unprompted)

`useUserProfile.ts` moved from `enabled` + a defensive `throw` guard to TanStack Query v5's `skipToken` — the idiomatic disable-with-narrowing pattern. The `queryFn` only exists in the branch where `userId` is narrowed to `string`, so the runtime guard disappears; `isPending` still reads `true` while skipped and no request fires. Same external contract, tighter types.

Nothing outstanding from this review.

Issue #182, opened 2026-05-18
Migrate forms to React 19 useActionState + useFormStatus


brendanbyrne:
## Context

Tenth PR. Converts the five forms from controlled inputs + manual `submitting` flags to React 19's `useActionState` + uncontrolled `FormData` reads + a reusable `<SubmitButton>` using `useFormStatus`. Native HTML validation handles the easy cases; Zod (from PR-B2) handles structured submit-time validation.

## Acceptance criteria

- [ ] Convert `Login.tsx` to `useActionState`. Inputs uncontrolled; submit-time Zod schema.
- [ ] Convert `Register.tsx` (twin of Login).
- [ ] Convert `Profile.tsx` password change form.
- [ ] Convert `DrinkTypeSelector.tsx` add-drink-type form.
- [ ] Convert `Onboarding.tsx` two-phase form.
- [ ] Extract a shared `<SubmitButton>` component using `useFormStatus().pending`.
- [ ] Native validation (`required`, `minLength`, `maxLength`, `pattern`) on every input.
- [ ] Each form's happy path and error path tested explicitly.

## References

- Compliance plan: PR-E1
- Standards: `docs/coding-standards/react.md` § 8

_Filed by Cowork on Brendan's behalf._

Issue #192, opened 2026-05-18
Remove non-null assertions (!) from the frontend


brendanbyrne:
## Context

Eighth PR. Removes the eight `!` sites the audit found. Most consequential: `Session.tsx` has five `id!` sites where `useParams<{ id: string }>()` actually returns `Partial<{ id: string }>` — the `!` masks a real bug when a user reaches `/sessions/:id` without an ID. The fix is a route-param check at the top of the component.

## Acceptance criteria

- [ ] `main.tsx:6` — replace `document.getElementById('root')!` with `if (!rootEl) throw new Error('Root element missing')`.
- [ ] `Session.tsx:13, 29, 40, 51, 63` — add a single param-check at the top: `if (!id) return <Navigate to="/" replace />`. Downstream uses become unconditional `id: SessionId`.
- [ ] `RaceSetupPicker.tsx:173-176` — replace `characterId!` / `bodyId!` / `wheelId!` / `gliderId!` with a narrowing function returning either the full setup or `null`. The "Continue" button is disabled when `null`.
- [ ] No `!` assertions remain in `git grep '\!\.' frontend/src` (allowing for `!=` and other false positives — visual review of the matches).
- [ ] The three flows (root mount, session deep-link without ID, RaceSetupPicker completion) still work.

## References

- Compliance plan: PR-D2
- Audit findings: `main.tsx:6`, `Session.tsx:13/29/40/51/63`, `RaceSetupPicker.tsx:173-176`
- Standards: `docs/coding-standards/typescript.md` § 2

_Filed by Cowork on Brendan's behalf._

PR #204, opened 2026-05-22
186: migrate polling hooks to TanStack Query (PR-C2)

### LINKED ISSUE #186: Migrate polling hooks (useSession, useSessions) to TanStack Query

brendanbyrne:
Closes #186. PR-C2 in the [frontend compliance plan](../blob/main/docs/designs/2026-05-16-frontend-compliance-plan.md) (Stream C). Depends on the merged C1 (#176), which put `createQueryClient` + `QueryClientProvider` in place — new queries inherit C1's central `QueryCache.onError` drift-logging, so no per-hook logging here.

## What changed

Replaces the hand-rolled `setInterval` / Page-Visibility / `endedRef` polling machinery with TanStack Query in the session data layer.

| File | Before | After |
|------|--------|-------|
| `hooks/useSession.ts` | `setInterval` + `visibilitychange` + `endedRef` | `useQuery` with `refetchInterval` that returns `false` once the session is closed/404; `refetchIntervalInBackground: false` pauses polling in a hidden tab |
| `hooks/useSessions.ts` | one `Promise.all([listSessions, getMySession])` | two queries — `['sessions']` and `['my-session']` (5s each); the `['my-session']` key is shared with BottomNav |
| `components/BottomNav.tsx` | re-fetch-on-navigation `useEffect` | `useQuery(['my-session'])` — dedupes with `useSessions`, updates via invalidation |
| `components/RunEntrySheet.tsx` | defaults loaded in a `useEffect` that `setState`s | `useQuery(['run-defaults'])`; defaults layered in via `picked ?? default` (no `setState`-in-effect) |

The public return shapes (`{ session, loading, ended }`, `{ sessions, mySessionId, loading }`) are unchanged, so `Session.tsx` / `Home.tsx` consumers were not rewritten.

### Invalidation wiring

Mutations now invalidate the keys they make stale, instead of relying on "the next poll picks it up":

- `Home` create session → `['my-session']`, `['sessions']`
- `Session` join / leave → `['my-session']`, `['sessions']`, `['session', id]`
- `Session` next-track / skip-turn → `['session', id]`
- `Session` run submitted → `['session', id]`

### Scope notes (confirmed with Brendan)

- **`useAuth` left as-is.** `login`/`register`/`logout` are plain async functions, not `useEffect` data-fetching, so they aren't a § 4/§ 6 violation; the plan permits keeping them. Their callers' submit state is PR-E1's (#182) job.
- **Profile password form left to PR-E1.** The `changePassword` fetch lives in `Profile.tsx`'s password form, which E1 converts wholesale to `useActionState` — C2 lifting just that one fetch would mean both PRs half-touch the same form.
- The silent-refresh `useEffect` on mount is auth bootstrap (a legit § 6 effect), untouched.
- One real refactor surfaced: seeding `RunEntrySheet`'s form state from the defaults via a `setState`-in-effect tripped `react-hooks` at **error** level. Fixed by deriving the effective value (`picked ?? default`) instead of copying server data into state.

## Tests added

- [x] `hooks/useSession.test.tsx` — resolves to data, stops polling on a closed session, marks ended on 404, keeps polling while active
- [x] `hooks/useSessions.test.tsx` — both queries resolve into the legacy shape; no-active-session case
- [x] `components/BottomNav.test.tsx` — Session tab enables once a session is found, stays disabled otherwise, navigates on tap
- [x] `components/RunEntrySheet.test.tsx` — pre-fills from defaults, user pick overrides the default, degrades to blank when defaults fail
- [x] `pages/Home.test.tsx` — create-session invalidates `['my-session']` + `['sessions']`
- [x] `pages/Session.test.tsx` — join/leave/next-track/skip-turn/run-submit invalidate the right keys

All MSW-at-the-fetch-boundary, provider-wrapped (react.md § 13). **Patch coverage: 91.5%** (frontend patch gate is 80%, blocking).

## How to verify

All commands run from the `frontend/` directory.

### 1. Automated checks (should all pass)

```bash
cd frontend
bun run typecheck      # expect: no output after "$ tsc -b" (clean)
bun run lint           # expect: warnings only, exit 0 (no "error" lines)
bun run test           # expect: "Test Files  20 passed (20)" / "Tests  176 passed"
bun run test:coverage  # expect: same pass count + a coverage table
```

### 2. Manual smoke test (dev server)

```bash
# From the repo root — runs backend + frontend together (concurrently)
bun run dev
```

Open the printed URL (e.g. `http://localhost:5173`) in Firefox **and** Chrome, log in, then:

1. **Background-tab polling.** Open a session (`/session/<id>`). Open browser DevTools → React Query panel (bottom-right icon, dev only). Confirm `['session', <id>]` refetches about every 2.5s. Switch to another browser tab for ~30s, then come back — polling resumes with **one** catch-up request, not a burst of backfilled ones.
2. **Create updates immediately.** On Home with no active session, tap **Start a Session → Random**. You land in the new session and the bottom-nav **Session** tab is enabled right away (no ~5s wait for the next poll).
3. **Join / leave reflect at once.** From a second account, open the same session and tap **Join Session** — your name appears in the player list within a poll cycle. Tap **Leave Session** — you return Home and the session list updates.
4. **Run submit shows up.** Tap **Submit Time**, fill the times, submit. Your time appears on the session immediately (invalidation), not after a delay.
5. **Run defaults pre-fill.** Open the run-entry sheet again — the drink and race setup are pre-filled from your last run ("From your last run" label). Pick a different drink; your choice sticks.
6. **No console warnings.** Watch the console through all of the above — no "Can't perform a React state update on an unmounted component" warnings.

### DIFF
--- a/frontend/src/components/BottomNav.tsx
+++ b/frontend/src/components/BottomNav.tsx
@@ -1,16 +1,18 @@
-import { useEffect, useState } from 'react';
+import { useQuery } from '@tanstack/react-query';
 import { useLocation, useNavigate } from 'react-router-dom';
 import { getMySession } from '../api/sessions';
 
 export default function BottomNav() {
   const location = useLocation();
   const navigate = useNavigate();
-  const [mySessionId, setMySessionId] = useState<string | null>(null);
 
-  // Re-check active session on navigation so the tab stays in sync
-  useEffect(() => {
-    getMySession().then(setMySessionId);
-  }, [location.pathname]);
+  // Shares the ['my-session'] key with useSessions, so the two dedupe to one
+  // fetch and the session mutations' invalidation keeps this tab in sync —
+  // replacing the old re-fetch-on-navigation useEffect (PR-C2).
+  const { data: mySessionId } = useQuery({
+    queryKey: ['my-session'],
+    queryFn: ({ signal }) => getMySession(signal),
+  });
 
   const sessionMatch = location.pathname.match(/^\/session\/(.+)$/);
   const sessionPath = sessionMatch
--- a/frontend/src/components/BottomNav.tsx
+++ b/frontend/src/components/BottomNav.tsx
@@ -9,6 +9,13 @@ export default function BottomNav() {
   // Shares the ['my-session'] key with useSessions, so the two dedupe to one
   // fetch and the session mutations' invalidation keeps this tab in sync —
   // replacing the old re-fetch-on-navigation useEffect (PR-C2).
+  //
+  // Deliberately no refetchInterval here: on Home, useSessions polls this key
+  // at 5s; on Session/Profile this stays fresh via mount, local-mutation
+  // invalidation, and refetchOnWindowFocus (on by default). The old effect
+  // only re-fetched on navigation — never on a timer — so a sit-idle tab
+  // missing another participant's close is not a regression, and refocus
+  // covers it.
   const { data: mySessionId } = useQuery({
     queryKey: ['my-session'],
     queryFn: ({ signal }) => getMySession(signal),
--- a/frontend/src/pages/Session.tsx
+++ b/frontend/src/pages/Session.tsx
@@ -1,5 +1,6 @@
 import { useState, useEffect } from 'react';
 import { useNavigate, useParams } from 'react-router-dom';
+import { useQueryClient } from '@tanstack/react-query';
 import { useAuth } from '../hooks/useAuth';
 import { useSession } from '../hooks/useSession';
 import {
@@ -21,6 +22,19 @@ export default function Session() {
   const { user } = useAuth();
   const { session, loading, ended } = useSession(id!);
   const navigate = useNavigate();
+  const queryClient = useQueryClient();
+
+  // Invalidate the queries a membership/state change makes stale. Membership
+  // changes (join/leave) ripple into the bottom-nav and Home list as well as
+  // this session's detail; track changes only touch this session's detail.
+  const invalidateMembership = () => {
+    void queryClient.invalidateQueries({ queryKey: ['my-session'] });
+    void queryClient.invalidateQueries({ queryKey: ['sessions'] });
+    void queryClient.invalidateQueries({ queryKey: ['session', id!] });
+  };
+  const invalidateSession = () => {
+    void queryClient.invalidateQueries({ queryKey: ['session', id!] });
+  };
   const [leaving, setLeaving] = useState(false);
   const [joiningSession, setJoiningSession] = useState(false);
   const [joinError, setJoinError] = useState<string | null>(null);
@@ -36,6 +50,7 @@ export default function Session() {
     setLeaving(true);
     try {
       await leaveSession(id!);
+      invalidateMembership();
       navigate('/');
     } catch {
       setLeaving(false);
@@ -47,6 +62,7 @@ export default function Session() {
     setJoinError(null);
     try {
       await joinSession(id!);
+      invalidateMembership();
     } catch (e) {
       setJoinError(e instanceof Error ? e.message : 'Failed to join session');
       setJoiningSession(false);
@@ -58,6 +74,7 @@ export default function Session() {
     setTrackError(null);
     try {
       await nextTrack(id!);
+      invalidateSession();
     } catch (e) {
       setTrackError(e instanceof Error ? e.message : 'Failed to pick track');
     } finally {
@@ -70,6 +87,7 @@ export default function Session() {
     setTrackError(null);
     try {
       await skipTurn(id!);
+      invalidateSession();
     } catch (e) {
       setTrackError(e instanceof Error ? e.message : 'Failed to skip track');
     } finally {
@@ -422,7 +440,12 @@ export default function Session() {
         <RunEntrySheet
           race={currentRace}
           onClose={() => setShowRunEntry(false)}
-          onSubmitted={() => setShowRunEntry(false)}
+          onSubmitted={() => {
+            // The submitted run isn't in the cached session detail yet;
+            // invalidate so it shows up without waiting for the next poll.
+            invalidateSession();
+            setShowRunEntry(false);
+          }}
         />
       )}
     </div>
--- a/docs/designs/2026-05-16-frontend-compliance-plan.md
+++ b/docs/designs/2026-05-16-frontend-compliance-plan.md
@@ -142,7 +142,7 @@ The PRs below address every line item.
 - **Dependencies:** PR-B2 (typed responses).
 - **Risk:** Low. Drop-in for hooks whose external contract stays the same.
 - **Verification:** Every place that calls `useDrinkTypes`/`useUserProfile`/etc. still works. Devtools panel shows the queries.
-- **Sign-off:** [ ]
+- **Sign-off:** [x]
 
 ### PR-C2: TanStack Query migration of polling hooks
 
@@ -159,7 +159,7 @@ The PRs below address every line item.
 - **Dependencies:** PR-C1.
 - **Risk:** Medium. Polling behavior change should be transparent if `refetchInterval` is tuned correctly; verify against a long-running session that you don't pile up requests in background tabs.
 - **Verification:** Open a session, switch tabs for 30 seconds, return — polling resumes without backfill spike. Submit a run on one device, confirm it appears on another within ~5 seconds. No `setState on unmounted component` warnings.
-- **Sign-off:** [ ]
+- **Sign-off:** [x]
 
 ---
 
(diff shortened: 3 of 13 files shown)


brendanbyrne (commented):
## Code review — PR-C2 (Issue #186): polling hooks → TanStack Query

**Verdict: Approve** — two minor suggestions, no blockers.

**What this PR does:** Replaces the hand-rolled `setInterval` + Page-Visibility + `endedRef` polling machinery in the session data layer with TanStack Query (`useSession`, `useSessions`, `BottomNav`, `RunEntrySheet`) and wires mutation-driven cache invalidation in `Home`/`Session`. Adds six co-located test files.

**Areas reviewed:** `react.md` (§§ 4, 6, 11, 13), `typescript.md` (§§ 2, 6, 7), `api-contract.md` (§ 3 polling), `design.md`.

**Automated checks (run locally):** `bun run typecheck` clean · `bun run lint` 0 errors / 155 warnings (all pre-existing patterns — default exports, `interface` vs `type`, void-in-handler — owned by other compliance-plan PRs) · `bun run test` 176 passed (20 files).

### Design-doc accuracy
**No changes required.** This is an implementation-mechanism swap. The polling contract (`api-contract.md` § 3, 2–3s + ETag) is unchanged, and `react.md` § 4 already prescribes exactly this TanStack Query migration. Nothing touches `design.md`'s data model, API surface, or architecture.

### What's good
- The `RunEntrySheet` fix is exactly right: deriving `picked ?? default` instead of copying server data into state sidesteps the setState-in-effect **error** (`react.md` § 6) cleanly, and the comment explains why.
- `useSession`'s `refetchInterval` uses the *real* contract (`status === 'closed'` / null on 404) and the comment explicitly flags that it diverges from the doc's illustrative `ended_at` example — good defensive note.
- Tests mock at the fetch boundary with MSW, assert user-visible behavior and request counts (not internal query state), and are provider-wrapped — textbook `react.md` § 13. Spying on `invalidateQueries` is a clean way to pin the wiring.
- Public return shapes preserved, so consumers stay untouched — minimal blast radius.

Two suggestions inline below. Neither blocks merge.

frontend/src/pages/Session.tsx
```diff
@@ -21,6 +22,19 @@ export default function Session() {
...
   const navigate = useNavigate();
+  const queryClient = useQueryClient();
+
+  // Invalidate the queries a membership/state change makes stale. Membership
+  // changes (join/leave) ripple into the bottom-nav and Home list as well as
+  // this session's detail; track changes only touch this session's detail.
+  const invalidateMembership = () => {
+    void queryClient.invalidateQueries({ queryKey: ['my-session'] });
+    void queryClient.invalidateQueries({ queryKey: ['sessions'] });
+    void queryClient.invalidateQueries({ queryKey: ['session', id!] });
```
brendanbyrne, line 33: 🔵 **Suggestion** — these new `['session', id!]` assertions add two more `id!` to a file that already has ~7, perpetuating the anti-pattern `react.md` § 11 / `typescript.md` § 2 call out. The §11 fix clears all of them at once with a guard at the top:

```tsx
const { id } = useParams<{ id: string }>();
if (!id) return <Navigate to="/" replace />;
```

After that, `id` is narrowed to `string` and every `id!` in the file (including `useSession(id!)` at line 23) can drop the `!`. Not a blocker — it's the established idiom here and the file-wide cleanup may belong to a dedicated compliance PR — but worth a look while you're already in this file.

frontend/src/components/BottomNav.tsx
```diff
@@ -1,16 +1,18 @@
...
-  const [mySessionId, setMySessionId] = useState<string | null>(null);
 
-  // Re-check active session on navigation so the tab stays in sync
-  useEffect(() => {
-    getMySession().then(setMySessionId);
-  }, [location.pathname]);
+  // Shares the ['my-session'] key with useSessions, so the two dedupe to one
+  // fetch and the session mutations' invalidation keeps this tab in sync —
+  // replacing the old re-fetch-on-navigation useEffect (PR-C2).
+  const { data: mySessionId } = useQuery({
```
brendanbyrne, line 19: 🔵 **Suggestion** — this `['my-session']` query has no `refetchInterval` of its own. `useSessions` polls the same key at 5s, but it's only mounted on `Home`. On `Session`/`Profile`, this query stays fresh via mount + invalidation + the default `refetchOnWindowFocus`. That's a slight change from the old re-fetch-on-navigation effect: a server-side close triggered by *another* participant while you sit on a non-Home page won't update the tab until you refocus or a local mutation invalidates `['my-session']`. For a polling app this is acceptable and `refetchOnWindowFocus` (on by default) covers the common case — flagging only so it's a conscious choice rather than an accident.

frontend/src/pages/Session.tsx
brendanbyrne: Agreed, and deferring to PR-D2 (#192) — its scope is exactly the file-wide `useParams` guard (`if (!id) return <Navigate to="/" replace />`) that drops *every* `id!` here, including the `useSession(id!)` at line 23. Doing the guard in C2 would overlap D2 and re-create the "two PRs half-touch one file" problem we just avoided with the password form. Left the two new `id!` consistent with the file's existing idiom so D2 sweeps them all in one pass.

frontend/src/components/BottomNav.tsx
brendanbyrne: Conscious choice — made it explicit in 8c4bc26 with a comment on the query. Worth adding: it's not a regression from the old behavior. The previous effect only re-fetched on `location.pathname` change (navigation), never on a timer, so an idle-on-Profile tab missing another participant's close was already the case. The new version is strictly better — `refetchOnWindowFocus` (default on) plus local-mutation invalidation cover more than navigation-only did. If we later want non-Home pages to catch remote closes while idle, a small `refetchInterval` here is the lever, but I'd rather not add background polling to Profile without a reason.

brendanbyrne (commented):
## Re-review — commit 8c4bc26

**Verdict: Approve** — both suggestions from the first pass are resolved. No new findings.

### How the feedback was addressed
- 🔵 **BottomNav `['my-session']` cadence** — resolved by *making the choice explicit* rather than adding a timer, which is exactly the right call: a second polling source on the shared key would just duplicate `useSessions`. The new comment documents that mount + local-mutation invalidation + default `refetchOnWindowFocus` keep the tab fresh, and that the idle-tab gap is not a regression vs. the old navigation-only effect. Conscious decision, recorded — that's what the flag asked for.
- 🔵 **`RunEntrySheet` setup-override coverage** — new `"lets the user override the default race setup"` test exercises the `picked ?? default` path for the setup IDs (the drink override was already covered; this closes the gap). Clean approach: `RaceSetupPicker` stubbed to a button that fires `onComplete`, and the game-data lists now carry an id-1 default + id-2 override so the rendered summary actually changes. Asserts user-visible output, not internals — matches `react.md` § 13.
- 🔵 **`Session.tsx` `id!`** — left as-is. As flagged, this was explicitly non-blocking and a fit for a dedicated compliance PR, so deferring is fine.

### Re-verified locally
- `bun run typecheck` — clean
- `bun run lint` — 0 errors (warnings unchanged, all pre-existing patterns)
- `vitest run` on the two changed files — `RunEntrySheet` 4 passed, `BottomNav` 3 passed

Also noted: PR-C1 and PR-C2 sign-off boxes checked in the compliance plan — consistent with the design-record sign-off convention; C2's box lands with its implementing PR.

Nothing else to raise. Approve/Request-Changes state remains Brendan's call; this is posted as `COMMENT`.

PR #205, opened 2026-05-22
175: Named exports everywhere

brendanbyrne:
## What & why

PR-D1 of the [frontend compliance plan](../blob/main/docs/designs/2026-05-16-frontend-compliance-plan.md) (Stream D). Converts all default exports to named exports and flips `import/no-default-export` to an error so the rule is now enforced, not just advisory.

Closes #175.

## Changes

- **11 default-export declarations → named exports:** `App`, `BottomNav`, `DrinkTypeSelector`, `RaceSetupPicker`, `RunEntrySheet`, and the `Home` / `Login` / `Onboarding` / `Profile` / `Register` / `Session` pages.
- **Every importer updated**, including `main.tsx`'s `import { App } from './App'` and `App.tsx`'s page imports.
- **`eslint.config.js`:** `import/no-default-export` flipped `'warn'` → `'error'`.
- **Test files updated:**
  - Plain default imports → named imports across the 8 affected `*.test.tsx`.
  - `vi.mock` factories that returned `{ default: ... }` now key on the named export: `RaceSetupPicker` (in RunEntrySheet.test + Onboarding.test), `DrinkTypeSelector` (Onboarding.test), `RunEntrySheet` (Session.test).

## Coverage note

`RaceSetupPicker`, `Register`, and `Profile` have no direct test today, so I checked whether their renamed lines would drag the blocking 80% patch gate. They won't: the only changed lines in those files are import statements and the function-declaration line, and istanbul emits no `DA` (coverable-line) records for either — so Codecov doesn't count them in the patch denominator. No smoke tests needed; this PR adds no new coverable lines in untested files.

## Tests touched

8 test files updated (import style + 4 mock-factory keys). No test logic changed; all 177 tests across 20 files still pass.

## Manual test steps

All commands run from the **repo root** (root Cargo + Bun workspaces — don't `cd` into `backend`/`frontend` first):

```bash
# 1. Static checks (run in the frontend workspace — these scripts are frontend-only)
cd frontend
bun run typecheck   # tsc -b — clean
bun run lint        # 0 errors (warnings remain for later PRs); the no-default-export rule is now at error level
bun run test        # 177 passed
bun run build       # tsc -b && vite build — succeeds
cd ..

# 2. Boot the app from the repo root and click through the core flow
bun run dev         # starts backend (cargo run) + frontend (vite) together
```

Then in the browser:
1. **Login** — log in with an existing account; you land on Home.
2. **Home** — the home screen renders (BottomNav visible).
[...]

PR #198, opened 2026-05-18
171: Branded ID types + `type` over `interface` in api/types.ts

brendanbyrne:
Closes #171

PR-B1 of the [frontend compliance plan](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/designs/2026-05-16-frontend-compliance-plan.md). First PR in the plan that lands with tests (PR-H2 sequenced the test infra ahead of it for exactly this).

## What changed

- **New `src/api/brand.ts`** — the `Brand<T, B>` phantom-symbol helper (typescript.md § 3) plus 11 branded ID types, each with a same-named constructor:
  - string brands: `UserId`, `SessionId`, `RunId`, `RaceId`, `DrinkTypeId`
  - number brands: `CharacterId`, `BodyId`, `WheelId`, `GliderId`, `TrackId`, `CupId`
- **`api/types.ts`** — every `interface` → `type` (clears its `consistent-type-definitions` warnings; that rule flips to error in PR-H1). Raw `string`/`number` ID fields on the response DTOs replaced with branded types.
- **Literal unions** — `SessionDetail.status` → `SessionStatus` (`'active' | 'closed'`), `SessionSummary.ruleset` / `SessionDetail.ruleset` → `SessionRuleset` (`'random' | 'default' | 'least_played' | 'round_robin'`). **Verified against the backend** `SessionStatus` / `SessionRuleset` enums in `backend/src/domain/enums.rs` — the plan's guessed `'open' | 'in_progress' | 'closed'` was wrong (the plan explicitly said to verify).
- **`api/sessions.ts` / `api/runs.ts`** — accept branded ID params; mint brands at the helper boundary (an `as` cast on the parsed `res.json()` body, or the `SessionId` constructor in `getMySession`). PR-B2 replaces these mint sites with Zod parses.
- **`useSession` / `useSessions` / `Session.tsx`** — thread `SessionId` through. `Session.tsx` types its route param via `useParams<{ id: SessionId }>()` — the URL boundary is where a raw string becomes a `SessionId`.

## Scope decisions (flagging for review)

Two ID fields are **deliberately left raw** — both judgement calls, happy to revisit:

[...]

PR #197, opened 2026-05-18
193: Vitest + RTL + MSW + coverage CI scaffolding

brendanbyrne:
Closes #193

## What this does

PR-H2 of the [frontend compliance plan](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/designs/2026-05-16-frontend-compliance-plan.md) — the frontend test infrastructure, sequenced ahead of the runtime-behavior PRs (B1, B2, C1, C2, E1, F1) so each of them lands with tests. No production code changes; this is pure infrastructure.

- **Deps:** `vitest`, `@testing-library/{react,user-event,jest-dom}`, `jsdom`, `msw`, `@vitest/coverage-istanbul` added to `frontend` devDependencies (via `bun add -d`). Scripts: `test` (`vitest run`), `test:watch` (`vitest`), `test:coverage` (`vitest run --coverage`).
- **`frontend/vitest.config.ts`:** jsdom environment, istanbul coverage, test glob `src/**/*.test.{ts,tsx}` + `src/__tests__/**`, `setupFiles` → `setupTests.ts`.
- **MSW:** `src/mocks/handlers.ts` (empty — later PRs add per-endpoint handlers), `src/mocks/server.ts` (Node server), `src/setupTests.ts` (MSW start/reset/stop, `jest-dom` matchers, RTL cleanup).
- **Example tests:** `src/utils/time.test.ts` (unit — formatter + parser, 10 cases), `src/pages/Login.test.tsx` (RTL — 3 cases). `it.todo` placeholders for PR-B2's schema test (`src/api/types.test.ts`) and PR-C1's hook test (`src/hooks/useGameData.test.tsx`).
- **ESLint:** a test-file override block relaxes the rules `typescript.md` § 12 explicitly permits in test code (`no-explicit-any`, `no-non-null-assertion`, `no-unsafe-*`). Everything else still applies to tests.
- **CI:** new path-filtered `.github/workflows/frontend.yml` runs `bun run test:coverage` and uploads to Codecov under `flags: frontend`. The backend coverage workflow gains a matching `backend/**` path filter (and is renamed `coverage.yml` → `backend.yml`, workflow name `Coverage` → `Backend`), so each side's CI skips on the other's PRs.
- **Codecov:** `codecov.yml` restructured into per-flag (`backend`/`frontend`) project + patch statuses with carryforward; `frontend/**` un-ignored. Frontend **patch** gate blocks at 80%; frontend **project** gate is `informational` until PR-H1's backfill (tracked on #185).
[...]

PR #208, opened 2026-05-26
192: remove non-null `!` assertions from the frontend

### LINKED ISSUE #192: Remove non-null assertions (!) from the frontend

brendanbyrne:
## Summary

PR-D2 of the frontend compliance plan. Removes all eight `!` assertions the audit flagged, plus catches up sign-off bookkeeping for the first eight PRs in the pickup order.

- `main.tsx` — root-mount uses an explicit `if (!rootEl) throw` instead of `!`.
- `Session.tsx` — `useParams<{ id: SessionId }>()` actually returns `Partial<{ id: SessionId }>` regardless of the route's path, so the five `id!` sites masked a real deep-link-without-id bug. Splits the page into a thin `Session` wrapper that does `if (!id) return <Navigate to="/" replace />` and a `SessionView` body that takes `id: SessionId` as a prop. Every downstream use is now unconditional. Adds a test for the redirect path.
- `RaceSetupPicker.tsx` — derives a `setup` object that is non-null only when all four ids are picked. The Confirm button is disabled until then; the four `!` assertions are gone.

Sign-off bookkeeping (docs only): the table rows and per-section boxes for PR-D1 ([#205](https://github.com/brendanbyrne/beerio-kart/pull/205)) and PR-H2 ([#197](https://github.com/brendanbyrne/beerio-kart/pull/197)) had been missed in earlier reconciliations and are now checked; **Merged PR** links added for H2 ([#197](https://github.com/brendanbyrne/beerio-kart/pull/197)), B1 ([#198](https://github.com/brendanbyrne/beerio-kart/pull/198)), B2 ([#199](https://github.com/brendanbyrne/beerio-kart/pull/199)), C1 ([#203](https://github.com/brendanbyrne/beerio-kart/pull/203)), C2 ([#204](https://github.com/brendanbyrne/beerio-kart/pull/204)), and D1 ([#205](https://github.com/brendanbyrne/beerio-kart/pull/205)) so every shipped PR section carries its merged PR alongside its Issue.

Closes #192.

## Verification

- `bun run --cwd frontend typecheck` — clean.
- `bun run --cwd frontend lint` — 0 errors, 126 pre-existing warn-down warnings; no `no-non-null-assertion` among them (was 8 before).
- `bun run --cwd frontend test` — 178/178 passing (including the new no-id redirect test).
- `bun run --cwd frontend build` — production build succeeds.
- `git grep '!\.\|![,)};]' frontend/src` returns only `Home.tsx:70`'s `username}!` punctuation, no assertions.

## Test plan

- [x] `cd` to repo root, run `bun run dev`. Open the app; it mounts cleanly (no "Root element missing" error).
- [x] Log in, open a session via the Home list, confirm the session page renders, the join/leave/next-track/skip-track buttons all still work end-to-end.
- [x] Manually navigate to `/session` (no `:id` segment) in the URL bar — the app redirects to `/` (home) without throwing.
- [x] Onboarding flow: pick a character, body, wheel, glider, then hit the confirm button. The Confirm button stays disabled until all four are picked and fires `onComplete` with the full setup once they are. Repeat from the run-entry sheet's setup picker.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -93,8 +93,12 @@ export function App() {
                 }
               />
               {/* Catch-all: unknown paths (e.g. /session with no :id) redirect
-                  home rather than rendering an empty <Routes>. PR-F1 will
-                  replace this with a real 404/errorElement. */}
+                  home rather than rendering an empty <Routes>. Load-bearing
+                  for the deep-link-without-id case — Session.tsx's
+                  `if (!id) <Navigate />` guard can't fire from real browser
+                  navigation because <Route path="/session/:id"> requires the
+                  param, so this catch-all is what users actually hit. PR-F1
+                  (#190) will replace this with a real 404 / errorElement. */}
               <Route path="*" element={<Navigate to="/" replace />} />
             </Routes>
           </AuthGate>
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -92,6 +92,10 @@ export function App() {
                   </RequireAuth>
                 }
               />
+              {/* Catch-all: unknown paths (e.g. /session with no :id) redirect
+                  home rather than rendering an empty <Routes>. PR-F1 will
+                  replace this with a real 404/errorElement. */}
+              <Route path="*" element={<Navigate to="/" replace />} />
             </Routes>
           </AuthGate>
         </AuthProvider>
--- a/frontend/src/components/RaceSetupPicker.tsx
+++ b/frontend/src/components/RaceSetupPicker.tsx
@@ -108,11 +108,16 @@ export function RaceSetupPicker({
     }
   }
 
-  const allSelected =
+  // Build the completed setup only when all four ids are picked. The Confirm
+  // button is disabled until `setup` is non-null, so `onComplete(setup)`
+  // never sees a partial value — the narrowing replaces four `!` assertions.
+  const setup =
     characterId !== null &&
     bodyId !== null &&
     wheelId !== null &&
-    gliderId !== null;
+    gliderId !== null
+      ? { characterId, bodyId, wheelId, gliderId }
+      : null;
 
   return (
     <div className="flex flex-col h-full">
@@ -190,18 +195,11 @@ export function RaceSetupPicker({
         )}
         <button
           onClick={() => {
-            if (allSelected) {
-              onComplete({
-                characterId: characterId!,
-                bodyId: bodyId!,
-                wheelId: wheelId!,
-                gliderId: gliderId!,
-              });
-            }
+            if (setup) onComplete(setup);
           }}
-          disabled={!allSelected}
+          disabled={!setup}
           className={`flex-1 py-2.5 text-sm font-semibold rounded-xl transition-colors ${
-            allSelected
+            setup
               ? 'bg-blue-500 text-white hover:bg-blue-600'
               : 'bg-gray-200 text-gray-400 cursor-not-allowed'
           }`}
--- a/docs/designs/2026-05-16-frontend-compliance-plan.md
+++ b/docs/designs/2026-05-16-frontend-compliance-plan.md
@@ -88,7 +88,7 @@ The PRs below address every line item.
 
 ### PR-B1: Branded ID types + `type` over `interface` in `api/types.ts`
 
-**Issue:** [#171](https://github.com/brendanbyrne/beerio-kart/issues/171)
+**Issue:** [#171](https://github.com/brendanbyrne/beerio-kart/issues/171) · **Merged PR:** [#198](https://github.com/brendanbyrne/beerio-kart/pull/198)
 
 - **Scope:**
   - Create `src/api/brand.ts` with the `Brand<T, B>` helper and exported brand types: `UserId`, `SessionId`, `RunId`, `RaceId`, `DrinkTypeId` (string brands); `CharacterId`, `BodyId`, `WheelId`, `GliderId`, `TrackId`, `CupId` (number brands).
@@ -106,7 +106,7 @@ The PRs below address every line item.
 
 ### PR-B2: Runtime-validated API responses (Zod)
 
-**Issue:** [#191](https://github.com/brendanbyrne/beerio-kart/issues/191)
+**Issue:** [#191](https://github.com/brendanbyrne/beerio-kart/issues/191) · **Merged PR:** [#199](https://github.com/brendanbyrne/beerio-kart/pull/199)
 
 - **Scope:**
   - Add Zod schemas for every API response shape in `src/api/types.ts`. Infer the TS types from the schemas (delete the hand-written types in favor of `z.infer<typeof ...>`).
@@ -129,7 +129,7 @@ The PRs below address every line item.
 
 ### PR-C1: TanStack Query setup + migration of static-data hooks
 
-**Issue:** [#176](https://github.com/brendanbyrne/beerio-kart/issues/176)
+**Issue:** [#176](https://github.com/brendanbyrne/beerio-kart/issues/176) · **Merged PR:** [#203](https://github.com/brendanbyrne/beerio-kart/pull/203)
 
 - **Scope:**
   - Wrap `App.tsx` in `QueryClientProvider`. Use the recommended default config (`refetchOnWindowFocus: true`, `retry: 1`, `staleTime: 30_000` for most queries).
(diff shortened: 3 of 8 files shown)


brendanbyrne (commented):
## Verdict: Approve

Removes the eight `!` assertions the audit flagged; splits `Session` into a thin guard wrapper + `SessionView` body so `id` narrows once; adds a catch-all route so `/session` (no `:id`) actually redirects; sign-off bookkeeping in the compliance plan.

### What was verified

- `bun run typecheck` — clean.
- `bun run lint` — 0 errors, 126 pre-existing warn-down warnings; no `no-non-null-assertion` among them (was 8 before).
- `bun run test` — 179/179 passing (includes the new no-id redirect test and the new catch-all test).
- `grep -E '\!\.|![,);}\]]' frontend/src/` — only the punctuation match in `Home.tsx:70` (`username}!`) and a `ms!` inside `utils/time.test.ts`, which `typescript.md` § 12 explicitly allows in test code.
- Standards walked: `typescript.md` § 2 / § 10 / § 12, `react.md` § 11 / § 13.
- **Design-doc accuracy:** no changes required — architecture, data model, and API surface are untouched.

### What's good

- **`SessionView` extraction** is the textbook fix for the `useParams` narrowing issue called out by name in `typescript.md` § 2 ("Most `!` is 'I know better than the compiler' and is wrong eventually (see `Session.tsx:13` in the audit…)"). All downstream `id` uses are unconditional `SessionId`, no `id!` left.
- **Branded `useParams<{ id: SessionId }>()`** plus `<SessionView id={id} />` keeps the brand minted exactly at the route boundary — no scattered `as SessionId` re-mints inside the page.
- **`main.tsx` rewrite** matches `typescript.md` § 10's "Narrow with `if (x)`" remediation for `!`.
- **`RaceSetupPicker`** narrowing — folding `allSelected: boolean` plus four `!` reads into one `setup: Setup | null` is the cleanest shape; the disabled state and the click handler now share a single source of truth instead of four parallel ones.
- **Tests added** for both the in-component redirect path and the App-level catch-all. The `Session.test.tsx` redirect test is the right shape — it renders at `/session` directly to exercise the guard that production traffic can't reach (because of the catch-all). See the inline comment on `App.tsx` for the full two-layer-defense note.

Two suggestions inline. Not blocking; both are about comment clarity / test-coverage parity with sibling components.

frontend/src/App.tsx
```diff
@@ -92,6 +92,10 @@ export function App() {
                   </RequireAuth>
                 }
               />
+              {/* Catch-all: unknown paths (e.g. /session with no :id) redirect
+                  home rather than rendering an empty <Routes>. PR-F1 will
+                  replace this with a real 404/errorElement. */}
+              <Route path="*" element={<Navigate to="/" replace />} />
```
brendanbyrne, line 102: 🔵 **Suggestion** — the catch-all is doing more load-bearing work than the comment suggests. Without it, `/session` (no `:id`) doesn't match `/session/:id` and renders an empty `<Routes>` — the `if (!id) return <Navigate />` guard in `Session.tsx` is **unreachable from a real browser URL** because the only route that mounts `Session` requires `:id`. So in production today: this catch-all is what fixes the deep-link-without-id bug the PR description names; the `Session.tsx` guard is a type-narrowing safety net (and the unit-test path).

Two small docs nits that would make this less fragile to a future hand:

- Reference `#190` (PR-F1) in the comment so the replacement is greppable: `// PR-F1 (#190) will replace this with a real 404/errorElement.`
- In `Session.tsx:18-22`, add half a sentence noting the catch-all preempts the redirect at runtime — so future-someone doesn't delete one without realizing the other depends on it.

Not a `react.md` § 11 violation — § 11 explicitly recommends exactly the `if (!id) return <Navigate to="/" replace />` pattern used here. This is a comment-clarity suggestion given the two-layer defense.

frontend/src/components/RaceSetupPicker.tsx
```diff
@@ -108,11 +108,16 @@ export function RaceSetupPicker({
...
+  // button is disabled until `setup` is non-null, so `onComplete(setup)`
+  // never sees a partial value — the narrowing replaces four `!` assertions.
+  const setup =
     characterId !== null &&
     bodyId !== null &&
     wheelId !== null &&
-    gliderId !== null;
+    gliderId !== null
+      ? { characterId, bodyId, wheelId, gliderId }
+      : null;
```
brendanbyrne, line 120: 🔵 **Suggestion** — no test was added for `RaceSetupPicker`. `typescript.md` § 12 / `react.md` § 13 want a test when the change establishes a behavior that could silently break — here, "Confirm stays disabled until all four are picked; `onComplete` is invoked with the full setup once they are." Three sibling components (`BottomNav`, `DrinkTypeSelector`, `RunEntrySheet`) already have tests, so adding `RaceSetupPicker.test.tsx` matches the local pattern.

The behavior is structurally equivalent to the pre-PR `allSelected` flag (so not a 🟡), but the test would be ~20 lines and would catch a future refactor of the narrowing — e.g., the kind of refactor that wraps `setup` in `useMemo` and accidentally captures stale ids.

Sketch:

```tsx
it('keeps Confirm disabled until all four pieces are picked, then fires onComplete with the full setup', async () => {
  const user = userEvent.setup();
  const onComplete = vi.fn();
  // ...render with mocked useCharacters/useBodies/useWheels/useGliders...
  const confirm = screen.getByRole('button', { name: /confirm setup/i });
  expect(confirm).toBeDisabled();
  // pick one item per step
  // ...
  expect(confirm).toBeEnabled();
  await user.click(confirm);
  expect(onComplete).toHaveBeenCalledWith({ characterId: 1, bodyId: 1, wheelId: 1, gliderId: 1 });
});
```

brendanbyrne (commented):
## Rereview (commit `b232579`) — Approve

Both prior suggestions resolved cleanly. No new findings.

### What changed

- **`App.tsx` catch-all comment** — now explains the catch-all is load-bearing for the deep-link-without-id case and references PR-F1 (#190).
- **`Session.tsx` guard comment** — notes the catch-all preempts the redirect at runtime, plus *"If you remove one, look at the other."* That last line is a nicer breadcrumb than the one I sketched — it preempts the silent-deletion failure mode without needing a future reader to reconstruct the two-layer defense from scratch.
- **`RaceSetupPicker.test.tsx`** (new) — pins `Confirm disabled until all four pieces are picked → enabled → onComplete fires with the full setup`. Matches `react.md` § 13 patterns: MSW at the fetch boundary, `getByRole` queries (accessible name resolves via the inner ``), `userEvent.setup()`, `QueryClientProvider` wrapper, no internals introspection. The "one item per list" trick is well-justified inline given the 150 ms auto-advance — keeps the test deterministic without `vi.useFakeTimers()`.

### Checks (still clean)

- `bun run typecheck` — clean.
- `bun run lint` — 0 errors, 126 pre-existing warn-down warnings.
- `bun run test` — 180/180 (up from 179; the new `RaceSetupPicker.test.tsx` is the +1).

Ready for merge from a code-review-skill perspective; merge gate stays with Brendan.

PR #210, opened 2026-05-26
179: remove `as` cast from RunEntrySheet onError handler (PR-D3)

brendanbyrne:
## Summary

- Replace `(e.target as HTMLImageElement).style.display = 'none'` in `RunEntrySheet.tsx:234` with `e.currentTarget.style.display = 'none'`. React types `currentTarget` to the element that owns the handler, so the cast was asserting a type the compiler can already prove — pure type-safety cleanup, identical runtime behavior.
- Compliance plan: flip PR-D2 sign-offs to `[x]` (section + summary table row) and add the **Merged PR** link to [#208](https://github.com/brendanbyrne/beerio-kart/pull/208), catching the plan up to current state.

This is PR-D3 in the [frontend compliance plan](../blob/main/docs/designs/2026-05-16-frontend-compliance-plan.md). Closes #179.

## Scope audit (typescript.md § 2, § 8)

Before making the fix I swept the codebase for `as` casts and annotation-only validations. The original PR-D3 AC named two sites — only one still applies:

- ✅ **`RunEntrySheet.tsx:234`** — fixed in this PR (line number drifted from `:176`; updated the plan to match).
- ⏭️ **`DrinkTypeSelector.tsx:43`** — the `const created: DrinkType = await res.json()` anti-pattern was already replaced with `parseBody(DrinkTypeSchema, res)` when PR-B2 ([#199](https://github.com/brendanbyrne/beerio-kart/pull/199)) landed Zod schemas. No D3 work needed at that site. Noted in the plan with a strikethrough.

Remaining `as` uses in the codebase are intentional:

- `api/brand.ts` — branded-type mint pattern (`typescript.md` § 2 carve-out: legitimate `as` at the type boundary).
- `api/result.test.ts` — narrowing a known shape inside a test, not at a runtime boundary.
- `api/result.ts:88` — `(API_ERROR_CODES as readonly string[]).includes(code)` widens a `readonly` literal tuple so `Array.includes(value: string)` accepts the wider type (§ 2's "tell the compiler something it can't verify"). Already documented inline at the call site.

## Test plan

This is a pure type-safety refactor with no behavior change — no new test added (per `frontend/CLAUDE.md` § Testing: "if the PR can't name a user-visible behavior that would silently break, the test is theater"). Verification:

- [x] `bun run --cwd frontend lint` — 0 errors, no new warnings on the modified line
- [x] `bun run --cwd frontend typecheck` — clean
- [x] `bun run --cwd frontend test` — 180/180 passing
[...]

PR #211, opened 2026-05-27
182: migrate forms to useActionState + useFormStatus (PR-E1)

### LINKED ISSUE #179: Remove `as` casts and annotation-only validation from the frontend
## Context

Ninth PR. Removes the two unchecked-cast sites the audit found, plus an audit for any new ones that crept in after PR-B1.

## Acceptance criteria

- [ ] `RunEntrySheet.tsx:176` — replace `(e.target as HTMLImageElement)` with `e.currentTarget`. Safe here specifically because the handler is attached directly to the `` and `target` / `currentTarget` are the same in a non-bubbling `onError`. Do not apply this substitution blindly in delegated handlers.
- [ ] `DrinkTypeSelector.tsx:43` — the `const created: DrinkType = await res.json()` line is annotation-as-validation. Replace with `DrinkTypeSchema.parse(...)` using the Zod schema from PR-B2.
- [ ] Audit for any other `as` casts or annotation-only validation introduced after PR-B1, including `e.target as ...` in event handlers more generally.

## References

- Compliance plan: PR-D3
- Standards: `docs/coding-standards/typescript.md` § 2, § 8

### LINKED ISSUE #182: Migrate forms to React 19 useActionState + useFormStatus

brendanbyrne:
## Summary

PR-E1 from the [frontend compliance plan](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/designs/2026-05-16-frontend-compliance-plan.md). Five forms move from controlled inputs + manual `submitting` flags to React 19's `useActionState` + uncontrolled `FormData` + a shared `<SubmitButton>` that reads `useFormStatus().pending`. Native HTML validation (`required`, `minLength`, `maxLength`) is the first line of validation; the existing backend errors handle the rest.

- New: `SubmitButton` (`frontend/src/components/SubmitButton.tsx`) + test
- New on `useAuth`: `changePassword(currentPassword, newPassword)` returning `Promise<string | null>` — mirrors `login`/`register`
- Migrated: `Login`, `Register`, `Profile` password change form (factored into `PasswordChangeForm` sub-component), `DrinkTypeSelector` add-drink form (factored into `AddDrinkTypeForm` sub-component), `Onboarding` two-phase save
- Tests: `Login` (updated for transition timing), `Register` (new), `Profile` (new), `Onboarding` (new happy-path coverage). `DrinkTypeSelector`'s existing tests pass without changes — they already drove the component through the MSW boundary.
- Compliance bookkeeping: D3 sign-off catch-up (PR [#210](https://github.com/brendanbyrne/beerio-kart/pull/210) merged) folded into the first commit.

Closes #182. Closes #179 (D3 sign-off bookkeeping).

## Design notes

- **Sub-component extraction for fresh `useActionState`.** `useActionState` state persists across re-renders of the hosting component. For the Profile password form and the DrinkTypeSelector add-drink form, this would let a stale "Password changed!" or last-error message bleed from one open of the card into the next. Factoring each into a sub-component mounted/unmounted by the toggle gives a fresh hook state on every open. The Login/Register pages don't need this — they're the entire view, not a card.
- **Onboarding dispatch.** `useActionState`'s `dispatch` function takes the action's payload directly (`(payload) => void`). It is passed as the `onComplete` / `onSelect` callback to `RaceSetupPicker` / `DrinkTypeSelector` — no `<form action={…}>` because the pickers aren't HTML forms.
- **The auto-close timer.** The Profile password form closes itself 1500 ms after success. The `setTimeout` body calls the parent's `onDone` (which calls `setEditMode(null)`). That setState lives in the timer callback, not synchronously in the effect body, so the `react-hooks` "setState in effect" rule (project_setstate_in_effect_is_error) does not fire.
- **Lint warnings.** This PR adds a handful of `@typescript-eslint/no-confusing-void-expression` / `no-misused-promises` warnings to the existing pile. They're the same warn-level rules PR-A1 introduced (`docs/designs/.../frontend-compliance-plan.md` § PR-A1, warn-down principle) and PR-H1 will flip back to error.

## Test plan

Automated:

```bash
cd frontend
bun run typecheck   # passes
bun run lint        # 0 errors, 111 warnings (warn-down rules; pre-existing pattern)
bun run test        # 24 files, 192 tests passing
bun run build       # succeeds
```

### Manual browser smoke test (Brendan)

Start the dev server (`bun run dev` from repo root) and exercise each form. Focus on parity with main — none of the visible flows should have changed.

1. **Login** (`/login`)
   - Happy path: enter valid creds → redirected to `/`. Button shows "Logging in..." while pending.
   - Error path: enter wrong password → red error message appears, form stays open, button re-enabled.
   - Empty submit: leave fields blank → click "Log In" → browser blocks submit with native "Please fill out this field" tooltip.

2. **Register** (`/register`)
   - Happy path: pick a new username + ≥8-char password → redirected to `/onboarding`. Button shows "Creating account..." while pending.
   - Error path: re-use an existing username → red "Username already taken" (or similar) error.
   - Empty submit: blank → blocked by native validation.
   - Short password: 5-char password → browser shows "Please lengthen this text to 8 characters or more".

3. **Profile → Change Password** (`/profile`)
   - Happy path: current = correct, new = ≥8 chars → "Password changed!" green message appears, form auto-closes ~1.5s later.
   - Error path: current = wrong → "Current password is incorrect" red error, form stays open.
   - Empty submit: blank → native validation blocks.
   - Short new password: 5 chars → native validation blocks.
   - Cancel: click Cancel → form closes immediately, no API call.
   - Re-open after success: click Change again after the auto-close → form is fresh, no leftover "Password changed!" / error message.

4. **DrinkTypeSelector → Add new** (visible on `/profile` → Preferred Drink → "+ Not listed? Add new", and on `/onboarding` second phase)
   - Happy path: type a name (e.g., "Cider"), toggle alcoholic, click Add → the new type is selected. Button shows "Adding..." while pending.
   - Duplicate name: add a name that already exists → the backend dedups by name and returns the existing row with 200 (no error — see `api-contract.md` § 1.4), so the existing drink is simply selected. This is correct, idempotent behavior. (An earlier draft of this checklist wrongly described a "Drink already exists" error here; the backend never returns one.)
   - Known bug, tracked separately (not PR-E1): a name that exists with a *different* alcoholic flag (e.g. add non-alcoholic "Punch" when alcoholic "Punch" exists) silently returns the existing alcoholic row, discarding your flag. Identity is currently name-only; fix is to make it (name, alcoholic). Tracked in [#212](https://github.com/brendanbyrne/beerio-kart/issues/212).
   - Empty name submit: blank → native validation blocks.
   - Cancel: click Cancel → form closes, no API call.
   - Re-open after success: click "+ Add new" again → form is fresh.

5. **Onboarding** (`/onboarding` — visible right after Register)
   - Happy path: complete the race-setup picker → advances to drink picker → select a drink → redirected to `/`.
   - Error path on race-setup: temporarily break the PUT (e.g., disable network for the request) → red error shown, stays on race-setup phase.
   - Error path on drink-type: succeed on race-setup, then break the second PUT → red error, stays on drink phase.


frontend/src/components/DrinkTypeSelector.tsx
```diff
@@ -163,3 +95,100 @@ export function DrinkTypeSelector({
...
+            className={`w-11 h-6 flex-shrink-0 rounded-full transition-colors relative ${
+              alcoholic ? 'bg-blue-500' : 'bg-gray-300'
+            }`}
+            aria-label="Toggle alcoholic"
+            aria-pressed={alcoholic}
+          >
+            <span
+              className={`block absolute top-0.5 left-0.5 w-5 h-5 bg-white rounded-full shadow transition-transform ${
+                alcoholic ? 'translate-x-5' : 'translate-x-0'
+              }`}
```
brendanbyrne: 🔵 **Suggestion** — template-literal className concat with a ternary violates `tailwind.md` § 4 ("Use `clsx` ... do not string-concatenate or use nested template literals"). The toggle JSX was carried into the new `AddDrinkTypeForm` essentially verbatim, but this is the cheapest moment to fix it because the same PR added `SubmitButton.tsx` which already imports `clsx` correctly.

```tsx
className={clsx(
  'w-11 h-6 flex-shrink-0 rounded-full transition-colors relative',
  alcoholic ? 'bg-blue-500' : 'bg-gray-300',
)}
```

Same swap on lines 168-170 for the inner span's `translate-x-*`. If you're saving `clsx` for a dedicated sweep PR, OK — but please confirm so a future reviewer knows it's deferred.

frontend/src/pages/Profile.tsx
```diff
@@ -368,3 +285,85 @@ export function Profile() {
     </div>
   );
 }
+
+type PasswordState =
+  | { status: 'idle' }
+  | { status: 'error'; error: string }
+  | { status: 'success' };
```
brendanbyrne: 🔵 **Suggestion** — `typescript.md` § 4 names `kind` as the project's discriminator ("Use `kind` as the discriminator name to match the Rust backend's serde tag convention"). `PasswordState` is purely local state so the wire-format rationale doesn't strictly apply, but it's the only discriminated union in this PR and it picks a different tag from the rule.

Options:
- rename `status` → `kind` (one keystroke), or
- carve out "local-state DUs may use `status`" in `typescript.md` § 4 so future readers don't trip on it.

Noting either way.

frontend/src/pages/Profile.tsx
```diff
@@ -368,3 +285,85 @@ export function Profile() {
...
+  // the timer callback, not synchronously in the effect body, so the
+  // "setState in effect" lint rule (project_setstate_in_effect_is_error)
+  // does not apply.
+  useEffect(() => {
+    if (state.status !== 'success') return;
+    const timer = setTimeout(onDone, 1500);
+    return () => {
+      clearTimeout(timer);
+    };
+  }, [state.status, onDone]);
```
brendanbyrne: 🔵 **Suggestion** — timer-reset hazard: the parent passes `onDone={() => setEditMode(null)}` (a fresh function every Profile render), and this effect depends on `onDone`. If anything causes Profile to re-render during the 1500 ms success window — TanStack-Query refetch tick, context update, sibling-state change — the cleanup fires, the `setTimeout` is reset, and auto-close is pushed out by however long it took to re-render. Rare in practice today, fragile in principle.

Options:
- `useCallback` the `onDone` in `Profile.tsx:263` (`onDone={useCallback(() => setEditMode(null), [])}`), or
- capture `onDone` in a ref inside `PasswordChangeForm` and drop it from the dep array, or
- `// eslint-disable-next-line react-hooks/exhaustive-deps -- timer should fire once on transition to success` with that reason.

The Compiler (PR-E2) would fix this for free, so deferring with a TODO referencing E2 is also fine if you'd rather not bikeshed now.

brendanbyrne (commented):
**Verdict:** Request Changes (one important deviation to resolve; rest is suggestion-level).

**What this PR does:** Migrates Login, Register, the Profile password form, the DrinkTypeSelector add-drink form, and the Onboarding two-phase save from controlled inputs + manual `submitting` flags to React 19's `useActionState` + `useFormStatus`. Adds a shared `SubmitButton`, extracts `PasswordChangeForm` and `AddDrinkTypeForm` as sub-components for fresh-state-per-open, and adds `useAuth.changePassword` mirroring `login`/`register`. Tests added across the migrated surface.

**Areas reviewed against:** `react.md` (§§ 1, 2, 5, 6, 8), `typescript.md` (§§ 2, 4, 6, 8, 12), `tailwind.md` (§§ 4, 5), `api-contract.md` (no new endpoints), `design.md` (no architecture change), the compliance plan PR-E1 scope checklist.

**Automated checks:**
- `bun run typecheck` — passes.
- `bun run lint` — 0 errors, 111 warnings (pre-existing warn-down pattern; PR description acknowledges).
- `bun run test` — 26 files, 201 tests passing.
- `bun run test:coverage` — 80.57% lines overall. New code on SubmitButton, the `readString` helper, `useAuth.changePassword`, and every new test file is well-covered; Profile.tsx's 50% reflects the pre-existing untested race-setup/drink-type save flows, not new debt from this PR.

---

## 🟡 Important

### 1. Compliance plan scope item not delivered: "Zod schema at submit time for each form"

`docs/designs/2026-05-16-frontend-compliance-plan.md` PR-E1 scope explicitly lists:

> Native validation (`required`, `minLength`, `maxLength`, `pattern`) on every input; **Zod schema at submit time for each form**.

`typescript.md` § 8 echoes this: *"Native HTML validation ... plus a **one-shot schema check at submit** is sufficient."*

The PR delivers the native-validation half but skips the Zod-at-submit half on every form. Pragmatically: for `Login`/`Register`/`PasswordChangeForm` the Zod schema would only re-check what `required`/`minLength` already enforce, and the backend is authoritative. For `AddDrinkTypeForm` there's a hand-rolled `if (!name.trim()) return { error: 'Name is required' }` standing in. The PR description argues *"Native HTML validation ... is the first line; the existing backend errors handle the rest."*

Either:
- **add the Zod schemas** to match the plan literally, or
- **amend the plan** with a sign-off note saying "Zod-at-submit dropped for E1; native + backend errors cover our current form complexity, revisit if forms add cross-field validation."

This is the kind of explicit deviation that should leave a trail so the next reviewer doesn't relitigate it.

## 🔵 Suggestions (see inline comments)

- `AddDrinkTypeForm` toggle keeps template-literal className concat (`tailwind.md` § 4).
- `PasswordState` uses `status` as the discriminator; `typescript.md` § 4 names `kind`.
- `useEffect` in `PasswordChangeForm` depends on a fresh-each-render `onDone` (timer-reset hazard).

## What's good

- **Sub-component extraction for fresh `useActionState`** is exactly right; the inline comments explaining *why* are the kind of "hidden constraint" comments the standards explicitly reward.
- `PasswordState` as a **proper discriminated union** prevents the "leftover error showing alongside success" bug class that flat flag pairs always have.
- **Real a11y improvements** on the alcoholic toggle (`aria-label="Toggle alcoholic"` + `aria-pressed={alcoholic}`) — these were missing pre-PR.
- `SubmitButton` is clean: single responsibility, uses `clsx`, accepts `pendingLabel`, has a focused unit test that documents the out-of-form contract.
- `changePassword` mirrors `login`/`register`'s `Promise<string | null>` signature — predictable from one call site.
- The **hidden-input pattern** for the `alcoholic` boolean is the canonical React-19-uncontrolled idiom for threading non-text state through `FormData`.
- `readString` helper + its test triad (string / missing / File) matches `typescript.md` § 12's parser-test guidance.
- **Test coverage on new code is solid** — happy + error paths through MSW for `useAuth`, both phases of Onboarding, both branches of the password change with timer-aware fake timers.

## Design-doc accuracy: no changes required

No `design.md` / `api-contract.md` / `data-model.md` updates needed: form pattern is documented in `react.md` § 8; no new endpoints; no schema changes. Compliance plan was updated for PR-D3 sign-off (link to merged PR #210) — bookkeeping correct.

brendanbyrne:
Pushed `65a49e9` + `3d9802a`. Summary of how each item was addressed:

## 🟡 Blocker — Zod at submit

Added `safeParse(Object.fromEntries(formData))` to all four real forms:

- `Login.tsx` → `LoginFormSchema = z.object({ username: z.string().min(1), password: z.string().min(1) })`
- `Register.tsx` → `RegisterFormSchema = z.object({ username: z.string().min(1), password: z.string().min(8) })`
- `Profile.tsx` `PasswordChangeForm` → `PasswordFormSchema = z.object({ current_password: z.string().min(1), new_password: z.string().min(8) })`
- `DrinkTypeSelector.tsx` `AddDrinkTypeForm` → `AddDrinkTypeFormSchema = z.object({ name: z.string().trim().min(1).max(60), alcoholic: z.enum(['true', 'false']).transform((v) => v === 'true') })` — the schema's transform also kills the hidden-input string-boolean conversion that previously lived inline.

The `readString` helper from the prior commit is now retired — the schemas absorb the `FormData.get(...)` read plus the type narrowing in one step.

Each form gets a new "Zod backstop" test that exercises the schema-failure branch by either stripping `required` (Login, AddDrinkType) or relying on jsdom's known no-op on `minLength` (Register, Profile password). Patch coverage **75% → 87.5%**.

**Onboarding amended in the plan, not the code.** `Onboarding.tsx` doesn't render a `<form>` — both phases delegate to `RaceSetupPicker` / `DrinkTypeSelector`, which dispatch already-typed `Setup` / `DrinkType` payloads to `useActionState` via callback. There's no FormData parse step where a submit-time schema fits. `docs/designs/2026-05-16-frontend-compliance-plan.md` PR-E1 scope now spells this out + a dated history entry so the next reviewer doesn't relitigate.

## 🔵 Suggestions

1. **`status` → `kind`.** Renamed in `PasswordState` and its three call sites. The tests don't reference the field directly (they assert on DOM), so they passed without changes.

2. **`clsx` on the toggle.** Two template-literal class concats in `AddDrinkTypeForm` swapped to `clsx(...)`. `clsx` was already imported by `SubmitButton`; one more import here.

3. **Timer-reset hazard.** Took option 1 — `closePasswordForm = useCallback(() => setEditMode(null), [])` in `Profile.tsx`, passed as `onDone` to `PasswordChangeForm`. Inline comment notes this becomes redundant once PR-E2 ships the Compiler. Considered option 4 (defer to E2) but agreed with the principle of not shipping a known fragility on the strength of "the next PR will fix it for free."

## Other

`3d9802a` repairs `docs/project-workflow.md` § Pre-push checks — its "unreachable-defensive-guard" example pointed at `forms.ts` which this PR deletes. Updated to point at `Login.tsx`'s `safeParse` pattern instead, which is the actually-shipped recommendation.

All four checks green (`Coverage`, `codecov/patch/{backend,frontend}`, `link-check`). Patch coverage 87.5%. 202 tests passing locally.

brendanbyrne (commented):
## Re-review — all four findings addressed cleanly

**Verdict:** Approve-ready (no remaining standards or design issues from my read). Holding off `APPROVE` per the repo convention that Brendan is the sole approver.

**What changed since `a6cb4ff`:**
- `65a49e9` — addresses the four findings (Zod-at-submit, `kind` rename, `clsx`, `useCallback`).
- `d0f8a70` / `3d9802a` — bonus: documents the pre-push coverage check + the two recurring patch-gate failure modes in `project-workflow.md`. Tangential to PR-E1 strictly but directly traceable to the codecov-gate friction Brendan hit on #210 and #211, so reasonable to bundle.

### Important finding 1 (Zod-at-submit) — resolved

All four real forms now have a `Zod.safeParse(Object.fromEntries(formData))` at the top of the action:

- `Login.tsx:15-18` — `LoginFormSchema` (`username.min(1)`, `password.min(1)`).
- `Register.tsx:14-17` — `RegisterFormSchema` (`username.min(1)`, `password.min(8)`).
- `Profile.tsx:306-309` — `PasswordFormSchema` (`current_password.min(1)`, `new_password.min(8)`).
- `DrinkTypeSelector.tsx:106-109` — `AddDrinkTypeFormSchema` (`name.trim().min(1).max(60)` and an `alcoholic` enum-transform that converts the hidden-input `'true' | 'false'` string into a real boolean before the body is built — that's slicker than the previous hand-rolled `=== 'true'`).

The `Onboarding` exemption is documented inline in the compliance plan PR-E1 scope (the picker children dispatch already-typed payloads via callback, so there's no `FormData` parse step where a submit-time Zod schema would fit). That's the right call and the right place for it.

**Backstop tests** prove the schemas actually fire — not just present, but exercised:
- `Register.test.tsx:72-88` strips the input's `minLength` (jsdom doesn't enforce it) and asserts the Zod path returns "Password must be at least 8 characters."
- `DrinkTypeSelector.test.tsx:86-100` removes the `required` attribute and asserts "Name is required."
- `Login.test.tsx:71` and `Profile.test.tsx:133` have parallel backstop tests.

This is exactly the right test shape for a "second line of defense" — without it the schemas would be there but unverified.

Bonus: the `readString` helper is gone (no callers after Zod takes over the read+narrow in one step), and `forms.test.ts` deleted alongside. Cleaner.

### Suggestion 2 (clsx for AddDrinkTypeForm toggle) — resolved

`DrinkTypeSelector.tsx:171-175` and `:179-184` — both `className`s on the toggle button and inner span swapped to `clsx`. Matches the pattern `SubmitButton.tsx` already uses in this PR.

### Suggestion 3 (`kind` discriminator) — resolved

`Profile.tsx:296-299` — `PasswordState` renamed from `status` → `kind` on the type, the `PASSWORD_INITIAL` constant, all three return-shape literals in the action, and all `state.kind === ...` reads in the JSX. Matches `typescript.md` § 4.

### Suggestion 4 (timer-reset hazard from fresh-each-render `onDone`) — resolved

`Profile.tsx:38-40` — `closePasswordForm = useCallback(() => setEditMode(null), [])`, passed at `:270` as `onDone={closePasswordForm}`. The inline comment names PR-E2 (Compiler) as the eventual cleanup so a future reader knows the `useCallback` is bridge-code, not a permanent fixture.

### Automated checks

- `bun run typecheck` — passes.
- `bun run lint` — 0 errors, **110** warnings (down from 111; the removed `readString` helper dropped one warning).
- `bun run test` — 25 files, **202** tests passing (+1 for the new DrinkTypeSelector tests).
- `bun run test:coverage` — 80.76% lines overall (up from 80.57%). DrinkTypeSelector.tsx 90.32% (up from 86.66%); Profile.tsx 52.11% (up from 50%, with the 199-235 gap being the still-untested race-setup/drink-type save flows that are out of PR-E1 scope by design).

### Bonus material: `project-workflow.md` § Pre-push checks

The addition is well-targeted. The "unreachable-defensive-guard blind spot" example specifically uses the `Object.fromEntries + Zod.safeParse` pattern as the recommended fix — which is the same pattern this PR adopts to eliminate `readString`'s now-dead `typeof value === 'string'` branch. That symmetry is nice and makes the doc concrete rather than abstract.

---

Nothing else to flag. Ready for Brendan's merge call.

brendanbyrne (commented):
## Re-review — `c9ba6f5` (test-fixture correctness)

**Verdict:** Still approve-ready. This commit improves test fidelity; no production code changed, no new findings.

**What changed:** One test in `DrinkTypeSelector.test.tsx` — the add-drink failure case — was relabeled from "shows the backend error message when adding fails" (with a `409 { error: 'Drink already exists' }` fixture) to "surfaces a backend error when the create request fails" (with a `500 { error: 'Something went wrong', code: 'internal' }` fixture), plus an explanatory comment.

**Why this is the right fix:** the old fixture was fictional on two counts, and I verified both against the source of truth:

1. **The backend never returns 409 for a duplicate.** `backend/src/routes/drink_types.rs:94-101` derives a deterministic UUID from the uppercased name and, on collision, returns the existing row with **200** — never a 409. The test's comment citing `api-contract.md § 1.4` ("dedups by name … returns the existing row with 200") is accurate against both the handler doc-comment (`drink_types.rs:72-76`) and the contract line ("returns existing on UUID collision").
2. **The old fixture omitted the `code` field** that the real envelope always carries (`api-contract.md § 3`: `{ error, code }`).

The new fixture is contract-accurate end to end:
- `500` is the right status for a generic server failure.
- `code: 'internal'` is a real registry value — `api-contract.md § 7` line 241 (`500 | internal`) and `error.rs` (`Self::Internal(_) => ErrorCode::Internal`, snake_case `"internal"`).
- It exercises the real parse path: `parseApiError` (`result.ts:96-125`) validates `{ error: z.string(), code: z.string().optional() }`, runs `code` through `isApiErrorCode` against the registry (so `'internal'` resolves to a recognized code, not the `unknown` fallback), and surfaces `.error` as the message — which is what the assertion reads.

This is exactly the "don't let a green MSW test prove only that you handle a fixture you invented — check it against the real status codes and error shapes" discipline. The previous version passed by construction against an impossible scenario; this one passes against a scenario the backend can actually produce.

Good call splitting the genuinely-distinct (name, different-`alcoholic`) dedup bug this surfaced out to #212 rather than smuggling a behavior fix into a forms-migration PR.

**Checks:** `bun run test src/components/DrinkTypeSelector.test.tsx` — 3 passed. (Test-only, single-file change; the full suite ran green at the prior head `3d9802a`.)

Nothing else to flag. Ready for Brendan's merge call.

PR #213, opened 2026-05-28
212: identify drink types by (name, alcoholic), not name alone

### LINKED ISSUE #212: Drink-type identity is name-only; same name + different alcoholic flag collide
## Context

`POST /api/v1/drink-types` derives each row's primary key as `uuid_v5(NAMESPACE, UPPERCASE(name))` — from the **name only** ([`backend/src/drink_type_id.rs:16`](https://github.com/brendanbyrne/beerio-kart/blob/main/backend/src/drink_type_id.rs#L16)). The `alcoholic` flag is not part of the identity, and the DB has `UNIQUE(name)` ([migration `m20260101_000001_initial_schema.rs:177`](https://github.com/brendanbyrne/beerio-kart/blob/main/backend/migration/src/m20260101_000001_initial_schema.rs#L177)).

On a name collision the handler returns the **existing row with 200** and silently ignores the submitted `alcoholic` flag ([`routes/drink_types.rs:98`](https://github.com/brendanbyrne/beerio-kart/blob/main/backend/src/routes/drink_types.rs#L98)). The dedup-and-return-existing behavior is intentional and documented (`api-contract.md` § 1.4).

## The bug

Adding a drink whose name already exists **with a different alcoholic flag** silently returns the existing row, discarding the caller's flag. Example: alcoholic "Punch" exists → user adds non-alcoholic "Punch" → they get the alcoholic "Punch", with no error and no indication their choice was dropped. Real drinks legitimately exist both ways (Punch, Cider, Ginger Beer, etc.).

Surfaced during PR [#211](https://github.com/brendanbyrne/beerio-kart/pull/211) (PR-E1) manual smoke testing. The *exact*-duplicate case (same name, same alcoholic → select existing) is correct and stays; only the differing-flag case is a bug.

## Decision

Make drink-type identity **`(name, alcoholic)`** rather than name-only (chosen over a 409-reject because rejecting leaves no way to add a legitimately-distinct non-alcoholic variant). Blast radius is small — all IDs are derived (no hardcoded UUIDs), the frontend treats IDs as opaque strings, and we're prelaunch (DB reset, no data migration).

## Acceptance criteria

- [ ] `drink_type_uuid` derives from `(UPPERCASE(name), alcoholic)` — distinct UUIDs for the alcoholic and non-alcoholic forms of the same name. Update its unit tests.
- [ ] `POST /drink-types` keys dedup on `(name, alcoholic)`; the create handler passes `req.alcoholic` into the UUID derivation. Exact-duplicate (same name + flag) still returns the existing row with 200; same name + different flag creates a distinct row.
- [ ] DB constraint changes from `UNIQUE(name)` to `UNIQUE(name, alcoholic)` (edit the consolidated migration in place per backend prelaunch convention; reset dev DB).
- [ ] Call sites updated to pass the flag: `seed.rs`, `test_helpers.rs`, `tests/api_integration.rs` seed loop.
- [ ] Case-insensitive name dedup preserved ("Beer"/"BEER" with the same flag still collide).
- [ ] Integration test: same name + different alcoholic → two distinct rows, each retrievable.
- [ ] `api-contract.md` § 1.4 wording updated to describe `(name, alcoholic)` dedup.

## References

- PR-E1: [#211](https://github.com/brendanbyrne/beerio-kart/pull/211) (where this surfaced)
- `api-contract.md` § 1.4

brendanbyrne:
## Summary

A drink type's deterministic UUID was derived from the uppercased **name only**, so the alcoholic and non-alcoholic forms of the same name collided. `POST /drink-types` dedup-returned the existing row and silently discarded the submitted `alcoholic` flag — adding non-alcoholic "Punch" when alcoholic "Punch" existed gave you the alcoholic one, no error. Real drinks (Punch, Cider, Ginger Beer) legitimately exist both ways.

This makes `(name, alcoholic)` the identity.

## Linked work

Closes #212. Surfaced during PR-E1 ([#211](https://github.com/brendanbyrne/beerio-kart/pull/211)) manual smoke testing.

## Changes

- `drink_type_uuid(name, alcoholic)` now hashes `UPPER(name) \x1f alcoholic`, giving the two forms distinct IDs; name matching stays case-insensitive.
- `POST /drink-types` passes `req.alcoholic` into the derivation. Exact `(name, alcoholic)` re-submit still returns the existing row with `200` (idempotent); same name + different flag creates a distinct row.
- DB `UNIQUE(name)` → composite `UNIQUE(name, alcoholic)` (backstop; the case-insensitive app-level dedup via the derived PK is the primary guard).
- Every `drink_type_uuid` call site updated to pass the flag (seed, test helpers, three service test modules, two integration suites).
- `api-contract.md` § 1.4 updated to document the new dedup semantics.

## Tests

- Unit: `drink_type_uuid("Punch", true) != drink_type_uuid("Punch", false)`; case-insensitivity preserved.
- Integration: `test_create_drink_type_same_name_different_alcoholic_are_distinct` — both Punch variants POST to distinct IDs, both persist and list; existing case-insensitive dedup test still green.
- Full backend suite: 298 unit + integration tests pass. `cargo clippy --all-targets` and `cargo +nightly fmt --check` clean.

## How to verify

```bash
# From repo root
cargo test --test api_integration drink_type   # 10 pass, incl. the new distinct-variants test
cargo test                                      # full suite green
```

Manual (optional):
```bash
# Reset dev DB first — schema changed (new composite unique index).
# Stop the dev server, delete the local SQLite file, restart.
# Then: add "Punch" alcoholic, then "Punch" non-alcoholic → two rows, both selectable.
```

⚠️ **Dev DB reset required** (prelaunch convention): the consolidated migration was edited in place to add the composite unique index. Delete the local SQLite file and restart so SeaORM recreates the schema.


backend/migration/src/m20260101_000001_initial_schema.rs
```diff
@@ -174,18 +174,29 @@ impl MigrationTrait for Migration {
...
             .await?;
 
+        // A drink type is identified by (name, alcoholic): the alcoholic and
+        // non-alcoholic forms of the same name are distinct drinks, so the
+        // uniqueness is composite rather than name-alone. (App-level dedup is
+        // case-insensitive via the derived PK; this index is the DB backstop.)
+        manager
+            .create_index(
+                Index::create()
+                    .name("idx_drink_types_name_alcoholic")
```
brendanbyrne, line 191: 🔵 **Suggestion** (no change required, just FYI) — this composite `UNIQUE(name, alcoholic)` index is weaker than it reads. Because the PK `id` is a pure function of `(UPPER(name), alcoholic)`, any insert that would trip this index already trips the PK first — and for case-variants ("Punch" vs "punch", same flag) this index *doesn't* fire (SQLite's default `BINARY` collation is case-sensitive) while the PK does. So it never catches anything the PK doesn't, and it doesn't enforce the case-insensitive half of the identity. Its real value is defense-in-depth against a future code path that builds an `id` *without* `drink_type_uuid` — a legitimate backstop, exactly as the comment above says. Flagging only so nobody later mistakes it for the case-insensitive guard. Index name matches the existing `idx_tracks_cup_position` convention — good.

brendanbyrne (commented):
## Code review — `event: COMMENT` (verdict is Brendan's call)

**Suggested verdict:** Request Changes — one small doc-accuracy fix before merge. The code itself is solid.

**What this PR does:** Derives a drink type's deterministic UUID from `(UPPER(name), alcoholic)` instead of name alone, so the alcoholic and non-alcoholic forms of the same drink (Punch, Cider, Ginger Beer) stop colliding. `UNIQUE(name)` → composite `UNIQUE(name, alcoholic)`; the flag is threaded through every call site.

**Standards consulted:** `seaorm.md` (§ 5 migrations, § 6 entities), `rust.md` (doc comments), `api-contract.md` § 1.4, `data-model.md`.

**Automated checks (run locally from repo root):**
- `cargo +nightly fmt --check` — clean
- `cargo test` — 379 passing, 0 failed (incl. the new distinct-variants integration test and the `drink_type_id` unit test)

---

### 🟡 Important — `docs/data-model.md` § Drink Types is now stale

`api-contract.md` § 1.4 was correctly updated, but its sibling schema doc wasn't, and now contains two factually-wrong statements ([data-model.md:124-133](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/data-model.md#L124-L133)):

- L124: `id: UUID (... deterministic via uuid_v5 of uppercased name)` → now `(uppercased name, alcoholic)`.
- L125: `name: TEXT (unique, ...)` → `name` is no longer standalone-unique; the constraint is composite `UNIQUE(name, alcoholic)`.
- L132 note: `uuid_v5(DRINK_TYPE_NAMESPACE, uppercase(name))` → stale derivation.
- L133 note: the dedup-on-collision description needs the alcoholic-flag nuance.

`data-model.md` maintains a `## Document history` section, so per `docs/CLAUDE.md` it's in scope and also needs a dated entry. Suggested replacement:

```
drink_types
├── id: UUID (primary key, deterministic via uuid_v5 of (uppercased name, alcoholic))
├── name: TEXT (not null, stored as-entered by first creator)
├── alcoholic: BOOLEAN (not null)
├── created_by: UUID (foreign key -> users, nullable — null for pre-seeded entries)
└── created_at: TIMESTAMP (not null)
```
Notes:
- UUID derived from `uuid_v5(DRINK_TYPE_NAMESPACE, "{uppercase(name)}\x1f{alcoholic}")`. Name matched case-insensitively; the `alcoholic` flag is part of the identity, so the two forms of the same name (e.g. "Punch") are distinct drinks with distinct IDs.
- A composite `UNIQUE(name, alcoholic)` index backstops the app-level dedup at the DB layer (the derived PK is the primary guard).
- If a user submits a drink whose `(name, alcoholic)` already exists (any casing of the name), the app returns the existing entry. The same name with the *other* flag creates a distinct drink.

---

### Design-doc accuracy
- `api-contract.md` § 1.4 — updated, accurate, with a history entry. ✅
- `data-model.md` — needed updating, **not done** (finding above).
- Consolidated-migration edit-in-place + dev-DB-reset is correct per prelaunch policy (`seaorm.md` § 5); the PR body flags the reset. `down()` drops the whole table, so the new index needs no explicit `down` handling — correct.

### What's good
- The `\x1f` (Unit Separator) `FIELD_SEP` with the documented `("AB", true)` vs `("A", "Btrue")` collision rationale is exactly the right defensive instinct.
- All 30 call sites threaded through; verified nothing hardcodes the old derived IDs (frontend + backend grep clean), so the ID-derivation change is safe.
- The integration test asserts the *full* invariant — distinct IDs, flag honored not discarded, both rows persist and list — not just "IDs differ."

brendanbyrne:
Addressed the 🟡 finding in b66df31 — `data-model.md` § Drink Types now matches `api-contract.md` § 1.4:

- `id` derivation → `uuid_v5` of `(uppercased name, alcoholic)`
- `name` line no longer says `unique` (the constraint is composite `UNIQUE(name, alcoholic)`)
- UUID note updated to the `"{uppercase(name)}\x1f{alcoholic}"` hash input
- dedup note now distinguishes a same-`(name, alcoholic)` collision (returns existing) from the same name with the other flag (distinct drink)
- added the composite-index backstop note
- dated `## Document history` entry added

Used your suggested replacement near-verbatim. Ready for re-review.

Issue #184, opened 2026-05-18
Frontend accessibility sweep — touch targets, focus management, modal a11y


brendanbyrne:
## Context

Fourteenth PR. Brings every interactive element to ≥44 px, adds modal a11y (focus trap, Escape, `role="dialog"`, `aria-modal`), removes the `autoFocus` regression, and verifies `eslint-plugin-jsx-a11y` is clean. Consider native `<dialog>` for the modals — it gives focus trap and Escape for free.

## Acceptance criteria

- [ ] Bump every <44 px button to `min-h-touch` (or `min-h-[44px]`). Hit list from audit: `DrinkTypeSelector.tsx` skip/cancel/add, `RaceSetupPicker.tsx` step pills, `Profile.tsx` password cancel/save, `Home.tsx` modal cancel, `Login.tsx` and `Register.tsx` submit buttons (borderline at `py-2.5`).
- [ ] Remove `autoFocus` from `DrinkTypeSelector.tsx:99` (use a manual focus ref only if needed).
- [ ] Add `role="dialog"`, `aria-modal="true"`, focus trap, and Escape handling to `Home.tsx`'s create-session modal and `RunEntrySheet.tsx`'s bottom sheet.
- [ ] `eslint-plugin-jsx-a11y` recommended produces zero warnings.
- [ ] Add `@axe-core/react` in dev to catch runtime issues.
- [ ] Tab through every screen keyboard-only. Each modal: Escape closes; focus returns to trigger. axe shows zero violations on the main flows.

## References

- Compliance plan: PR-G2
- Standards: `docs/coding-standards/react.md` § 10; `docs/coding-standards/tailwind.md` § 5

_Filed by Cowork on Brendan's behalf._

PR #214, opened 2026-05-28
180: adopt React Compiler, Document Metadata, drop manual memoization

### LINKED ISSUE #180: Adopt React Compiler, ref-as-prop, and React 19 Document Metadata
## Context

Eleventh PR. Installs the React Compiler (v1.0 stable since October 2025) into the Vite config, converts any `forwardRef` (none expected) to ref-as-prop, and adopts React 19's native Document Metadata for per-page `<title>` / `<meta>`. Removes hand-written `useMemo`/`useCallback` where the Compiler can handle them.

## Acceptance criteria

- [ ] Install `babel-plugin-react-compiler`; enable in `vite.config.ts` (dev + prod).
- [ ] Verify `eslint-plugin-react-hooks` recommended preset includes Compiler-specific rules.
- [ ] Audit any `forwardRef` usage and convert to ref-as-prop. (None expected — audit found none.)
- [ ] Add per-page `<title>` via React 19's native Document Metadata where appropriate.
- [ ] Remove now-redundant `useMemo` / `useCallback` calls where the Compiler handles them. Keep them only at sites with a measured reason (see `react.md` § 7 carve-outs).
- [ ] Build succeeds; dev mode works; no Compiler errors in console.
- [ ] Profile a representative page (Session.tsx with active polling) before and after — re-render count equal or lower.

## References

- Compliance plan: PR-E2
- Standards: `docs/coding-standards/react.md` § 2, § 5, § 7

brendanbyrne:
## Summary

PR-E2 of the [frontend compliance plan](docs/designs/2026-05-16-frontend-compliance-plan.md). Turns the React Compiler on, adopts React 19 Document Metadata for per-page titles, and removes hand-written memoization the Compiler now handles.

Closes #180.

## What changed

- **React Compiler enabled** ([vite.config.ts](frontend/vite.config.ts)). `@vitejs/plugin-react@6` (Vite 8 / oxc) no longer exposes a `babel` option, so the Compiler is wired in the documented v6 way: `@rolldown/plugin-babel` + the plugin's `reactCompilerPreset()` helper (it carries a preconfigured filter for React/hook files and targets React 19 by default). Active in dev and prod.
- **eslint:** no config change. `eslint-plugin-react-hooks@7`'s `flat.recommended` (added back in PR-A1) already ships the full Compiler rule suite at error level (`immutability`, `preserve-manual-memoization`, `purity`, `set-state-in-effect`, `use-memo`, …). Verified via `--print-config`.
- **Removed manual memoization** per `react.md` § 7: the two `useMemo` derivations + two `useCallback` drag handlers in `RunEntrySheet`, and the `refresh`/`closePasswordForm`/auth `useCallback`s in `useGameData`, `useUserProfile`, `Profile`, `useAuth`.
  - **One carve-out kept:** `useAuth`'s `clearAuth` stays `useCallback`. It's a dependency of the mount (silent-refresh) effect *and* is registered into module scope via `setOnAuthFailure`, so it must be referentially stable. `exhaustive-deps` (a static rule that can't see the Compiler) enforces this, and `react.md` § 6 forbids silencing it. Documented inline.
- **Document Metadata:** per-page `<title>` added to all six `pages/` components (Login, Register, Home, Profile, Session, and a phase-aware one for Onboarding). React 19 hoists these to `<head>`.
- **forwardRef:** none in the codebase — nothing to convert (AC confirmed).
- **Test:** added a slide-to-DQ interaction test covering the now-unwrapped drag handlers (previously zero coverage).

## Automated verification

- `react-compiler-healthcheck`: **29/29 components compiled**, no incompatible libraries, StrictMode detected.
- `bun run build` succeeds; dev transforms inject the Compiler runtime (`Session.tsx` → `_c(4)`, `useAuth.tsx` → `_c(1)` with the explicit `useCallback` preserved).
- Dev server starts clean — no Compiler errors/warnings in the console; key modules transform to 200.
- `bun run lint`: 0 errors. `bun run test`: **204 passing**. Diff (patch) coverage: **100%** of changed coverable lines.

## Manual test plan

Run `bun run dev` from the repo root, open the app on a phone-width viewport (Pixel 9 Pro reference), and check **both Chrome and Firefox**:

1. **Document titles** — watch the browser tab title as you navigate:
   - `/login` → "Log In · Beerio Kart"
   - `/register` → "Register · Beerio Kart"
   - Home → "Home · Beerio Kart"
   - Onboarding → "Pick Your Setup · Beerio Kart", then "Pick Your Drink · Beerio Kart" as the phase advances
   - Profile → "Profile · Beerio Kart"
   - A session → "Session · Beerio Kart"
   - Navigating back to Home resets the title (no stale title bleed).
2. **Auth still works** — log in, reload the page (silent refresh restores the session — confirm in the Network tab that `/auth/refresh` fires **once**, not in a loop), change password (success message → auto-closes after ~1.5s), log out.
3. **Run entry** — open a race's run sheet: defaults pre-fill, drink/setup overrides work, and the **slide-to-DQ** control: drag the thumb fully right → "Disqualified" appears; tap to undo; a short drag snaps back without DQ.
4. **Re-render profiling (the one step I couldn't automate)** — with React DevTools → Profiler, record an active Session page through a couple of polling ticks. Re-render count should be **equal or lower** than before this PR. This is expected by construction: the Compiler only *adds* memoization, and `useAuth`'s context value (previously a fresh object literal every render, forcing all consumers to re-render) is now memoized — so this should strictly *reduce* re-renders.

## Notes

- Bookkeeping: the plan still lists PR-E1 (#182) and PR-E2 (#180) sign-off boxes unchecked. Those are yours to tick on merge.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -25,6 +25,7 @@
     "zod": "^4.4.3"
   },
   "devDependencies": {
+    "@babel/core": "^7.29.7",
     "@eslint/js": "^9.39.4",
     "@rolldown/plugin-babel": "^0.2.3",
     "@tailwindcss/vite": "^4.2.2",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -26,6 +26,7 @@
   },
   "devDependencies": {
     "@eslint/js": "^9.39.4",
+    "@rolldown/plugin-babel": "^0.2.3",
     "@tailwindcss/vite": "^4.2.2",
     "@testing-library/jest-dom": "^6.9.1",
     "@testing-library/react": "^16.3.2",
@@ -35,6 +36,7 @@
     "@types/react-dom": "^19.2.3",
     "@vitejs/plugin-react": "^6.0.1",
     "@vitest/coverage-istanbul": "^4.1.6",
+    "babel-plugin-react-compiler": "^1.0.0",
     "eslint": "^9.39.4",
     "eslint-config-prettier": "^10.1.8",
     "eslint-plugin-import": "^2.32.0",
--- a/frontend/src/components/RunEntrySheet.test.tsx
+++ b/frontend/src/components/RunEntrySheet.test.tsx
@@ -1,5 +1,5 @@
 import { http, HttpResponse } from 'msw';
-import { describe, expect, it, vi } from 'vitest';
+import { afterEach, describe, expect, it, vi } from 'vitest';
 import { fireEvent, render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
@@ -230,6 +230,15 @@ describe('RunEntrySheet', () => {
     expect(img.style.display).toBe('none');
   });
 
+  // pinSliderGeometry pollutes HTMLElement.prototype globally, so cleanup must
+  // run even if an assertion throws mid-test — an end-of-body restore() would
+  // leak the mock into later tests. afterEach always drains it (react.md § 13).
+  let restoreGeometry: (() => void) | null = null;
+  afterEach(() => {
+    restoreGeometry?.();
+    restoreGeometry = null;
+  });
+
   // The slide-to-DQ control measures its track from `offsetWidth` (a layout
   // value jsdom reports as 0) and maps the pointer's clientX through
   // `getBoundingClientRect`. Pin both so the drag math is deterministic: a
@@ -253,7 +262,7 @@ describe('RunEntrySheet', () => {
         y: 0,
         toJSON: () => ({}),
       } as DOMRect);
-    return () => {
+    restoreGeometry = () => {
       rect.mockRestore();
       delete (HTMLElement.prototype as { offsetWidth?: number }).offsetWidth;
     };
@@ -267,7 +276,7 @@ describe('RunEntrySheet', () => {
         () => new HttpResponse(null, { status: 500 }),
       ),
     );
-    const restore = pinSliderGeometry();
+    pinSliderGeometry();
 
     render(
       <RunEntrySheet race={race} onClose={vi.fn()} onSubmitted={vi.fn()} />,
@@ -284,7 +293,6 @@ describe('RunEntrySheet', () => {
     fireEvent.mouseUp(track);
 
     expect(await screen.findByText('Disqualified')).toBeInTheDocument();
-    restore();
   });
 
   it('snaps back without disqualifying when released before the threshold', async () => {
@@ -295,7 +303,7 @@ describe('RunEntrySheet', () => {
         () => new HttpResponse(null, { status: 500 }),
       ),
     );
-    const restore = pinSliderGeometry();
+    pinSliderGeometry();
 
     render(
       <RunEntrySheet race={race} onClose={vi.fn()} onSubmitted={vi.fn()} />,
@@ -312,7 +320,6 @@ describe('RunEntrySheet', () => {
     fireEvent.mouseUp(track);
 
     expect(screen.queryByText('Disqualified')).not.toBeInTheDocument();
-    restore();
   });
 
   it('degrades to blank fields when the run-defaults request fails', async () => {
--- a/frontend/src/components/RunEntrySheet.test.tsx
+++ b/frontend/src/components/RunEntrySheet.test.tsx
@@ -230,6 +230,91 @@ describe('RunEntrySheet', () => {
     expect(img.style.display).toBe('none');
   });
 
+  // The slide-to-DQ control measures its track from `offsetWidth` (a layout
+  // value jsdom reports as 0) and maps the pointer's clientX through
+  // `getBoundingClientRect`. Pin both so the drag math is deterministic: a
+  // 300px-wide track with its left edge at x=0. The confirm threshold is
+  // `width - thumb(44) - 4 = 252`.
+  function pinSliderGeometry() {
+    Object.defineProperty(HTMLElement.prototype, 'offsetWidth', {
+      configurable: true,
+      value: 300,
+    });
+    const rect = vi
+      .spyOn(HTMLElement.prototype, 'getBoundingClientRect')
+      .mockReturnValue({
+        left: 0,
+        top: 0,
+        right: 300,
+        bottom: 48,
+        width: 300,
+        height: 48,
+        x: 0,
+        y: 0,
+        toJSON: () => ({}),
+      } as DOMRect);
+    return () => {
+      rect.mockRestore();
+      delete (HTMLElement.prototype as { offsetWidth?: number }).offsetWidth;
+    };
+  }
+
+  it('disqualifies the run when the slider is dragged past the threshold', async () => {
+    mockGameData();
+    server.use(
+      http.get(
+        '/api/v1/runs/defaults',
+        () => new HttpResponse(null, { status: 500 }),
+      ),
+    );
+    const restore = pinSliderGeometry();
+
+    render(
+      <RunEntrySheet race={race} onClose={vi.fn()} onSubmitted={vi.fn()} />,
+      { wrapper: Wrapper },
+    );
+
+    const thumb = (await screen.findByText('»')).parentElement;
+    if (!thumb?.parentElement) throw new Error('slider track not found');
+    const track = thumb.parentElement;
+
+    // Grab the thumb, drag past the 252px threshold, and release.
+    fireEvent.mouseDown(thumb);
+    fireEvent.mouseMove(track, { clientX: 300 });
+    fireEvent.mouseUp(track);
+
+    expect(await screen.findByText('Disqualified')).toBeInTheDocument();
+    restore();
+  });
+
+  it('snaps back without disqualifying when released before the threshold', async () => {
+    mockGameData();
+    server.use(
+      http.get(
+        '/api/v1/runs/defaults',
+        () => new HttpResponse(null, { status: 500 }),
+      ),
+    );
+    const restore = pinSliderGeometry();
+
+    render(
+      <RunEntrySheet race={race} onClose={vi.fn()} onSubmitted={vi.fn()} />,
+      { wrapper: Wrapper },
+    );
+
+    const thumb = (await screen.findByText('»')).parentElement;
+    if (!thumb?.parentElement) throw new Error('slider track not found');
+    const track = thumb.parentElement;
+
+    // A short drag (well under the 252px threshold) releases without confirming.
+    fireEvent.mouseDown(thumb);
+    fireEvent.mouseMove(track, { clientX: 50 });
+    fireEvent.mouseUp(track);
+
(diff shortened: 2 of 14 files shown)


frontend/src/components/RunEntrySheet.test.tsx
```diff
@@ -230,6 +230,91 @@ describe('RunEntrySheet', () => {
...
+        top: 0,
+        right: 300,
+        bottom: 48,
+        width: 300,
+        height: 48,
+        x: 0,
+        y: 0,
+        toJSON: () => ({}),
+      } as DOMRect);
+    return () => {
```
brendanbyrne: 🔵 **Suggestion — test isolation.** `pinSliderGeometry()` redefines `HTMLElement.prototype.offsetWidth` and spies on `getBoundingClientRect` *globally*, and the returned `restore()` is only called at the end of the test body. If an assertion throws first (e.g. `findByText('Disqualified')` times out), `restore()` never runs and the prototype pollution leaks into every subsequent test — latent cross-test flakiness.

Guarantee cleanup with `try`/`finally`, or register it in `afterEach`:
```ts
it('disqualifies …', async () => {
  const restore = pinSliderGeometry();
  try {
    // … render, drag, assert …
  } finally {
    restore();
  }
});
```
(`react.md` § 13 — tests shouldn't pollute global state.)

frontend/package.json
```diff
@@ -26,6 +26,7 @@
   },
   "devDependencies": {
     "@eslint/js": "^9.39.4",
+    "@rolldown/plugin-babel": "^0.2.3",
```
brendanbyrne, line 30: 🔵 **Suggestion — implicit `@babel/core` peer.** `@rolldown/plugin-babel` declares `@babel/core` as a *peer* dependency; it's currently satisfied only by Bun auto-installing `@babel/core@7.29.0` (it's in `bun.lock` but not a direct devDependency here). The Compiler's entire build pass rides on that transitively-resolved peer. It works today, but consider adding `@babel/core` as an explicit devDependency so it survives a peer-resolution change, a different package manager, or a clean CI install.

brendanbyrne (commented):
## Review — Approve ✅

**What this PR does:** Turns on the React Compiler (wired via `@rolldown/plugin-babel` + `reactCompilerPreset()` since plugin-react v6/oxc dropped the `babel` option), adds React 19 Document Metadata `<title>` to all six pages, and removes hand-written `useMemo`/`useCallback` the Compiler now handles — keeping one documented carve-out (`useAuth`'s `clearAuth`). PR-E2 of the frontend compliance plan; closes #180.

**Areas covered:** `react.md` (§ 2, § 5, § 6, § 7, § 13), `typescript.md` § 12, `design.md`, compliance plan PR-E2.

**Verified locally:** `bun run lint` → 0 errors (111 pre-existing warnings, none new); `tsc -b` clean; `vitest run` → 204 passing. Confirmed plugin-react@6.0.1 lists `@rolldown/plugin-babel` and `babel-plugin-react-compiler` as `optionalPeers`, so the vite.config wiring is the officially-sanctioned v6 integration path.

Two non-blocking 🔵 suggestions inline.

### What's good

- **The `clearAuth` carve-out reasoning is exactly right.** `clearAuth` is an effect dependency (the silent-refresh mount effect) *and* registered into module scope via `setOnAuthFailure`. The Compiler's memoization is a **performance** optimization, not a **correctness** guarantee — relying on it to keep the mount effect from re-firing would be wrong. Keeping `clearAuth` as an explicit `useCallback` (a real stability guarantee) while dropping `login`/`register`/`logout` (whose stability is only a re-render concern the Compiler handles) is the correct distinction, and it's documented inline.
- `SlideToConfirm`'s unwrapped `handleMove`/`handleEnd` are safe — called inline through JSX props, never via `addEventListener`, so referential stability was never load-bearing there.
- Document Metadata applied cleanly to all six pages incl. the phase-aware Onboarding title; only one route mounts at a time so titles swap without bleed.
- New slide-to-DQ test backfills drag handlers that previously had zero coverage; the 252px threshold math checks out.

### Design-doc accuracy: no changes required

`design.md` § Tech Stack ("React + Vite", "Tailwind") stays accurate. The Compiler, babel wiring, memoization policy, and `<title>` usage are coding-standards / build-tooling details owned by `react.md` and compliance-plan PR-E2 — not `design.md`.

### Standards-gap note (future docs tweak, not this PR)

`react.md` § 7 enumerates three memoization carve-outs (profiler bottleneck, third-party referential stability, expensive computation). The `clearAuth` case is a legitimate *fourth*: a function used as an effect dependency where correctness must not depend on Compiler memoization. Worth adding to § 7 so the next person doesn't re-derive it.

### Out-of-scope observation

`SlideToConfirm` is a `<div>` with mouse/touch handlers — no `role="slider"`, no keyboard support, no `aria-*` (a `react.md` § 10 gap). Pre-existing; this PR only touched the handlers' memoization. It's why the new test queries the thumb by `»` text + `parentElement` traversal rather than a role. Belongs to the compliance plan's accessibility stream, not here.

brendanbyrne:
Thanks for the review. Both 🔵 suggestions addressed in d28e84f:

- **Test isolation** — `pinSliderGeometry()` now registers its teardown into an `afterEach` instead of an end-of-body `restore()`, so a thrown assertion can't leak the `HTMLElement.prototype` mock into later tests. The two drag tests just call `pinSliderGeometry()` and let `afterEach` clean up.
- **Explicit `@babel/core`** — added as a direct devDependency (`^7.29.7`) so the Compiler build pass doesn't ride on transitive peer resolution.

On the two non-blocking notes I left as-is:
- **react.md § 7 fourth carve-out** (effect-dependency-correctness) — agreed it's worth documenting so the next person doesn't re-derive the `clearAuth` reasoning. Per your "not this PR" framing I've left it out of #214; happy to do it as a small standalone docs commit (with a Document history entry) — say the word and I'll push it.
- **`SlideToConfirm` a11y gap** (no `role="slider"`/keyboard) — pre-existing, belongs to the accessibility stream (PR-G2 / #184), not here.

brendanbyrne:
Heads-up @brendanbyrne — two commits landed after your approval, both worth a quick re-glance before merge:

- **d28e84f** — the two review fixes you requested (slide-to-DQ test cleanup moved to `afterEach`; explicit `@babel/core` devDependency).
- **015df7f** — `vite.config.ts`: share the `/api` proxy between `server` and `preview`. `vite preview` doesn't inherit `server.proxy`, so a production-build smoke test (`bun run preview`) couldn't reach the backend on :3000 — every relative `/api` call failed. This surfaced while I was verifying the prod build fires `/auth/refresh` once (vs StrictMode's two in dev). Verified: `/api/v1/sessions/mine` returns 401 through the preview proxy, matching :3000 directly. Dev stays on 5173, preview on Vite's default 4173 (kept distinct on purpose so the URL says which build you're testing).

CI is green and all 204 tests pass. Nothing else changed from the approved diff.

frontend/src/pages/Session.tsx
```diff
@@ -161,6 +161,7 @@ function SessionView({ id }: { id: SessionId }) {
 
   return (
     <div className="min-h-screen bg-gray-50 pb-20">
+      <title>Session · Beerio Kart</title>
```
brendanbyrne: 🔵 **Suggestion** — tab title is stale during the loading/error state. This `<title>` lives inside the loaded-view return, after the loading/error early-returns (`SessionView` returns at lines 127/135 before reaching here). While a session loads, the tab keeps the *previous* route's title (e.g. "Home · Beerio Kart") until data resolves, then flips to "Session · Beerio Kart". Same shape on `Profile.tsx` (title at the final return, after the loading/error/not-found early-returns).

Brief given polling, but a visible inconsistency. Optional fix: hoist the `<title>` above the early returns so it's set the moment the route mounts. No standard covers loading-state titles, so this is opinion, not a `react.md` violation.

brendanbyrne (commented):
## Review — adopt React Compiler, Document Metadata, drop manual memoization

**Verdict: Approve** (one minor, non-blocking suggestion).

**What this PR does:** Turns on the React Compiler (wired via `@rolldown/plugin-babel` + `reactCompilerPreset()`, since plugin-react v6 transforms with oxc rather than Babel), adopts React 19 Document Metadata for per-page `<title>`s, and removes hand-written `useMemo`/`useCallback` the Compiler now subsumes — keeping one documented carve-out.

**Areas covered:** `react.md` (§ 2, § 5, § 6, § 7, § 13), `frontend/CLAUDE.md`, `design.md`.

- `bun run lint` — 0 errors (111 warnings, all pre-existing in `Session.tsx`/`time.ts` at lines untouched by this PR).
- `bun run typecheck` — passes.
- `bun run test` — 204 passed (25 files).
- `bun run build` — succeeds; Compiler confirmed **active** (73% of build time in `@rolldown/plugin-babel`). Vitest uses its own transform pipeline, so I ran the real build to prove the `vite.config.ts` Compiler wiring actually works end-to-end — it does.

### What's good
- **The `clearAuth` carve-out is correct and necessary.** The mount effect (`useAuth.tsx`) depends on `[clearAuth]` and fires the silent-refresh; an unstable reference would re-fire it every render (a `/auth/refresh` loop). Keeping `useCallback` is a hard runtime guarantee rather than betting a mount-once effect on a transform `exhaustive-deps` can't see — a textbook `react.md` § 7(b) carve-out, and the inline comment explains exactly that.
- **Dropping `useCallback` on `SlideToConfirm`'s `handleMove`/`handleEnd` is safe** — they're inline JSX handlers (`onMouseMove`/`onMouseUp`), never `addEventListener` references in an effect, so referential stability was never load-bearing. The `Profile.tsx` comment was also correctly updated from "PR-E2 will make this redundant" to present tense.
- **The new slide-to-DQ test** covers a previously-untested interaction; the `afterEach` prototype-restore (with the comment on why an end-of-body `restore()` would leak into later tests) is exactly the right `react.md` § 13 test-isolation discipline.
- **The `apiProxy` extraction** in `vite.config.ts` (shared between `server` and `preview`) is a real fix with a clear comment — `vite preview` doesn't inherit `server.proxy`.

### Design-doc accuracy
No changes required. `design.md` describes architecture/data model, not build tooling or memoization strategy — the Compiler standard already lives in `react.md`. The compliance-plan sign-off boxes are yours to tick on merge, as the PR notes.

One 🔵 suggestion is left inline.

brendanbyrne:
Addressed the 🔵 stale-title suggestion in 97597e6.

Hoisted a shared `const pageTitle = <title>…</title>` in both `Session.tsx` and `Profile.tsx` and render it in **every** branch (loading / ended / edit-modes / default), not just the loaded view. The tab title now tracks the route from mount instead of snapping in once data resolves — and it also avoids a title revert when entering Profile's edit sub-views.

(Caught a self-inflicted bug while doing it: the Session `pageTitle` briefly had a literal `·` instead of `·`; fixed and byte-verified before commit.)

Lint 0 errors, 204 tests pass, build clean, patch coverage 100%.

Issue #206, opened 2026-05-22
Co-locate src/__tests__/App.test.tsx per typescript.md § 12


brendanbyrne:
## Context

Surfaced in the [PR #205](https://github.com/brendanbyrne/beerio-kart/pull/205) review (PR-D1, named exports). Out of scope there — PR-D1 only touched the file's import line.

`docs/coding-standards/typescript.md` § 12 calls for **co-located** tests (`*.test.tsx` next to the code under test). One file predates that rule and still lives in a sibling directory:

- `frontend/src/__tests__/App.test.tsx`

This is the only test in `src/__tests__/`. `frontend/CLAUDE.md` § Testing also describes `src/__tests__/` as the home for end-to-end **integration** flow tests (login → create session → submit run), distinct from co-located unit tests — so this isn't a purely mechanical move.

## The decision to make

`App.test.tsx` is an integration test (renders the full provider tree), which is exactly what `frontend/CLAUDE.md` says `__tests__/` is *for*. So § 12's "co-locate" and CLAUDE.md's "integration tests live in `__tests__/`" are in mild tension for this file. Resolve it one of two ways:

1. **Move it** to `frontend/src/App.test.tsx` (co-located beside `App.tsx`), treating § 12 as the governing rule for all tests, and narrow CLAUDE.md's `__tests__/` guidance to "multi-page flow tests that don't belong to one component."
2. **Keep it** in `__tests__/` and add an explicit carve-out in `typescript.md` § 12 that integration/flow tests live in `src/__tests__/`, co-location applying to unit tests only.

Either is fine; the point is to make the two docs agree and stop the lone file from looking like an oversight.

## Acceptance criteria

- `App.test.tsx` location matches whatever the two docs settle on.
- `typescript.md` § 12 and `frontend/CLAUDE.md` § Testing no longer contradict each other on where flow/integration tests live.
- `bun run test` still green.

## Effort

S — a `git mv` plus a one-paragraph doc reconciliation, or just the doc carve-out.

Issue #185, opened 2026-05-18
Frontend final lint cleanup, DTO drift-check CI, and test-coverage backfill


brendanbyrne:
## Context

Final compliance PR. Sweeps any residual `string`-typed status fields to literal unions, adds a CI check that flags `backend/src/` DTO changes lacking a corresponding `frontend/src/api/types.ts` change (cheap drift detection without committing to Rust→TS codegen — see the type-sync research doc for the full rationale), and runs the test-coverage backfill that catches anything earlier PRs in this milestone missed.

The per-PR testing convention introduced by `typescript.md` § 12 and `react.md` § 13 means every prior PR should have added tests for what it touched. This PR is the safety net for anything that slipped through — it should be small if the earlier PRs held to the standard.

## Acceptance criteria

- [ ] Convert any remaining `string`-typed status / kind fields to literal unions.
- [ ] Confirm no TS `enum` usage anywhere (audit confirms none today; `no-restricted-syntax` rule keeps it that way).
- [ ] Add a GitHub Actions check that fails the PR if `backend/src/dtos/` (verify the actual path during implementation) changes without `frontend/src/api/types.ts` also changing. False positives are acceptable — contributor adds a one-line update.
- [ ] Document the drift check in `docs/coding-standards/typescript.md` § 11 and link from `docs/research/rust-to-ts-codegen.md`.
- [ ] **Test-coverage backfill.** Run `bun run test:coverage` and verify no production code outside the standards' carve-outs (pure presentational components, app shell composition, generated types — see `typescript.md` § 12) is uncovered. Add tests for any straggler files surfaced. If this turns into a multi-day task, earlier PRs in this milestone missed their AC; flag it in the PR description.
- [ ] **Flip the Codecov frontend project gate to blocking.** PR-H2 (#193) set `coverage.status.project.frontend.informational: true` in `codecov.yml` because the frontend was un-ignored before it had tests. Once the test-coverage backfill above lands, change it to `informational: false` so frontend project-coverage regressions block PRs the same way backend's already do. (The frontend *patch* gate already blocks — PR-H2 set it; only the *project* gate needs flipping.) Update `docs/design.md` § Coverage & CI to drop the "informational until PR-H1" qualifier.

## References

- Compliance plan: PR-H1
- Type-sync evaluation: `docs/research/rust-to-ts-codegen.md`
- Standards: `docs/coding-standards/typescript.md` § 4, § 11, § 12; `docs/coding-standards/react.md` § 13

[...]

PR #218, opened 2026-05-31
190: migrate router to createBrowserRouter with error boundaries and lazy routes

### LINKED ISSUE #190: Upgrade router to createBrowserRouter with error boundaries and lazy routes

brendanbyrne:
## Reviewer notes

Diff looks large but is mostly mechanical: `App.tsx` is a near-total rewrite of the routing tree (old `<Routes>` JSX → declarative `routes` table), and `BottomNav.tsx`'s tab loop was unrolled to split the `<NavLink>` (enabled) and `<button disabled>` (no-session) cases. The three new fallback/skeleton components are small and self-contained. No behavior change to any page's internals.

One judgment call worth a glance: the catch-all (`path: '*'`) still **redirects unknown paths to `/`** rather than showing a 404. An earlier `App.tsx` comment had floated "PR-F1 will replace this with a 404" — I kept the redirect (friendlier per design.md's "never a burden" / "sensible defaults"; a dedicated 404 isn't in the AC). The `errorElement` work is orthogonal — it handles render *crashes* on matched routes, not unmatched paths.

## Summary

PR-F1 of the [frontend compliance plan](docs/designs/2026-05-16-frontend-compliance-plan.md). Migrates react-router to `createBrowserRouter` + `RouterProvider`, adds an app-level error boundary plus per-route `errorElement`, code-splits every page with `lazy()` + `<Suspense>`, converts the bottom-nav buttons to real `<NavLink>`s, and flips the last two warn-level lint rules (`no-explicit-any`, `no-non-null-assertion`) to error.

Closes #190.

## What changed

- **Router migration** ([App.tsx](frontend/src/App.tsx)): `BrowserRouter` + `<Routes>` → `createBrowserRouter` + `RouterProvider` (react.md § 11, declarative mode). The provider stack `ErrorBoundary > QueryClientProvider > AuthProvider > AuthGate` wraps `RouterProvider` — none of those need router context, so this is simpler than a layout route, and React context still flows into route elements. The `RequireAuth`/`GuestOnly` guards stay element wrappers (they use `<Navigate>`, which has router context inside `RouterProvider`). All routes preserved with identical guard semantics; catch-all `*` still redirects home.
- **Lazy routes / code-splitting** (react.md § 11): all six pages are `lazy()`-loaded via the named-export adapter `(m) => ({ default: m.X })` and wrapped in `<Suspense fallback={<PageSkeleton/>}>` (inside the guard, so no skeleton flash on redirect). `vite build` now emits a separate chunk per page (Login/Register/Onboarding/Home/Profile/Session) instead of one bundle.
- **Error boundaries** (react.md § 9): the app-level [`react-error-boundary`](frontend/src/components/AppErrorFallback.tsx) is the outermost catch-all for crashes *outside* the routed tree; each route gets a route-scoped [`errorElement`](frontend/src/components/RouteErrorFallback.tsx) through a shared `route()` factory — so a render crash on one page is caught route-scoped (via `useRouteError`) and never blanks the app. New: [`PageSkeleton`](frontend/src/components/PageSkeleton.tsx) (Suspense fallback, `role="status"`).
- **BottomNav → NavLink** ([BottomNav.tsx](frontend/src/components/BottomNav.tsx), react.md § 10/§ 11): `<button onClick={navigate}>` → `<NavLink>`, which renders a real `<a>` (middle/cmd-click work, keyboard `Enter` activates, screen readers announce a link) and sets `aria-current="page"` itself. The Session tab, when there's no reachable session, stays a disabled `<button>` (NavLink has no disabled state). Home uses `end` so `/` isn't active on every path.
- **Lint flip** ([eslint.config.js](frontend/eslint.config.js)): `no-explicit-any` + `no-non-null-assertion` `warn` → `error`. Offenders were already removed (the `any`-typed `res.json()` by PR-B2, the `!` assertions by PR-D2), so the flip lands with zero violations.

## Automated verification

```bash
cd frontend
bun run typecheck   # tsc -b → 0 errors
bun run lint        # eslint → 0 errors (warnings are pre-existing warn-down rules owned by G1/G2/H1)
bun run test        # vitest → 217 passing (29 files)
bun run build       # emits separate dist/assets/{Login,Register,Onboarding,Home,Profile,Session}-*.js chunks
```

Diff (patch) coverage: **100%** of changed coverable lines.

## How to verify

Start the backend and frontend from the repo root (two terminals):

```bash
cargo run                 # terminal 1 — API on :3000
cd frontend && bun run dev # terminal 2 — app on :5173 (proxies /api → :3000)
```

Open http://localhost:5173 on a phone-width viewport (Pixel 9 Pro reference), and check **both Chrome and Firefox**:

- [x] **Every route navigates.** Log in → land on Home. Use the bottom nav to move Home ↔ Session ↔ Profile. Visit `/onboarding` and `/register` directly. No blank screens, no console errors.
- [ ] **Lazy chunks load per route.** Open DevTools → Network (filter JS) and hard-reload. Navigating to a page you haven't visited fetches a new `Profile-*.js` / `Session-*.js` / etc. chunk on demand (not all upfront). On a throttled connection ("Slow 3G") you briefly see the loading skeleton before the page paints.
- [x] **NavLink semantics.** The active tab is highlighted and (DevTools → inspect) carries `aria-current="page"`; the others don't. **Middle-click** or **Cmd/Ctrl-click** a tab opens it in a new tab (real anchor — the old `<button>` couldn't). With no active session, the Session tab is greyed out and not focusable via Tab.
- [x] **Catch-all.** Manually enter a bogus URL like http://localhost:5173/does-not-exist → redirected to Home (or `/login` if logged out), not a blank page.
- [ ] **Route-scoped error boundary.** Temporarily add `throw new Error('test');` at the top of one page component (e.g. [Profile.tsx](frontend/src/pages/Profile.tsx)). Navigate there → you see **"This page ran into a problem"** with a working **Go home** link — and the global "Something went wrong" screen does **not** appear. Other routes still work. Revert the throw.

## Author checklist

- [x] Tests added or updated for new logic (NavLink roles + aria-current, route-crash layering, catch-all, per-route errorElement, authenticated lazy routing, the three new components).
- [x] Docs updated if applicable — none needed (no `docs/` files in scope; the compliance-plan sign-off is Brendan's on merge).
- [x] Schema changes: none.
- [x] Tested locally end-to-end (automated suite + build; manual steps above for the reviewer).

### DIFF
--- a/frontend/src/__tests__/App.test.tsx
+++ b/frontend/src/__tests__/App.test.tsx
@@ -68,10 +68,14 @@ describe('App shell', () => {
     ).toBeInTheDocument();
   });
 
-  it('attaches a route-scoped errorElement to every top-level route', () => {
+  it('attaches the route-scoped errorElement to every top-level route', () => {
     expect(routes.length).toBeGreaterThan(0);
     for (const r of routes) {
-      expect(r.errorElement).toBeDefined();
+      // toEqual, not toBeDefined: pin that it's specifically RouteErrorFallback
+      // (route-scoped), so a regression wiring the global AppErrorFallback — or
+      // any other element — onto a route fails here. React elements compare
+      // structurally under toEqual.
+      expect(r.errorElement).toEqual(<RouteErrorFallback />);
     }
   });
 
--- a/frontend/src/__tests__/App.test.tsx
+++ b/frontend/src/__tests__/App.test.tsx
@@ -1,20 +1,35 @@
 import { http, HttpResponse } from 'msw';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 import { render, screen } from '@testing-library/react';
+import { RouterProvider, createMemoryRouter } from 'react-router-dom';
+import { ErrorBoundary } from 'react-error-boundary';
 import { server } from '../mocks/server';
-import { App } from '../App';
+import { App, AppProviders, routes } from '../App';
+import { AppErrorFallback } from '../components/AppErrorFallback';
+import { RouteErrorFallback } from '../components/RouteErrorFallback';
 
-// Smoke test for the app shell wired up in PR-C1 (Issue #176): the full
-// provider tree — QueryClientProvider → BrowserRouter → AuthProvider — must
-// compose and mount. With no valid refresh cookie the silent refresh 401s,
-// AuthGate finishes loading, and an unauthenticated visitor is redirected to
-// the login screen. MSW mocks the network at the fetch boundary (react.md
-// § 13).
+// App shell wired up across PR-C1 (provider tree) and PR-F1 (Issue #190 —
+// createBrowserRouter + RouterProvider, lazy routes, app-level + per-route
+// error boundaries). The full stack — ErrorBoundary → QueryClientProvider →
+// AuthProvider → AuthGate → RouterProvider — must compose and mount, an
+// unauthenticated visitor must land on the login screen, unknown paths must
+// route through the catch-all, and a render crash in a route must be caught by
+// that route's scoped fallback rather than the global one. MSW mocks the
+// network at the fetch boundary (react.md § 13).
 
 // Force the dev-only React Query Devtools branch off: the panel touches
 // browser APIs jsdom doesn't implement, and the test only needs the shell to
 // mount. The `import.meta.env.DEV && …` line still evaluates either way.
 vi.stubEnv('DEV', false);
+
+function unauthenticated() {
+  server.use(
+    http.post(
+      '/api/v1/auth/refresh',
+      () => new HttpResponse(null, { status: 401 }),
+    ),
+  );
+}
 
 afterEach(() => {
   vi.unstubAllEnvs();
@@ -22,12 +37,7 @@ afterEach(() => {
 
 describe('App shell', () => {
   it('renders the login screen for an unauthenticated visitor', async () => {
-    server.use(
-      http.post(
-        '/api/v1/auth/refresh',
-        () => new HttpResponse(null, { status: 401 }),
-      ),
-    );
+    unauthenticated();
 
     render(<App />);
 
@@ -37,22 +47,61 @@ describe('App shell', () => {
   });
 
   it('routes unknown paths through the catch-all instead of rendering blank', async () => {
-    // PR-D2 / Issue #192 added <Route path="*" /> so paths that don't match
-    // any declared route (e.g. /session with no :id segment) redirect to "/"
-    // rather than rendering an empty <Routes>. With no auth, the / route
-    // bounces to /login, so the user-visible end state is the login screen.
-    server.use(
-      http.post(
-        '/api/v1/auth/refresh',
-        () => new HttpResponse(null, { status: 401 }),
-      ),
-    );
+    // Paths that don't match any declared route (e.g. /session with no :id
+    // segment) redirect to "/" rather than dead-ending. With no auth the /
+    // route bounces to /login, so the user-visible end state is the login
+    // screen. Driven through a memory router so the initial path is explicit
+    // (the module-scope browser router shares real history across renders).
+    unauthenticated();
 
-    window.history.pushState({}, '', '/session');
-    render(<App />);
+    const router = createMemoryRouter(routes, {
+      initialEntries: ['/session'],
+    });
+    render(
+      <AppProviders>
+        <RouterProvider router={router} />
+      </AppProviders>,
+    );
 
     expect(
       await screen.findByRole('button', { name: /log in/i }),
     ).toBeInTheDocument();
+  });
+
+  it('attaches a route-scoped errorElement to every top-level route', () => {
+    expect(routes.length).toBeGreaterThan(0);
+    for (const r of routes) {
+      expect(r.errorElement).toBeDefined();
+    }
+  });
+
+  it('catches a route render crash with the route-scoped fallback, not the global one', async () => {
+    // A thrown render error inside a route element must be caught by that
+    // route's errorElement (RouteErrorFallback) — the app-level ErrorBoundary
+    // (AppErrorFallback) stays dormant. This is the layering the app relies on
+    // so one broken page can't blank the whole tree.
+    const errorSpy = vi
+      .spyOn(console, 'error')
+      .mockImplementation(() => undefined);
+
+    function Boom(): never {
+      throw new Error('kaboom');
+    }
+    const router = createMemoryRouter(
+      [{ path: '/', element: <Boom />, errorElement: <RouteErrorFallback /> }],
+      { initialEntries: ['/'] },
+    );
+    render(
+      <ErrorBoundary FallbackComponent={AppErrorFallback}>
+        <RouterProvider router={router} />
+      </ErrorBoundary>,
+    );
+
+    expect(
+      await screen.findByText(/this page ran into a problem/i),
+    ).toBeInTheDocument();
+    expect(screen.queryByText(/something went wrong/i)).not.toBeInTheDocument();
+
+    errorSpy.mockRestore();
   });
 });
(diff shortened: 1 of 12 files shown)


frontend/src/components/RouteErrorFallback.tsx
```diff
@@ -0,0 +1,43 @@
...
+      {import.meta.env.DEV && (
+        <pre className="max-w-full overflow-auto text-xs text-red-500">
+          {detail}
+        </pre>
+      )}
+      <Link
+        to="/"
+        className="min-h-[44px] flex items-center px-6 py-2.5 bg-blue-500 text-white font-semibold rounded-xl hover:bg-blue-600 transition-colors"
+      >
+        Go home
```
brendanbyrne: 🔵 **Suggestion** — the route fallback offers only "Go home", which is the weak recovery for its most common real-world trigger. In a code-split SPA, the thing that most often actually trips a route `errorElement` isn't a logic crash — it's a **dynamic-import failure** after a redeploy: a user on a stale `index.html` navigates to a route whose chunk hash no longer exists, `lazy()` throws, and this fallback renders. "Go home" is a client-side `<Link>` nav, so it tries to load *another* chunk from the same stale manifest and can loop without recovering. `AppErrorFallback` already does the honest thing for shell crashes (`window.location.reload()`, which re-fetches `index.html` + a fresh manifest) — consider giving this fallback a reload affordance alongside "Go home". Not required by `react.md` § 9; a judgment call, but the chunk-load case is the realistic one for `lazy()` routes.

frontend/src/__tests__/App.test.tsx
```diff
@@ -37,22 +47,61 @@ describe('App shell', () => {
...
 
     expect(
       await screen.findByRole('button', { name: /log in/i }),
     ).toBeInTheDocument();
   });
+
+  it('attaches a route-scoped errorElement to every top-level route', () => {
+    expect(routes.length).toBeGreaterThan(0);
+    for (const r of routes) {
+      expect(r.errorElement).toBeDefined();
```
brendanbyrne: 🔵 **Suggestion (test quality)** — `toBeDefined()` asserts existence, not identity, so this passes for *any* truthy element: a regression that attached `AppErrorFallback` (the global one) or a stale element to a route would stay green, despite the test name promising the *route-scoped* fallback. The contract it gestures at is already proven behaviorally by the next test (`catches a route render crash with the route-scoped fallback, not the global one`), so no coverage is actually missing — this is just a loose matcher. Tighten to `expect(r.errorElement).toEqual(<RouteErrorFallback />)` (React elements compare structurally under `toEqual`), or drop it as redundant. Minor: the loop also covers the catch-all `route('*', <Navigate/>)`, whose `errorElement` is dead code (a redirect can't render-crash), so "every top-level route" is vacuously true for that entry.

brendanbyrne (commented):
## Code Review — Approve ✅

Migrates the router to `createBrowserRouter` + `RouterProvider`, adds an app-level error boundary plus per-route `errorElement`, code-splits all six pages with `lazy()` + `<Suspense>`, converts BottomNav tabs to real `<NavLink>`s, and flips `no-explicit-any` / `no-non-null-assertion` to error.

**Standards reviewed:** `react.md` §§ 9, 11, 13 · `typescript.md` § 5 · `design.md` · plus a `test-quality-review` pass on the six changed test files.

**Automated checks (run locally):**
- `bun run lint` → **0 errors** (109 warnings, all in files this PR didn't touch — owned by #185 / G1–H1).
- `bun run typecheck` (`tsc -b`) → clean.
- `bun run test` → **217 passing / 29 files**.

### Findings
Two 🔵 Suggestions, both inline. No blockers.

### Design-doc accuracy
✅ **No changes required.** `design.md` is the backend/architecture-level doc and delegates UI Screens / routing to companion docs; it has no frontend routing section this PR would touch. Frontend architecture here is governed by `react.md` §§ 9/11 and the compliance plan. The author's "no docs in scope" is correct.

### Milestone context — `Hardening: Frontend standards compliance`
Deferrals verified before flagging anything:
- Keeping the integration test in `src/__tests__/` (vs co-locating) is owned by sibling **#206** (open) — not flagged.
- The 109 pre-existing lint warnings are owned by **#185** / G1–H1 — not flagged.
- No dedicated 404 page is a deliberate, documented choice (redirect-home per design.md's "never a burden"); the AC never asked for one — not a finding.
- Broader a11y is sibling **#184**; the new components here meet the floor on their own.

### What's good
- **The error-boundary layering is genuinely well done** — app-level `ErrorBoundary` *outside* `RouterProvider`, per-route `errorElement` *inside* — and the behavioral test (`catches a route render crash with the route-scoped fallback, not the global one`) proves the isolation with disjoint fallback strings so the "global stays dormant" half can't accidentally match.
- **Suspense nested *inside* the auth guard** so a redirect short-circuits before the skeleton renders — no skeleton flash on bounce. Nice detail, and the comment says why.
- **BottomNav → `<NavLink>` is a real a11y win** (true anchors → middle/cmd-click + keyboard `Enter` + screen-reader "link", `aria-current="page"` for free), tested thoroughly including the `end`-prop negative case and the disabled-`<button>` no-session fallback.
- `PageSkeleton` ships `role="status"` + an `sr-only` label — an accessible loading state, pinned by a test.
- The named-export → `React.lazy` adapter is correct and documented; the `react-refresh/only-export-components` opt-out on `routes` is justified and cites the `useAuth.tsx` precedent.
- Comments explain *why*, not *what* — exactly the things a future reader would otherwise reverse-engineer.

brendanbyrne:
Thanks — both 🔵 suggestions addressed in 719908e.

**1. RouteErrorFallback reload affordance** ([RouteErrorFallback.tsx](frontend/src/components/RouteErrorFallback.tsx)) — fully agree on the chunk-load-after-redeploy framing; that's the realistic trigger for a `lazy()` route's `errorElement`, and "Go home" alone can loop on a stale manifest. Added a primary **Reload** button (`window.location.reload()` → re-fetches `index.html` + a fresh manifest) with "Go home" demoted to a secondary text link for the actual-render-crash case. The rationale is now in the file's header comment. New test stubs `window.location.reload` and asserts the click fires it.

**2. `toBeDefined()` → `toEqual(<RouteErrorFallback />)`** ([App.test.tsx](frontend/src/__tests__/App.test.tsx#L71)) — tightened. The structural test now pins *identity*, so a regression wiring the global `AppErrorFallback` (or any other element) onto a route fails here instead of staying green. Re the catch-all's `errorElement` being dead code for a redirect: kept the `route()` factory uniform (a redirect can't render-crash, so it's harmless), and `toEqual` makes the assertion non-vacuous for that entry too — it still pins the element rather than just asserting existence.

Also covered the `String(error)` branch in `AppErrorFallback` (non-`Error` thrown value) while in the area, which clears the last codecov partial.

Verification after the changes: `bun run typecheck` clean · `bun run lint` 0 errors · `bun run test` **218 passing** · build still emits per-route chunks · patch coverage **96.7%** (codecov-style, counting branch partials).

PR #221, opened 2026-05-31
184: accessibility sweep — touch targets, modal focus, a11y lint

### LINKED ISSUE #184: Frontend accessibility sweep — touch targets, focus management, modal a11y

brendanbyrne:
## Summary

PR-G2 of the [frontend compliance plan](docs/designs/2026-05-16-frontend-compliance-plan.md). The accessibility sweep: brings interactive elements to the 44 px touch-target floor, makes the two modals real dialogs (focus trap, Escape, focus restore, `role`/`aria`), removes the `autoFocus` regression, makes the DQ slider keyboard-operable, adds runtime axe auditing in dev, and flips the four held-back `jsx-a11y` rules from `warn` to `error` now that every offender is fixed.

Closes #184.

Standards: [`react.md` § 10](docs/coding-standards/react.md), [`tailwind.md` § 5](docs/coding-standards/tailwind.md).

## What changed

**New `useModalA11y` hook** ([useModalA11y.ts](frontend/src/hooks/useModalA11y.ts)) — traps Tab focus inside the dialog, closes on Escape, restores focus to the trigger on close, and can be *suspended* while a full-screen sub-view owns focus (RunEntrySheet's drink/setup pickers).

> **Design note — hook vs native `<dialog>`.** The plan flagged native `<dialog>` as the recommended escape hatch "because focus trap is tricky." I went with a small hook instead because **jsdom doesn't implement `showModal()`'s focus trap, Escape handling, or focus restoration** — so with native `<dialog>` the AC's keyboard behaviors would be untestable (assertions would be theater, which the test-quality standard forbids). The hook keeps the existing markup a visual no-op *and* is fully exercisable in jsdom. Native `<dialog>` remains a reasonable future swap if we add a browser-level e2e layer.

- **Modals** — Home create-session modal and the RunEntrySheet bottom sheet get `role="dialog"`, `aria-modal="true"`, the focus trap, and Escape. Backdrops became keyboard-excluded `<button>`s, which clears `no-static-element-interactions` without losing click-to-dismiss (and drops the `stopPropagation` shim since the content is now a backdrop sibling, not a child).
- **Touch targets → `min-h-touch`** — DrinkTypeSelector skip/add/cancel, RaceSetupPicker step pills + Skip/Confirm, Profile password Cancel/Save, Home modal Cancel, Login/Register submit.
- **`autoFocus` removed** from the add-drink form input.
- **DQ slider** made keyboard-operable (`role="button"` + Enter/Space); the thumb is now presentational and drag handlers consolidated on the track (pointer behavior unchanged).
- **Accessible names** added to the time inputs, drink-name, and password inputs (axe label coverage on the main flows).
- **`@axe-core/react`** wired dev-only in [main.tsx](frontend/src/main.tsx) via dynamic import — verified absent from the production bundle.
- **Lint** — `jsx-a11y/no-static-element-interactions`, `click-events-have-key-events`, `label-has-associated-control`, `no-autofocus`: `warn` → `error`.

## Test plan

Automated (all green):

```bash
cd frontend
bun run test       # 233 pass (new: useModalA11y unit tests; Home Escape+restore; RunEntrySheet Escape / keyboard-DQ / touch-DQ)
bun run lint       # 0 errors — the 4 jsx-a11y rules are now error-level with zero violations
bun run typecheck  # clean
bun run build      # succeeds; `grep -r axe dist/` finds nothing
```

Patch coverage: **98%** on changed source lines.

Manual (run on the Pixel 9 Pro viewport — DevTools device toolbar):

```bash
cd frontend && bun run dev   # open the printed URL
```

**Keyboard — modals**
1. On Home, Tab to **Start a Session** → Enter. The *Pick a Ruleset* dialog opens; focus lands on **Random**.
2. Tab repeatedly → focus cycles only between **Random** and **Cancel** (never reaches the page behind). Shift+Tab cycles backward.
3. **Escape** → dialog closes, focus returns to **Start a Session**.
4. Reopen, click the dimmed backdrop → closes (mouse dismiss still works).
5. Enter a session → open the run-entry sheet (**Submit Time**). Tab through the time fields → focus stays in the sheet. **Escape** → sheet closes.

**Keyboard — DQ slider**
6. In the sheet, Tab to *Slide … to disqualify* and press **Enter** (or Space) → flips to the Disqualified state. (Pointer drag still works.)

**Touch targets**
7. Confirm these are comfortably tappable (≥ 44 px tall): DrinkTypeSelector skip/add/cancel, RaceSetupPicker step pills + Skip/Confirm, Profile password Cancel/Save, Home modal Cancel, Login/Register submit.

**axe (runtime a11y, dev only)**
8. Open the browser console. Walk Login → Home → create session → run-entry sheet and confirm **no axe violations** are logged on these main flows.

## Notes / scope
- The RunEntrySheet's nested drink/setup pickers are full-screen views with their own Back control (not modal dialogs), so they're out of this PR's scope — the trap suspends while they're open rather than fighting them for focus.
- 🤖 Generated with [Claude Code](https://claude.com/claude-code)


frontend/src/hooks/useModalA11y.ts
```diff
@@ -0,0 +1,108 @@
+import { useEffect, useRef } from 'react';
+
+// Tab-order members inside a modal. Mirrors the set browsers treat as
+// sequentially focusable; `[tabindex="-1"]` is programmatically focusable but
+// not part of Tab order, so it's excluded.
+const FOCUSABLE_SELECTOR = [
+  'a[href]',
+  'button:not([disabled])',
```
brendanbyrne: 🔵 **Suggestion — tighten the selector to match the comment above it.** `'button:not([disabled])'` (and the `input`/`select`/`textarea`/`a` siblings) will match an element carrying `tabindex="-1"` — `querySelectorAll` returns it via *this* clause even though the trailing `[tabindex]:not([tabindex="-1"])` clause tries to exclude it. The lines 3–5 comment states `[tabindex="-1"]` "is excluded," so today the code and comment disagree.

No current dialog trips it — the only `tabindex={-1}` element (the backdrop button) is a *sibling* of the dialog, outside this `node.querySelectorAll` root — so it's latent. But since this is shared infrastructure meant for reuse, tightening closes it:

```ts
const FOCUSABLE_SELECTOR = [
  'a[href]',
  'button:not([disabled]):not([tabindex="-1"])',
  'textarea:not([disabled]):not([tabindex="-1"])',
  'input:not([disabled]):not([tabindex="-1"])',
  'select:not([disabled]):not([tabindex="-1"])',
  '[tabindex]:not([tabindex="-1"])',
].join(', ');
```

Not a blocker.

frontend/src/components/RunEntrySheet.tsx
```diff
@@ -70,6 +71,11 @@ export function RunEntrySheet({
   const [showDrinkPicker, setShowDrinkPicker] = useState(false);
   const [showSetupPicker, setShowSetupPicker] = useState(false);
 
+  // Trap focus / close-on-Escape / restore-focus for the sheet (react.md § 10).
+  // The trap suspends while a full-screen sub-picker is open so it doesn't fight
+  // the picker for focus; the picker has its own Back control.
+  const dialogRef = useModalA11y(onClose, !showDrinkPicker && !showSetupPicker);
```
brendanbyrne, line 77: 🔵 **Suggestion — sub-picker focus handling is incomplete (out of PR-G2 scope, follow-up only).** Suspending the trap while a drink/setup picker is open is reasonable, but those pickers render as opaque `fixed inset-0 z-50` overlays *on top of the still-mounted sheet* (lines 420–475). The consequence: (1) on open, focus stays on the now-covered trigger button and nothing traps focus inside the picker, so a keyboard / screen-reader user can Tab into the occluded sheet behind it; (2) on close, the trap re-runs and re-seats focus to the *first* control in the sheet rather than the drink/setup button the user came from.

This isn't in PR-G2's acceptance criteria — the compliance plan lists only the Home modal and the RunEntrySheet bottom sheet, and the PR documents the tradeoff — so **not a blocker**. The clean fix when these views are addressed: move focus into the picker on open and mark the sheet `inert` while a picker is open (which also handles the SR-leak), then restore to the trigger on close.

Flagging because no sibling Issue currently owns sub-view focus management — recording it so it isn't lost.

brendanbyrne (commented):
## Code review — accessibility sweep (PR-G2)

**Verdict: Approve** — no blockers. Three 🔵 suggestions below, none gating. (As always, the formal approve is Brendan's; this review posts as a comment.)

**What this PR does:** Brings the frontend to the `react.md` § 10 / WCAG accessibility floor — 44 px touch targets, two real focus-trapped modals via a new `useModalA11y` hook, a keyboard-operable DQ slider, dev-only axe auditing — and flips four `jsx-a11y` rules from `warn` back to `error` now that every offender is fixed.

**Areas reviewed:** `react.md` (§ 2, 6, 7, 10, 13), `tailwind.md` (§ 5), `typescript.md` (§ 9), `design.md`, and the PR-G2 scope in the frontend compliance plan.

**Verified locally (all green):**
- `bun run lint` → **0 errors**, 90 warnings — all pre-existing in `Session.tsx`/`time.ts`, none in changed files.
- `bun run test` → **233 pass (30 files)**.
- Test quality: a `test-quality-review` pass mutation-tested the four highest-risk new tests (focus-restore no-op, `onTouchStart` neutered, Enter-handler removed) and confirmed each test **fails** when its claimed behavior is broken — the assertions pin real behavior, not green-by-construction. `fireEvent.keyDown(document, …)` correctly targets the document-level listener the hook installs.

### Design-doc accuracy
**No changes required.** `design.md` is architecture/data-model and delegates UI to `user-workflows.md`; the a11y floor and modal contract already live in `react.md` § 10 / `tailwind.md` § 5, which this PR *satisfies* rather than changes. The new hook is a small reusable utility, below the design-doc threshold.

### 🔵 Suggestion (docs, not in this diff) — record `useModalA11y` as the canonical overlay pattern
`react.md` § 10 frames the modal rule around native `<dialog>` / "a custom overlay." The hook's jsdom-testability rationale is sound and well-documented in the code. A one-line addition to § 10 pointing future overlays at `useModalA11y` (instead of re-rolling a trap) would keep the next modal consistent. Optional, and a separate docs edit since `react.md` isn't touched here.

### What's good
- **The hook-vs-native-`<dialog>` call is the right one and well-justified** — jsdom genuinely doesn't implement `showModal()`'s focus trap / Escape / restore, so native would force theater assertions. Correctly prioritizes the testing standard (`typescript.md` § 12).
- **Effect discipline is exemplary** in `useModalA11y.ts`: the trigger is captured *before* the trap seats focus (effect ordering), `onClose` is read through a ref so the keydown listener isn't reinstalled per render, and `[active]` deps are honest to `exhaustive-deps` — no suppressions (`react.md` §§ 6–7).
- **Flipping the four rules to `error` brings `eslint.config.js` back into line with `typescript.md` § 9** (`jsx-a11y/recommended` is error-level by default) — this closes the warn-down debt rather than masking it.
- `min-h-touch` (not a bare `py-3`) is the correct fix — it enforces the 44 px floor independent of padding, per `tailwind.md` § 5.

Nice work.

brendanbyrne:
Thanks for the review — addressed the three 🔵 suggestions and the codecov partials in `d7b7fc5`:

- **Focus selector (useModalA11y.ts:8)** — tightened: every clause now excludes `:not([tabindex="-1"])`, so the code matches its comment (a focusable tag carrying `tabindex="-1"` opts out of the trap). Was latent (backdrop button is a sibling outside the query root); fixed since the hook is shared infra.
- **Sub-picker focus (RunEntrySheet:77)** — agreed it's out of PR-G2 scope; recorded as **#222** (Ready / Medium, Hardening milestone) with the `inert`-the-sheet + move-focus-in/restore-on-close approach you outlined, so it isn't lost.
- **`react.md` § 10 pointer to `useModalA11y`** — good call; I'll add it as a **post-merge docs commit** rather than fold it in here. The hook only exists on `main` once this merges, and touching any `.md` re-triggers the repo-wide link-check on an otherwise code-only PR. (The post-merge skill now prompts for exactly this kind of follow-up doc.)
- **Codecov branch partials** — closed with real-behavior tests: Tab from *outside* the dialog re-enters the trap (both directions), mid-list Tab is left to the browser (trap only acts at the edges), and the DQ slider activates on **Space** as well as Enter. `useModalA11y` is now fully covered; 236 tests pass.

brendanbyrne:
Update on the `react.md` § 10 pointer: did it **in-PR** after all (`5981de4`) rather than deferring — the standards doc should ship with the hook it documents. § 10's Modals rule now points future overlays at `useModalA11y` and records the hook-vs-native-`<dialog>` rationale. link-check passed (the relative link resolves; no repo-wide rot). All checks green.

Issue #220, opened 2026-05-31
Tighten up README.md files


brendanbyrne:
Some of the READMEs, in particular the top level one, are getting bloated.  I think a good rule to instate is that, in general, it should be preferred to only refer to the files and directories directly in the same directory as you.  For example, the top level README shouldn't contain a detailed description of the files in `docs`.  It should just point to docs and explain what it is.

The top level README also shouldn't have the whole project structure listed in it.  It doesn't provide much value.  It's cumbersome to keep updated and you can get the same information by just running `tree`.  The intent is that if the purpose and contents of a directory aren't self evident, then there should be a readme present.

**To the agent: when starting this issue, sync with me to make sure what we're in agreement on how the files should be organized.**

PR #224, opened 2026-05-31
206: co-locate App/routing tests; reconcile test-location docs

### LINKED ISSUE #206: Co-locate src/__tests__/App.test.tsx per typescript.md § 12

brendanbyrne:
Closes #206.

## What & why

`frontend/src/__tests__/` held two tests that don't belong there. Both import `routes`/`AppProviders`/`App` from `../App` and verify the **App module** — provider wiring, route→lazy-chunk mapping, error-boundary layering. They *target one source file*; they are **not** the multi-screen user journeys (`login → create session → submit run`) that `frontend/CLAUDE.md` § Testing and `react.md` § 13 reserve `src/__tests__/` for.

So they were misplaced relative to `typescript.md` § 12 ("co-locate tests next to the file they test"). #206 surfaced this for `App.test.tsx`; the docs also contradicted each other on where flow tests live.

## Decisions taken (synced with Brendan)

1. **Move, don't carve out.** `App.test.tsx` is a test *of `App.tsx`*, not a flow → co-locate it, and make the docs agree.
2. **Move `routing.test.tsx` too.** It was added *after* #206 was filed (so the issue's "only test in `__tests__/`" was stale) and is the same shape — the authenticated-routing half of the same App-module test pair. Leaving it would recreate the exact oversight #206 fixes. Beyond the issue's literal text but squarely its intent; confirmed before expanding scope.

## Changes

- `git mv frontend/src/__tests__/{App,routing}.test.tsx → frontend/src/`, beside `App.tsx`.
- Fix the relative imports `../` → `./` (matching how `App.tsx` imports its own siblings; this project uses relative imports, no path aliases).
- `src/__tests__/` is now empty and lapses from version control. No `.gitkeep` — `react.md` § 13 still documents it as the reserved home for flow tests; the first real flow test recreates it.
- **Docs reconciled** so the three no longer contradict on flow-test location:
  - `typescript.md` § 12 — co-location triggers on "targets a specific source file" (true even when the test mounts a broad tree); sole exception is cross-screen flows → `src/__tests__/`.
  - `frontend/CLAUDE.md` § Testing — "Unit/Integration" split reframed as "Co-located/Flow" to key off *what a test targets*, not its breadth.
  - `react.md` § 13 — already correct ("multi-screen sequences"); unchanged, now consistent with both.

**No config changes needed:** vitest `include: ['src/**/*.test.{ts,tsx}']` and its coverage `exclude` both already match the new paths; eslint's `**/*.test.{ts,tsx}` test override catches them. The vestigial `src/__tests__/**` globs in both configs stay valid for when the dir returns.

## Verification (reviewer can reproduce)

```sh
cd frontend
bun run test         # 30 files / 236 tests pass — incl. App.test.tsx & routing.test.tsx at their new paths
bun run typecheck    # clean (tsc -b, exit 0)
bun run lint         # 0 errors (90 pre-existing warnings in Session.tsx/time.ts, untouched here)
ls src/__tests__/    # gone — no longer tracked
```

Link-check (lychee, offline) on the two edited markdown files: 0 errors.

## Risk

Test-relocation + docs only — **no production code touched**, so no coverage impact (test files are excluded from coverage; the diff has zero coverable lines). The only behavioral risk would be the moved tests failing to resolve imports at the new path, which `bun run test` rules out.


frontend/src/App.routing.test.tsx
```diff
@@ -2,7 +2,7 @@ import { describe, expect, it, vi } from 'vitest';
 import { render, screen } from '@testing-library/react';
 import { RouterProvider, createMemoryRouter } from 'react-router-dom';
 import type { ReactNode } from 'react';
-import { AppProviders, routes } from '../App';
+import { AppProviders, routes } from './App';
```
brendanbyrne, line 5: 🔵 **Suggestion — file naming vs. a gap in the new § 12 wording** (non-blocking)

The PR's new doctrine — codified in `typescript.md` § 12 and `frontend/CLAUDE.md` § Testing — is *"a test that targets a specific source file lives next to that file."* This file imports `{ AppProviders, routes } from './App'`, so it targets `App.tsx` — same as `App.test.tsx`. `src/` now holds two test files targeting one source, and this one has no matching `routing.tsx`. The example the new rule leans on (`App.test.tsx` beside `App.tsx`) implies the `Foo.test.tsx ↔ Foo.tsx` pairing, so a reader who sees `routing.test.tsx` and looks for its source comes up empty.

This is a judgment call, not a violation — splitting App's tests into shell/error-boundary concerns (`App.test.tsx`) vs. authenticated route→page mapping (`routing.test.tsx`) is reasonable, and `routing.test.tsx` is arguably clearer than `App.routing.test.tsx`. Options if you want the pairing unambiguous: rename to `App.routing.test.tsx`, or keep as-is and accept concept-named test files.

The durable takeaway is a **docs gap**: § 12 now says co-location keys on "targets a specific source file" but is silent on whether the test *filename* must mirror that source's name — and this file lands exactly on that unaddressed edge. Worth one clarifying sentence in a future § 12 pass; does not need to block this PR.

brendanbyrne (commented):
## Code review — Approve (one non-blocking Suggestion)

**What this PR does:** Moves `App.test.tsx` and `routing.test.tsx` out of `src/__tests__/` to sit beside `App.tsx` (the module they both target), fixes their relative imports `../` → `./`, and reconciles the three docs that contradicted each other on test location (`typescript.md` § 12, `frontend/CLAUDE.md` § Testing, `react.md` § 13).

### Verified on the branch
- `bun run typecheck` → clean (`tsc -b`, exit 0)
- `bun run test` → **30 files / 236 tests pass**
- `bun run lint` → **0 errors** (90 warnings, all pre-existing in `Session.tsx`/`time.ts`, untouched here)
- `src/__tests__/` gone; **no stray `../` imports** remain in either moved file — the diff's hunks cover every relative import, so nothing dangles at the new depth.
- Config globs verified to still match: vitest `include: ['src/**/*.test.{ts,tsx}']` and its coverage `exclude`, plus the eslint `**/*.test.{ts,tsx}` test override, all match the new paths. The vestigial `src/__tests__/**` globs stay valid for when the dir returns.

### Design-doc accuracy
**No `design.md` change required** — `design.md` § Coverage & CI excludes test files from coverage and is unaffected by relocating them (doc-doesn't-cover). The three *standards* docs that needed reconciling were updated and verified mutually consistent: `typescript.md` § 12 (co-location triggers on "targets a specific source file," holds even for broad-tree mounts; sole exception = cross-screen flows → `src/__tests__/`); `frontend/CLAUDE.md` § Testing ("Unit/Integration" reframed to "Co-located/Flow"); `react.md` § 13 (unchanged, now consistent).

### What's good
- **Scope discipline.** `routing.test.tsx` was added *after* #206 was filed, so the issue's "only test in `__tests__/`" was stale. The PR caught that, correctly classified it as the same App-module shape (it mocks every page → a route→chunk *mapping* check, not a user journey), and confirmed the scope expansion before acting. Leaving it would have recreated the exact oversight #206 fixes.
- **Vestigial config globs left in place on purpose**, with comments, rather than churned out now and back in later — the right tradeoff, and no `.gitkeep` cruft.
- **The moved tests are genuinely strong** (unchanged here, but worth noting): `App.test.tsx` uses `toEqual(<RouteErrorFallback />)` with a comment on *why* not `toBeDefined()` (pins the route-scoped fallback so a regression to the global one fails), and verifies the route-scoped-vs-global error-boundary layering; `routing.test.tsx` is `it.each` over the route table. No assertion-free or status-only weak tests.
- **History entries are decision-log-appropriate** (relevant since main just adopted "history is a decision log, not a changelog" in a commit this branch predates): the `typescript.md` entry records the decision and its why (a contradiction removed, three docs reconciled), not a restatement of the diff.

### Notes (not findings)
- **Skipped the `test-quality-review` subagent**: the test files are *moved*, not logically changed — only import paths differ; the bodies are byte-identical and already passed review. Eyeballed both instead.
- **Branch is behind main** (predates the #220 `design.md` slim-down + the history-rule reframe). None of those commits touch this PR's four files, so no conflict risk — flagging only so not-rebasing is a conscious choice.
- **`link-check` runs repo-wide** on any `.md` edit, so a CI failure there could be pre-existing rot elsewhere, not this PR.

The one Suggestion is inline. `event: COMMENT` — Approve/Request-Changes stay with Brendan.

frontend/src/App.routing.test.tsx
brendanbyrne: Done in 60de3ca. Renamed `routing.test.tsx` → `App.routing.test.tsx` so the `App.tsx` pairing is explicit, and closed the docs gap: § 12 and `frontend/CLAUDE.md` § Testing now state the test filename mirrors its target source (`Foo.test.tsx`), with an optional concern infix (`App.routing.test.tsx`) when one source has several focused test files.

Considered merging the two back into `App.test.tsx` for a clean 1:1 pairing, but they need incompatible auth strategies — `App.routing` module-mocks `./hooks/useAuth` to force a signed-in user, while `App.test` exercises the *real* `AuthProvider` via MSW (401 on `/auth/refresh`). `vi.mock` is file-scoped and hoisted, so one file can't host both; the split is load-bearing, only the name needed fixing.

Tests 236/236, typecheck + lint clean.

Issue #195, opened 2026-05-18
Add CI workflow running frontend and backend lint/typecheck/clippy checks


brendanbyrne:
## Context

CI today runs only two workflows: `coverage.yml` (backend `cargo llvm-cov`) and `link-check.yml` (markdown links). **Nothing in CI runs frontend `lint`/`typecheck`, and nothing runs backend `clippy`/`fmt --check`.** Those checks exist only in the lefthook pre-commit hook (`lefthook.yml`), which is local-only and bypassable with `git commit --no-verify`.

This surfaced during PR-A1 of the frontend compliance plan ([#194](https://github.com/brendanbyrne/beerio-kart/pull/194) / #187). That plan deliberately starts strict-lint rules at `warn` so CI "stays green" — but there is currently no frontend lint job in CI for the warn-level strategy to keep green. The pre-commit hook is the only gate, which means a `--no-verify` commit (or a contributor with the hook uninstalled) lands unchecked.

A CI safety-net should enforce the same checks the pre-commit hook does, for both halves of the repo.

## Acceptance criteria

- [ ] Add a GitHub Actions workflow (e.g. `.github/workflows/checks.yml`) triggered on `pull_request` and `push` to `main`, mirroring the conventions of the existing workflows: `concurrency` group with `cancel-in-progress`, `permissions: contents: read`, `timeout-minutes`, `persist-credentials: false` on checkout, and **SHA-pinned actions** (dereference the specific version tag, not the floating major).
- [ ] **Frontend job:** install with Bun, then run `bun run lint` and `bun run typecheck` (and optionally `bun run build`). Run `bun run lint` as-is — it fails on ESLint *errors* only. Do **not** add `--max-warnings 0` while the frontend compliance plan is in flight: the plan intentionally keeps rules at `warn` until later PRs flip them back to `error`, and a zero-warnings gate would break CI for the whole compliance period.
- [ ] **Backend job:** run `cargo fmt --check` and `cargo clippy -- -D warnings` — the two backend checks the pre-commit hook runs that CI currently does not. Backend tests are already covered by `coverage.yml`, so this job need not re-run them.
- [ ] Keep the jobs reasonably fast (cache Bun / cargo artifacts as the existing `coverage.yml` does with `Swatinem/rust-cache`).

## Notes

- The backend half is in scope deliberately: `clippy`/`fmt --check` have the same "pre-commit-hook-only, bypassable" gap as the frontend checks.
[...]

PR #225, opened 2026-05-31
217: tighten test assertions — pin specific errors/codes; add lints

### LINKED ISSUE #217: Tighten test assertions: pin specific errors/codes in rejection tests
## Context

A full audit checked every Rust test in `backend/` and every Vitest test in `frontend/` for whether each test actually verifies what its name claims — i.e. the failure modes the literature attributes to LLM-written tests (assertion-free tests, tautological/over-mocked tests, weak/over-broad assertions, etc.).

**Verdict:** the suite is high quality. Of ~515 tests, ~63 are flagged, but ~50 are the **same single pattern** and only **1 is High severity**. None is a security hole. This issue tracks tightening that one systemic pattern plus a handful of one-offs, and adding two near-free lint rules so it can't reappear.

**The dominant pattern:** a test asserts that an operation *fails* (`assert!(result.is_err())`) or *returns an HTTP status* (`res.assert_status(409)`) but never pins **which** error — even though we maintain a machine-readable `ErrorCode` registry precisely so failures are distinguishable. These tests pass if the code fails *for the wrong reason*; they're looser than their names promise (not broken).

Full analysis + cited anti-pattern taxonomy lives in `docs/drafts/unit_test_audit.md` (currently a gitignored local draft — ask Cowork to promote it to `docs/research/` if a durable in-repo link is wanted).

## Acceptance criteria

**Remove the one empty test**
- [ ] `backend/src/services/sessions/races.rs::test_lazy_check_assertion` — a deliberate no-op (`// No-op test by design — this comment IS the assertion.`). Delete it (the invariant is already covered by `test_pending_excludes_closed_session` and siblings) and move its explanatory prose to a module / `get_pending_races` doc-comment.

**Pattern A — assert the specific error, not just `is_err()` / HTTP status**
- [ ] Service-layer unit tests on multi-gate functions assert the specific `Error` variant (and message substring where the name implies a code): in `runs/submission.rs` (not-participant, duplicate, session-closed, invalid-track-time, lap-sum-mismatch, lap-time-over-max, character-FK, drink-type-FK, non-owner-delete, delete-from-closed), `sessions/lifecycle.rs` (join-closed, double-join, invalid-ruleset, join-while-active, create-while-active), and `sessions/races.rs` (`next_track` host gate, closed-session, `skip_turn` no-race). Use `assert!(matches!(result, Err(Error::Conflict { .. })))` etc. **Leave as-is:** the single-failure-mode `validate_time_fields_*` tests and the raw-FK-violation insert.
- [ ] Integration tests assert the error-body `code`, not just the status — **auth/security first:** `tests/auth_middleware.rs` (refresh-as-access → `token_invalid`) and `tests/auth_integration.rs` (refresh-as-Bearer / `type:refresh` / **expired** → `token_expired`; also collapse the duplicate refresh-as-access pair). Then `tests/api_integration.rs` (nonexistent user/drink-type 404s, invalid-FK / all-or-nothing / empty-name 400s, other-users 403), `tests/session_integration.rs` (join-closed 409, invalid-ruleset 400, join-twice 409, nonexistent-session 404, skip: unknown 404 / already-submitted 409 / after-leaving 403), the four register-validation 400s, and the two byte-identical admin-403 tests.

**Pattern B — frontend: assert a refresh happened, not just that invalidate was called**
- [ ] `frontend/src/pages/Home.test.tsx` and `Session.test.tsx` (join / next-track / skip / run-submitted): the `vi.spyOn(queryClient,'invalidateQueries')`-as-sole-assertion tests should assert an observable outcome (UI/data actually refreshed). The leave→navigation assertion is already fine.

**Pin serialization wire-format where round-trips don't**
- [ ] `backend/src/domain/enums.rs` — `SessionStatus` / `SessionRuleset` sea_orm round-trips are invariant under a symmetric `string_value` typo; add literal pins (e.g. `assert_eq!(SessionStatus::Active.to_value(), "active".into())` and a multi-word ruleset).
- [ ] `backend/tests/auth_integration.rs` — the "rotated cookie" test should `assert_ne!(new_cookie, original_cookie)` (rotation is currently unverified).

**Low-severity cleanups (opportunistic)**
- [ ] `domain/ids.rs::test_distinct_id_types_do_not_unify` (compile-time property tested at runtime — move to `trybuild` or rename); `domain/numeric.rs` lap-sum proptest (recomputes expected with the SUT's own arithmetic — add a perturbed-total err case); `shutdown.rs::test_supervised_catches_panic_and_returns` (assert the panic log line via `tracing_test`); `services/notifications.rs` JSON round-trip duplicated by its sibling; `components/SubmitButton.test.tsx` misleading "pending fallback" title; `__tests__/App.test.tsx` `toBeDefined()` on `errorElement` (assert identity instead).

**Lint to prevent regressions**
- [ ] Rust: add `#![warn(clippy::assertions_on_result_states)]` to the backend crate root (flags `assert!(r.is_ok()/is_err())`; it's in clippy's `restriction` group, so it must be opted in).
- [ ] Frontend: adopt `eslint-plugin-vitest` `recommended` + `eslint-plugin-testing-library` React config (the suite already passes these, so it just locks the state in).
- [ ] Optional: a one-line `docs/coding-standards/testing.md` rule — *"a test named for a specific failure must assert the specific `code`/variant, not just `is_err()`/the HTTP status."*

## References
- `docs/drafts/unit_test_audit.md` — full audit: cited anti-pattern taxonomy (Part 1), findings by severity (Part 3), per-file coverage (Part 4), and the false-positive traps that were checked and cleared (Part 5).
- In-repo templates for the right assertion style: `backend/src/error.rs`, `backend/src/extract.rs`, `backend/tests/db_timeout.rs` (they assert variant + `code` + body).

brendanbyrne:
## What & why

Closes #217. A backend+frontend audit found the suite is high quality but shares one pervasive habit: rejection tests that assert *that* an operation failed (`is_err()` / `assert_status(409)`) without pinning *which* error — even though we keep a machine-readable `ErrorCode` registry so failures are distinguishable. Those tests pass if the code fails at the wrong gate. This PR tightens that pattern plus a handful of one-offs, and turns on lints so it can't reappear.

No production behavior changes — this is tests + lint config + docs. (Two benign production tweaks: `#[derive(Debug)]` on the private `ValidatedRunTimes`, and the crate-root lint attr.)

## Backend

- **Pattern A — pin the failure.** Service-layer tests (`runs/submission.rs`, `sessions/lifecycle.rs`, `sessions/races.rs`) assert the specific `Error` variant via `err.code()` (+ a message substring where the code is generic). Integration tests (`auth_middleware`, `auth_integration`, `api_integration`, `session_integration`) assert `body["code"]`, security-adjacent auth first.
- **H1 — empty test.** Deleted `test_lazy_check_assertion` (a deliberate no-op); its prose moved into `get_pending_races`'s doc-comment. The invariant is already covered by `test_pending_excludes_closed_session` and siblings.
- **Serialization pins.** `SessionStatus` / `SessionRuleset` / `DrinkCategory` / `RunFlagReason` now pin literal `to_value()` strings, not just self-round-trips (a symmetric `string_value` typo would survive a round-trip).
- **Lows.** Perturbed-total proptest for `assert_lap_sum`; renamed the misnamed `test_distinct_id_types_do_not_unify` (it tested a runtime tautology, not the compile-time property); `supervised()` now asserts the panic log line via `tracing_test`; merged the duplicate notification round-trip into its wire-shape sibling.
- **Lint.** `#![warn(clippy::assertions_on_result_states)]` at the crate root; all remaining `is_ok()`/`is_err()` asserts converted to `.unwrap()`/`.unwrap_err()` (`Password` keeps no `Debug`, so its rejection tests use `.err().expect(...)`). lefthook's `rust-clippy` now runs `--all-targets` so the lint actually reaches `#[cfg(test)]` modules (it's a no-op otherwise).

## Frontend

- **Pattern B — outcome over choreography.** The `invalidateQueries`-spy tests (Home create; Session join / next-track / skip / run-submitted) now assert the observable refresh via **stateful MSW**: navigation to the new session, Join→Leave, the new track name, the "Your Time" card. The spy assertions are kept only as a secondary check of *which* keys.
- **SubmitButton.** The "pending fallback" test rendered outside a form (`pending` always false), so it never exercised the branch it named. It now renders inside a form with a held action so `pending` is true.
- **Lint.** Adopted `@vitest/eslint-plugin` recommended + `eslint-plugin-testing-library` flat React config, scoped to test files.

## Docs

- New `docs/coding-standards/testing.md` — cross-cutting assertion rules (pin the specific error/`code`; assert the observable outcome; pin the wire literal) + the lints that enforce them. Indexed in the coding-standards README.
- **The source audit is now in-repo.** The backend+frontend test audit that produced these rules was a gitignored draft, so `testing.md` could only point at the issue for the findings. It's promoted to `docs/research/unit-test-audit.md` (taxonomy, per-file findings, cleared false-positives); `testing.md` and the README now link it instead of #217. (`docs/drafts/` is `--exclude-path`'d from the lychee link-check, so the reference only resolves once the file is tracked.)

## Follow-up commits (added during review)

Three small, scoped additions spun out of a "what's still ignored by the linters / what did the deviations leave untested?" pass:

1. **Lint ratchet** (`57d7659`) — flipped 8 type-checked eslint rules `warn → error` because they had **zero** violations (`no-unsafe-{assignment,member-access,argument,call,return}`, `no-unnecessary-type-assertion`, `no-deprecated`, `prefer-regexp-exec`); added `coverage` to `globalIgnores`; merged duplicate `useGameData` imports. eslint warnings **113 → 105**, errors stay **0**. Left deliberately parked (still have offenders): the Rust `needless_raw_string_hashes` allow (**11**) and `too_many_lines` (**3**) — both remain "Phase H scaffolding".
2. **Guard coverage** (`658bc79`) — see decision 2 below.
3. **Audit promotion** (`29a25a4`) — see Docs above.

Worth noting separately: CI runs neither clippy nor eslint nor `tsc` today — they're enforced only in the lefthook pre-commit/pre-push hooks, with the CI lint/typecheck jobs tracked in #195.

## Decisions worth a look (deviations from the issue spec)

- **Refresh-cookie rotation is asserted *functionally*, not by `assert_ne!`.** The issue suggested `assert_ne!(new_cookie, original_cookie)`, but `RefreshClaims` carries no `jti` and the refresh path reuses the same version — two tokens minted in the same wall-clock second are byte-identical, so `assert_ne!` would be **flaky**. The test instead asserts the rotated cookie is itself usable to refresh. Whether value-rotation/reuse-detection (which needs a `jti`) is actually a goal is **filed as #226**.
- **The "refresh-as-access" rejection fires at `decode::<AccessClaims>` (message `"Invalid token"`), not the explicit `token_type` guard.** A refresh token lacks the `username` claim, so it fails one gate *earlier*; the test (and its comment) now match reality. That left the `token_type` guard unreachable-via-real-tokens and **untested**, so this PR adds a crafted-token test (`658bc79`): it forges an access-shaped token with `token_type: "refresh"` — the only way to reach the guard — and asserts `code: token_invalid` + message `"Invalid token type"`, proving the guard itself fired. (The mirror guard in the refresh route is similarly defensive; not separately covered.)
- **Two eslint rules start at `warn`, not `error`** (`vitest/no-conditional-expect`, `testing-library/no-node-access`) — they fire on legitimate existing patterns (the `try { … expect.unreachable() } catch { expect(err) }` thrown-error shape; raw-node drag simulation for the custom DQ slider). Flipping them needs the underlying rework, **filed as #227**.

## Verification

```bash
# Backend — tests, lint (as the hook now runs it), patch coverage
cargo test
cargo llvm-cov --workspace --summary-only      # patch on this diff measured at 98.3%

# Frontend — tests + lint (0 errors; 105 pre-existing compliance warnings remain)
cd frontend && bun run test && bun run lint
```

All green locally: backend **378** tests, frontend 236 tests, clippy `--all-targets` clean, eslint **0 errors / 105 warnings**, link-check clean (lychee: 65 links OK across the changed docs).


backend/src/services/auth.rs
```diff
@@ -449,13 +449,13 @@ mod tests {
             rate_limit_per_minute: 60,
         });
 
-        assert!(validate_access_token(&token, &wrong_config).is_err());
+        validate_access_token(&token, &wrong_config).unwrap_err();
```
brendanbyrne: 🔵 **Suggestion** — `validate_access_token` returns `jsonwebtoken::errors::Error`, not the project `Error`/`ErrorCode`, so `.unwrap_err()` here (and at line 458, `test_validate_token_garbage_input`) is the most this layer can do — but it still discards *which* failure. The names imply specific kinds ("wrong secret" → `InvalidSignature`; "garbage input" → a parse/`InvalidToken` kind) that aren't pinned, so a regression that failed for a different reason would still pass. Low value to fix, since the security-relevant discrimination (`token_expired` vs `token_invalid`) is already pinned at the boundary that matters in `auth_middleware.rs` / `auth_integration.rs` via `body["code"]`. If you want the unit layer to match its name: `assert!(matches!(err.kind(), ErrorKind::InvalidSignature))` / `ErrorKind::InvalidToken`. Fine to leave as-is given the integration coverage — per `testing.md` § 1, single-failure-mode is the exception and these are close to it.

backend/tests/api_integration.rs
```diff
@@ -495,6 +499,11 @@ async fn test_update_user_invalid_fk_returns_400() {
         }))
         .await;
     res.assert_status(axum::http::StatusCode::BAD_REQUEST);
+    // The FK check (not a type-parse failure) is what rejects: the code is the
+    // generic bad_request, and the message names the offending entity.
+    let body: Value = res.json();
+    assert_eq!(body["code"], "bad_request");
+    assert!(body["error"].as_str().unwrap().contains("character"));
```
brendanbyrne: 🔵 **Suggestion** (cosmetic) — the substring `"character"` does distinguish the FK gate from the sibling all-or-nothing 400 (which carries no entity name), so this satisfies `testing.md` § 1. But `"character"` is a slightly loose token; the actual message is `"Invalid character_id"` (`helpers.rs` `require_exists`), so `contains("Invalid character_id")` would pin the gate more tightly with no downside.

docs/coding-standards/README.md
```diff
@@ -51,4 +55,5 @@ Some rules are checked by tooling on every CI run (`cargo fmt --check`, `cargo c
...
+- 2026-05-31 — Added `testing.md` (cross-cutting test-assertion rules: pin the specific error / `code` rather than `is_err` / the HTTP status, assert the observable outcome rather than an interaction spy, pin the serialized wire literal) plus the lints that hold them — `#![warn(clippy::assertions_on_result_states)]` with lefthook `cargo clippy --all-targets`, and the `@vitest/eslint-plugin` + `eslint-plugin-testing-library` presets. Companion to Issue [#217](https://github.com/brendanbyrne/beerio-kart/issues/217)'s suite-wide tightening pass (the dominant gap was rejection tests that asserted *that* an op failed, not *which* error).
```
brendanbyrne: 🔵 **Suggestion** — this new `2026-05-31` history entry is inserted *between* the `2026-05-16` and `2026-05-18` entries, breaking the otherwise-chronological order. Move it below the `2026-05-18` line so the log stays in date order (newest last). Minor doc hygiene, not blocking.

brendanbyrne (commented):
## Code review — test-assertion tightening (#217)

**Reviewer assessment: Approve** (I can't set the APPROVE state — that's Brendan's — so this posts as a comment). No 🔴 Critical or 🟡 Important findings; three 🔵 Suggestions inline, none blocking.

**What this PR does:** Converts the suite's pervasive `is_err()` / status-only rejection tests to pin the specific `Error` variant / body `code` (+ a message substring where the code is generic), adds `#![warn(clippy::assertions_on_result_states)]` + lefthook `--all-targets` and the vitest/testing-library eslint presets to hold the pattern, converts the frontend `invalidateQueries`-spy tests to assert the observable refetched UI via stateful MSW, and adds `docs/coding-standards/testing.md`. No production behavior change (two benign tweaks: `#[derive(Debug)]` on a private `ValidatedRunTimes`, the crate-root lint attr).

**Areas covered:** `testing.md` (cross-cutting), `rust.md` § 7, `react.md` § 13 / `typescript.md` § 12, `api-contract.md` § 7 (error registry), `design.md` § Coverage & CI — plus the `test-quality-review` skill run by a subagent over all 17 changed test files.

**Verified:**
- `cargo clippy --all-targets --all-features -- -D warnings` — clean (exit 0). The lint reaches `#[cfg(test)]` only because of the `--all-targets` change; confirmed it's not a no-op.
- `cargo test` — pass (exit 0).
- `bun run lint` — 0 errors (113 pre-existing warnings, incl. the two new `warn`-level rules firing on pre-existing patterns as documented); `tsc -b` clean.
- Backend rejection tests pin the gate the test name claims: every `to_string().contains(...)` / `body["error"]` substring checked against the real strings in `error.rs` / `helpers.rs` / the services; the two same-status 409s (`session_closed` vs generic `conflict`) and the two `admin_required` 403s are now distinguished — the rule's motivating case.
- Frontend Pattern B genuinely flips MSW server state on each mutation and asserts the refetched UI (new heading, Join→Leave, "Your Time" card); confirmed each outcome assertion would fail if the invalidation were removed — not a false-outcome assertion.
- `prop_assume!` in the new lap-sum negative proptest keeps ~99.9% of cases (≈50/50 +/- offsets) — not gutted.
- `logs_contain("task panicked")` matches the exact string `supervised()` emits.
- `test_lazy_check_assertion` (a genuinely assertion-free no-op) was deleted — good.

**Design-doc accuracy:** no changes required. This is a coding-standards addition (new `testing.md`, indexed in `coding-standards/README.md`), not an architecture / data-model / API-surface change; `design.md` is correctly untouched.

**What's good:** the lint + `--all-targets` is the durable structural win (makes the regression un-reintroducible, not just a one-time sweep); the refresh-rotation test correctly avoids a flaky `assert_ne!` on byte-identical same-second tokens and asserts the *functional* property instead, with the reasoning documented; the `enums.rs` literal wire pins alongside the round-trips are exactly `testing.md` rule 3; the honest renames (`test_id_types_construction_smoke`) stop a test from claiming a compile-time property it only smoke-tests at runtime.

backend/src/services/auth.rs
brendanbyrne: Partially applied in b120fec. Pinned the **wrong-secret** test to `ErrorKind::InvalidSignature` — that's a stable cause that matches the name.

Left **garbage-input** on `.unwrap_err()` (with a comment): I checked the actual kind and it's `Base64(InvalidByte(3, 46))` — i.e. an artifact of the `.` in this particular string, not the `InvalidToken` the name might imply. A different garbage string would give a different kind, so pinning it would over-fit. That one really is the single-failure-mode exception in `testing.md` § 1.

backend/tests/api_integration.rs
brendanbyrne: Done in b120fec — tightened to `contains("Invalid character_id")` (the full `require_exists` message). Agreed, no downside.

docs/coding-standards/README.md
brendanbyrne: Fixed in b120fec — moved the 2026-05-31 entry below 2026-05-18 so the log is chronological (newest last).

brendanbyrne (commented):
## Re-review (@ b120fec) — all three suggestions resolved ✅

Follow-up to [my earlier review](https://github.com/brendanbyrne/beerio-kart/pull/225#pullrequestreview-4397171346). The 4 new commits address every point; I re-ran the checks on the new head and verified each fix against source.

**Prior suggestions — resolved:**
1. ✅ `auth.rs` — `test_validate_access_token_with_wrong_secret` now pins `ErrorKind::InvalidSignature`; confirmed that's the real kind (`validate_access_token` does genuine signature validation, doesn't disable it). The garbage-input test is deliberately left as a bare `.unwrap_err()` with a comment invoking `testing.md` § 1's single-failure-mode exception — a defensible call.
2. ✅ `api_integration.rs` — tightened to `contains("Invalid character_id")`; verified that's the exact message `require_exists` emits, and `character` is the first FK checked so it's the gate under test.
3. ✅ `coding-standards/README.md` — the `2026-05-31` entry now follows `2026-05-18`; chronological order restored.

**New additions — both sound and valuable:**
- **Crafted-token guard test** (`auth_middleware.rs`) — forges an *access-shaped* JWT with `token_type: "refresh"` signed with the real key, so it decodes cleanly and execution actually reaches the `token_type` guard the real-refresh-token test can't reach (that one dies at `decode::<AccessClaims>` for lack of `username`). Verified the crafted claims match `AccessClaims` field-for-field, and it asserts the real `token_invalid` + `"Invalid token type"`, while the existing test now asserts the complementary `"Invalid token"`. This closes a genuine gap — the explicit guard was previously never exercised. 
- **eslint ratchet** (`eslint.config.js`) — flips the five `no-unsafe-*` rules + `no-unnecessary-type-assertion`/`no-deprecated`/`prefer-regexp-exec` to `error` (suite verified clean at `error`), correctly leaves `vitest/no-conditional-expect` and `testing-library/no-node-access` at `warn` (the only real `no-node-access` offender is RunEntrySheet's drag sim, as the comment states), and adds `coverage` to `globalIgnores`. Honest "ratchet only tightens" framing.
- Audit promoted to `docs/research/unit-test-audit.md` and relinked from `testing.md` (`[audit]` ref resolves; ref-link block intact).

**Checks on b120fec — all green:** `cargo clippy --all-targets --all-features -- -D warnings` (0 warnings), `cargo test` (0 failed), `bun run lint` (0 errors; warnings 113→105), `tsc -b` clean.

Two trivial comment-text nits inline — cosmetic, not worth a round-trip on their own. **Approve** (Brendan sets the state).

backend/src/services/auth.rs
```diff
@@ -449,13 +449,24 @@ mod tests {
...
     fn test_validate_token_garbage_input() {
         let config = test_config();
-        assert!(validate_access_token("not.a.valid.jwt", &config).is_err());
+        // Single-failure-mode case (testing.md § 1's exception): any malformed
+        // token is rejected, and the exact jsonwebtoken ErrorKind here is an
+        // artifact of this particular string (a Base64 decode error on the `.`),
+        // not a behavioral contract worth pinning — unlike the wrong-secret test
+        // above, where InvalidSignature is the stable, name-matching cause.
+        validate_access_token("not.a.valid.jwt", &config).unwrap_err();
     }
```
brendanbyrne, line 470: 🔵 **Nit** (cosmetic, optional) — for the 4-segment string `"not.a.valid.jwt"`, jsonwebtoken more likely returns `ErrorKind::InvalidToken` (wrong segment count — JWT expects 3) than a Base64 decode error. Since the test intentionally doesn't assert the kind, this is just comment imprecision, not a behavior issue — leave it or reword to "a parse error (`InvalidToken`/Base64), not a behavioral contract."

frontend/eslint.config.js
```diff
@@ -156,6 +163,39 @@ export default defineConfig([
...
+    // docs/coding-standards/testing.md.
+    ...vitest.configs.recommended,
+    files: ['**/*.test.{ts,tsx}', 'src/__tests__/**/*.{ts,tsx}'],
+    rules: {
+      ...vitest.configs.recommended.rules,
+      // Started at `warn`, not `error`, per this file's preamble: it fires on
+      // the legitimate `try { …; expect.unreachable() } catch { expect(err) }`
+      // shape used to assert on a thrown error's nested fields (result.test.ts,
+      // runs.test.ts). Flip to `error` if those convert to `.rejects`.
+      'vitest/no-conditional-expect': 'warn',
```
brendanbyrne, line 183: 🔵 **Nit** (cosmetic, optional) — `runs.test.ts` actually uses `.rejects.toThrow(...)`, not the `try { … expect.unreachable() } catch { expect(err) }` shape; only `result.test.ts` uses that pattern. The comment over-attributes but errs safe (keeps the rule at `warn`). Drop `runs.test.ts` from the list if you touch this.

backend/src/services/auth.rs
brendanbyrne: Checked this empirically rather than guessing — against the locked `jsonwebtoken`, `"not.a.valid.jwt"` returns `Base64(InvalidByte(3, 46))` (the `.` at byte 3 trips base64 decode of the header segment), not `InvalidToken`. So the comment is accurate as written; leaving it. (The `InvalidToken`-on-wrong-segment-count path is plausible for *other* malformed inputs, just not this one — which is exactly why the test doesn't pin the kind.)

frontend/eslint.config.js
brendanbyrne: Half right — fixed in 49b118e, but `runs.test.ts` stays in the list. Its `.rejects.toThrow(...)` calls (L59/105/126) indeed don't trigger the rule, but it has 3 *other* offenders (L147/148/164) that do: discriminated-`Result` narrowing —

```ts
expect(result.ok).toBe(false);
if (!result.ok) {
  expect(result.error.code).toBe('internal'); // expect inside `if` → no-conditional-expect
}
```

So dropping it would be wrong (the rule can't flip to `error` until those are reworked too). What was inaccurate was the *shape* attribution — the comment now describes both `result.test.ts`'s `try/expect.unreachable()/catch` and `runs.test.ts`'s Result-narrowing, and points at #227.

PR #230, opened 2026-06-13
226: refresh-token rotation with reuse detection (backend)

### LINKED ISSUE #226: Implement refresh-token rotation with reuse detection (token families + `jti`)
## Context

Surfaced reviewing PR #225 (Issue #217). The rotated-cookie test
(`backend/tests/auth_integration.rs`) was changed from the issue's suggested
`assert_ne!(new_cookie, original_cookie)` to a **functional** rotation assertion
(the rotated cookie is itself usable to refresh), because `assert_ne!` would be
**flaky**:

- `RefreshClaims` (`backend/src/services/auth.rs`) is `{ sub, refresh_token_version, exp, iat, token_type }` — there is **no `jti`** (unique token id).
- The refresh path reuses the same `refresh_token_version` and the `exp`/`iat` are second-granularity, so **two refresh tokens minted in the same wall-clock second are byte-identical**.

## The question

Is **value-rotation / reuse-detection** an intended security property of refresh
tokens? Refresh-token rotation is normally a theft-mitigation: when the
legitimate client refreshes, a previously-issued (possibly stolen) refresh token
should become unusable, and reuse of an old token should be detectable.

Today neither holds: the token value can be identical across refreshes, and
there's no per-token identifier to detect replay.

## If yes (rotation is a security goal)

- Add a `jti` (unique id) to `RefreshClaims` so every minted token is distinct.
- Consider refresh-token **reuse detection** (persist the issued `jti`, or bump
  `refresh_token_version` on each refresh so the prior token is invalidated).
- Then the rotated-cookie test can pin the stronger property
  (`assert_ne!` on the value), not just functional rotation.

## If no (rotation is just "re-issue with a fresh expiry")

- Document that decision (e.g. in `api-contract.md` §§ 1.1/4 or ADR 0031) so the
  looser, functional-only test is understood as intentional rather than a gap.

## References

- `backend/src/services/auth.rs` — `RefreshClaims`, `create_refresh_token`
- `backend/src/routes/auth.rs` — the refresh handler (token rotation)
- `backend/tests/auth_integration.rs` — the rotated-cookie test (functional assertion)

_Follow-up from PR #225 review._

brendanbyrne:
## Summary

Backend half of #226 — refresh-token **rotation with reuse detection** (token families + `jti`), implementing [ADR-0040](../blob/main/docs/decisions/0040-refresh-token-reuse-detection.md) and [RFC 9700](https://datatracker.ietf.org/doc/rfc9700/) § 4.14.

Today two refresh tokens minted in the same second are byte-identical (`RefreshClaims` has no `jti`), so a stolen refresh token survives a legitimate refresh and reuse is undetectable. This adds per-token state: each login starts a **family**; each refresh mints a successor token (new `jti`) in that family and marks its predecessor *used*; replaying a used token past a grace window revokes the whole family.

## What the refresh endpoint now does

| Presented token | Result |
|---|---|
| Live (`used_at IS NULL`) | `200` — rotate: mark used, insert successor, return a **new** cookie value (distinct `jti`) |
| Used, **within** grace (`refresh_grace_seconds`, default 10s) | `200` — reissue the family's live successor (absorbs concurrent/retried refreshes) |
| Used, **past** grace | `401 token_reuse_detected` — revoke the whole family (its live token stops working too) |
| Revoked / unknown / expired | `401 token_invalid` / `token_expired` (unchanged; the global `refresh_token_version` still revokes everything) |

## Changes

- **Migration** — new `refresh_tokens` table (`id`=`jti`, `user_id` FK ON DELETE CASCADE, `family_id`, `used_at`, `expires_at`), **edited into the consolidated migration in place** per `seaorm.md` § 5 (prelaunch convention) rather than the issue spec's append-only note — flagged at planning, this is the deviation.
- **Entity** — hand-written `refresh_tokens` + behavior; wired into `mod`/`prelude`/`schema_drift`.
- **`services/auth.rs`** — `jti` + `family_id` on `RefreshClaims` (still no `username`, so a refresh token still can't decode as `AccessClaims`); `create_refresh_token` takes them; new `prune_refresh_tokens`.
- **`routes/auth.rs`** — register/login start a family; refresh rewritten (live / grace / lost-race / reuse, in a transaction with a conditional `UPDATE … WHERE used_at IS NULL` as the race primitive); logout & password-change clear the user's families alongside the existing version bump.
- **`error.rs`** — `token_reuse_detected` code; **`config.rs`** — `refresh_grace_seconds` (default 10); **`main.rs`** — hourly `prune_refresh_tokens` sweep.
- **Docs** — `data-model.md` (the table), `api-contract.md` §§ 1.1 / 4 / 7.

## Manual test steps

Run the backend from the repo root (`cargo run` — loads `.env`, which must set `JWT_SECRET`). The grace window defaults to 10s and can't be set below 1, so this demo shows the grace reissue *and* reuse detection via timing:

```bash
BASE=http://localhost:3000/api/v1

# 1. Register — save the refresh cookie, keep a copy of the ORIGINAL.
curl -s -c jar.txt -X POST $BASE/auth/register \
  -d '{"username":"reuse-demo","password":"password123"}' | jq .access_token
cp jar.txt original.txt

# 2. Rotate: refresh with the original cookie. jar.txt now holds a NEW cookie value.
curl -s -c jar.txt -b jar.txt -X POST $BASE/auth/refresh | jq .access_token

# 3. Replay the ORIGINAL immediately (within the 10s grace) -> 200, reissued, no logout.
curl -s -b original.txt -X POST $BASE/auth/refresh -w '\nHTTP %{http_code}\n'

# 4. Wait out the grace window, replay the ORIGINAL again -> 401 token_reuse_detected.
sleep 11
curl -s -b original.txt -X POST $BASE/auth/refresh -w '\nHTTP %{http_code}\n'
#   -> {"error":"Refresh token reuse detected","code":"token_reuse_detected"}  HTTP 401

# 5. Reuse revoked the whole family — the rotated (live) cookie now fails too.
curl -s -b jar.txt -X POST $BASE/auth/refresh -w '\nHTTP %{http_code}\n'
#   -> HTTP 401, code "token_invalid" (Refresh token has been revoked)
```

Multi-device check: register, then `POST /auth/login` again as the same user (a second family); triggering reuse on one family leaves the other still refreshing.

## Automated tests

`cargo test --test auth_integration` (black-box) + unit tests cover:
- `test_refresh_reuse_detected_revokes_family` (grace=0)
- `test_refresh_within_grace_window_reissues_instead_of_revoking`
- `test_two_families_coexist_and_reuse_revokes_only_one`
- `test_refresh_with_valid_cookie_…_rotated_cookie` now also asserts the cookie **value** differs (the `jti` makes tokens distinct — the old flaky-`assert_ne!` concern is gone)
- `test_prune_refresh_tokens_deletes_only_expired_rows`

Full suite green (298 lib unit + integration); `cargo clippy --all-targets` and `rustfmt` clean. New uncovered lines are only defensive error-propagation arms; `entities/`, `migration/`, `main.rs` are codecov-ignored, so the patch gate sees the well-covered handler/service/error code.

## Follow-up (separate PR, off `main` after this merges)

Frontend: single-flight silent refresh (concurrent 401s queue behind one in-flight refresh) + handle `token_reuse_detected` as a hard "signed out for security" logout. This is why this PR does **not** `Closes #226` — the issue's frontend AC lands with that PR.

Refs #226 · [ADR-0040](../blob/main/docs/decisions/0040-refresh-token-reuse-detection.md)

### DIFF
--- a/backend/src/routes/auth.rs
+++ b/backend/src/routes/auth.rs
@@ -3,21 +3,24 @@ use axum::{
     http::{HeaderMap, StatusCode, header},
     response::IntoResponse,
 };
+use chrono::{NaiveDateTime, TimeDelta, Utc};
 use sea_orm::{
-    ActiveModelTrait, ActiveValue::NotSet, ColumnTrait, EntityTrait, IntoActiveModel, QueryFilter,
-    Set,
+    ActiveModelTrait, ActiveValue::NotSet, ColumnTrait, ConnectionTrait, EntityTrait,
+    IntoActiveModel, QueryFilter, QueryOrder, Set, TransactionTrait, sea_query::Expr,
 };
 use serde::{Deserialize, Serialize};
+use uuid::Uuid;
 
 use crate::{
     AppState,
+    config::Config,
     domain::{Password, PasswordHash, UserId, Username},
-    entities::users,
+    entities::{refresh_tokens, users},
     error::Error,
     extract::Json,
     middleware::auth::User,
     services::auth as auth_service,
-    timeout::db_query,
+    timeout::{db_query, db_txn},
 };
 
 // ── Request / Response types ────────────────────────────────────────
@@ -106,6 +109,79 @@ fn extract_refresh_cookie(headers: &HeaderMap) -> Option<String> {
         })
 }
 
+/// Mint a refresh token in the given family and persist its `refresh_tokens`
+/// row as the live tip (`used_at = NULL`). Generates a fresh `jti`, so the
+/// minted token is byte-distinct from every other (ADR-0040). The row's
+/// `expires_at` mirrors the JWT expiry so the prune sweep can drop it once the
+/// JWT is dead. Generic over the connection so it works on a plain handle
+/// (register / login / password-change start a new family) or inside the
+/// refresh transaction (a successor in an existing family).
+async fn mint_refresh_in_family<C: ConnectionTrait>(
+    conn: &C,
+    user_id: &UserId,
+    refresh_token_version: i32,
+    family_id: &str,
+    config: &Config,
+) -> Result<String, Error> {
+    let jti = Uuid::new_v4().to_string();
+    let expires_at = (Utc::now() + TimeDelta::days(config.jwt_refresh_expiry_days)).naive_utc();
+
+    // `created_at` is stamped by `refresh_tokens::ActiveModelBehavior`.
+    db_query(
+        refresh_tokens::ActiveModel {
+            id: Set(jti.clone()),
+            user_id: Set(user_id.into()),
+            family_id: Set(family_id.to_string()),
+            used_at: Set(None),
+            expires_at: Set(expires_at),
+            created_at: NotSet,
+        }
+        .insert(conn),
+    )
+    .await?;
+
+    Ok(auth_service::create_refresh_token(
+        user_id,
+        refresh_token_version,
+        family_id,
+        &jti,
+        config,
+    )?)
+}
+
+/// Re-mint a JWT for the family's current live tip, without rotating. Used when
+/// a refresh loses a concurrent race or arrives within the grace window: the
+/// successor already exists, so we hand back a token for it rather than
+/// revoking the family on a false-positive reuse. No live tip means the family
+/// was revoked — surfaced as `token_invalid`.
+async fn reissue_from_family_tip<C: ConnectionTrait>(
+    conn: &C,
+    user_id: &UserId,
+    refresh_token_version: i32,
+    family_id: &str,
+    now: NaiveDateTime,
+    config: &Config,
+) -> Result<String, Error> {
+    let tip = db_query(
+        refresh_tokens::Entity::find()
+            .filter(refresh_tokens::Column::FamilyId.eq(family_id))
+            .filter(refresh_tokens::Column::UsedAt.is_null())
+            .filter(refresh_tokens::Column::ExpiresAt.gt(now))
+            .order_by_desc(refresh_tokens::Column::CreatedAt)
+            .one(conn),
+    )
+    .await?
+    .ok_or_else(|| Error::token_invalid("Refresh token has been revoked"))?;
+
+    Ok(auth_service::create_refresh_token(
+        user_id,
+        refresh_token_version,
+        family_id,
+        &tip.id,
+        config,
+    )?)
+}
+
 // ── Handlers ────────────────────────────────────────────────────────
 
 /// POST /api/v1/auth/register
@@ -167,9 +243,11 @@ pub async fn register(
 
     db_query(new_user.insert(&state.db)).await?;
 
-    // Generate tokens
+    // Generate tokens. Registration starts a fresh token family (ADR-0040).
     let access_token = auth_service::create_access_token(&user_id, &username, &state.config)?;
-    let refresh_token = auth_service::create_refresh_token(&user_id, 0, &state.config)?;
+    let family_id = Uuid::new_v4().to_string();
+    let refresh_token =
+        mint_refresh_in_family(&state.db, &user_id, 0, &family_id, &state.config).await?;
     let headers = make_refresh_headers(&refresh_token, &state.config)?;
 
     Ok((
@@ -248,11 +326,19 @@ pub async fn login(
     // (mirrors the PasswordHash / ImagePath from_db pattern).
     let stored_username = Username::from_db(user.username, "users.username")?;
 
-    // Generate tokens
+    // Generate tokens. Each login starts a new token family (ADR-0040), so a
+    // user's devices each get an independent family that rotates on its own.
     let access_token =
         auth_service::create_access_token(&user_id, &stored_username, &state.config)?;
-    let refresh_token =
-        auth_service::create_refresh_token(&user_id, user.refresh_token_version, &state.config)?;
+    let family_id = Uuid::new_v4().to_string();
+    let refresh_token = mint_refresh_in_family(
+        &state.db,
+        &user_id,
+        user.refresh_token_version,
+        &family_id,
+        &state.config,
+    )
+    .await?;
     let headers = make_refresh_headers(&refresh_token, &state.config)?;
 
     Ok((
@@ -270,19 +356,25 @@ pub async fn login(
 /// POST /api/v1/auth/refresh
 ///
 /// Reads the refresh token from the `HttpOnly` cookie (NOT from the request body
-/// or Authorization header). If valid and the `refresh_token_version` matches
-/// the DB, returns a new access token and rotates the refresh cookie.
+/// or Authorization header), then rotates it with **reuse detection** (ADR-0040):
 ///
-/// "Rotation" means issuing a fresh refresh JWT with a fresh expiry on every
-/// successful refresh. This extends the session window without bumping the
-/// version (which would invalidate other devices).
+/// 1. Validate the JWT (signature + expiry) and the `refresh_token_version`
+///    (the global per-user revoke-all, bumped on logout / password change).
+/// 2. Look the token's row up by `jti`. A **live** row (`used_at IS NULL`) is
+///    rotated: it's marked used and a successor is minted in the same family.
+///    An **already-used** row presented past the grace window is **reuse** —
+///    the whole family is revoked and the call fails `token_reuse_detected`.
+/// 3. A used row *within* the grace window, or a lost rotation race, reissues
+///    the family's live successor instead of revoking — the backstop against
+///    spuriously logging out concurrent / retried refreshes.
 ///
 /// # Errors
 ///
-/// Returns `Unauthorized` if the refresh cookie is missing, the JWT fails
-/// validation, the token type is not `refresh`, the user no longer exists,
-/// or the token's `refresh_token_version` doesn't match the DB (revoked via
-/// logout or password change). `Internal` for token-issue or DB failures.
+/// Returns `Unauthorized` (`token_invalid` / `token_expired` /
+/// `token_reuse_detected`) if the cookie is missing, the JWT fails validation,
+/// the token type is wrong, the user is gone, the version doesn't match, the
+/// row is missing/expired, or reuse is detected. `Internal` / `Timeout` for
+/// token-issue or DB failures.
 #[tracing::instrument(skip_all, fields(user_id = tracing::field::Empty))]
 pub async fn refresh(
     State(state): State<AppState>,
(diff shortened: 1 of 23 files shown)


backend/src/routes/auth.rs
```diff
@@ -322,27 +414,114 @@ pub async fn refresh(
...
+            // one of several concurrent refreshes flips it and rotates.
+            let claim = db_query(
+                refresh_tokens::Entity::update_many()
+                    .col_expr(refresh_tokens::Column::UsedAt, Expr::value(now))
+                    .filter(refresh_tokens::Column::Id.eq(&claims.jti))
+                    .filter(refresh_tokens::Column::UsedAt.is_null())
+                    .exec(&txn),
+            )
+            .await?;
+            if claim.rows_affected == 0 {
```
brendanbyrne: 🟡 **Important (verify / document — not broken)** — this lost-race branch is likely **unreachable on SQLite**; the losing refresh errors instead of reaching `rows_affected == 0`.

The handler does `db.begin()` (deferred `BEGIN`) → `find_by_id` (this establishes the read snapshot) → `update_many … WHERE used_at IS NULL`. Under WAL with two genuinely concurrent refreshes on the same live token: A commits its UPDATE; B's UPDATE then hits a snapshot conflict and SQLite returns `SQLITE_BUSY_SNAPSHOT` **immediately** — `busy_timeout` does *not* apply to snapshot conflicts — which `From<DbErr>` maps to a **500**, not the intended graceful `rows_affected == 0 → reissue_from_family_tip`.

The security-critical property still holds (the conditional UPDATE guarantees exactly one winner — that's the part that matters). What doesn't hold on the current engine is the *graceful loser* path; the real backstop is the grace window on the client's retry (retry sees `used_at` set within grace → reissue). This 0-rows branch is correct for the future Postgres path (READ COMMITTED blocks then returns 0 rows).

Suggest either (a) a comment noting SQLite relies on retry+grace and the 0-rows branch is the Postgres behavior, or (b) starting the txn as a write (`BEGIN IMMEDIATE`) so the loser serializes and actually sees 0 rows. No test exercises true concurrency, so this is currently unverified either way. (`seaorm.md` § 3, `tokio.md` § 6.)

backend/src/routes/auth.rs
```diff
@@ -106,6 +109,79 @@ fn extract_refresh_cookie(headers: &HeaderMap) -> Option<String> {
...
+        config,
+    )?)
+}
+
+/// Re-mint a JWT for the family's current live tip, without rotating. Used when
+/// a refresh loses a concurrent race or arrives within the grace window: the
+/// successor already exists, so we hand back a token for it rather than
+/// revoking the family on a false-positive reuse. No live tip means the family
+/// was revoked — surfaced as `token_invalid`.
+async fn reissue_from_family_tip<C: ConnectionTrait>(
```
brendanbyrne, line 163: 🔵 **Suggestion** — dual-expiry drift on grace reissue. `reissue_from_family_tip` mints a JWT with a fresh `exp` (now + TTL) while the row's `expires_at` stays at the tip's original value, so a reissued JWT can briefly outlive its row; after pruning, such a token yields `token_invalid` rather than `token_expired`.

Harmless — both are clean 401s and the row is authoritative via the `row.expires_at < now` check — just slightly surprising. A one-line comment (or clamping the reissued `exp` to the row's `expires_at`) would remove the head-scratch for a future reader.

backend/src/services/auth.rs
```diff
@@ -169,13 +192,17 @@ pub fn create_access_token(
 pub fn create_refresh_token(
     user_id: &UserId,
     refresh_token_version: i32,
+    family_id: &str,
```
brendanbyrne, line 195: 🔵 **Suggestion** — `family_id` / `jti` are passed as raw `&str` between modules here and in the route helpers. `rust.md` § 2 wants newtyped domain identifiers (`FamilyId`, `Jti`) rather than raw strings across module boundaries.

Counter-precedent worth weighing: `RefreshClaims.sub` is already a raw `String`, so the claims layer is already stringly-typed — adding two more is locally consistent. Low priority; flagging the tension rather than asserting a fix.

backend/src/services/auth.rs
```diff
@@ -241,6 +268,28 @@ pub fn clear_refresh_cookie(config: &Config) -> String {
...
+/// Background DB hygiene (ADR-0040 § Row lifecycle), run periodically from
+/// `main`. Once a token's JWT `exp` has passed it is rejected at decode before
+/// its row is ever consulted, so an expired row carries no remaining
+/// reuse-detection value — it's safe to drop.
+///
+/// # Errors
+///
+/// Returns [`Error::Timeout`] if the delete exceeds the query budget, or a
+/// mapped `DbErr` for any other database failure.
+pub async fn prune_refresh_tokens(db: &DatabaseConnection) -> Result<u64, Error> {
```
brendanbyrne: 🔵 **Suggestion** — `prune_refresh_tokens(db: &DatabaseConnection)`: `seaorm.md` § 4 says service fns take `&impl ConnectionTrait`, reserving `&DatabaseConnection` for top-level orchestration. It's only ever called from the background task with the pool handle, so this is borderline — but the generic bound matches the rest of the service layer and costs nothing.

backend/tests/auth_integration.rs
```diff
@@ -189,6 +206,130 @@ async fn test_refresh_with_valid_cookie_returns_new_access_token_and_rotated_coo
...
+    // Immediately replay cookie1 (well within the 10 s window).
+    let retry = server
+        .post("/api/v1/auth/refresh")
+        .add_cookie(cookie::Cookie::new("refresh_token", &cookie1))
+        .await;
+    retry.assert_status(StatusCode::OK);
+    let retry_body: Value = retry.json();
+    assert!(
+        retry_body["access_token"].is_string(),
+        "a within-grace replay should reissue a working token, not 401"
```
brendanbyrne: 🔵 **Suggestion** (test quality) — `["access_token"].is_string()` alongside the `200` assertion is redundant-but-weak; the status already carries the proof (any string, incl. `""`, passes). The follow-on refresh at the end of this test is the real liveness proof.

Two optional tightenings: assert non-empty (`as_str().is_some_and(|s| !s.is_empty())`), and — to actually pin the grace contract of "reissue the *existing* successor, not mint a new one" — decode the reissued cookie and assert its `jti` equals the live successor's via `validate_refresh_token`. As written the test proves "*a* working tip," not "cookie2's tip."

backend/tests/auth_integration.rs
```diff
@@ -189,6 +206,130 @@ async fn test_refresh_with_valid_cookie_returns_new_access_token_and_rotated_coo
...
+
+    // Family B is untouched — the other device still refreshes. (Reuse revokes
+    // by family, and does not bump the global `refresh_token_version`.)
+    let rb = server
+        .post("/api/v1/auth/refresh")
+        .add_cookie(cookie::Cookie::new("refresh_token", &b1))
+        .await;
+    rb.assert_status(StatusCode::OK);
+    let rb_body: Value = rb.json();
+    assert!(rb_body["access_token"].is_string());
```
brendanbyrne: 🔵 **Suggestion** (test quality) — same weak-matcher note: the `rb.assert_status(OK)` from `/refresh` already proves family B rotated successfully, so this `is_string()` is harmless but redundant. Optionally chain one more refresh on B's rotated cookie to prove the successor is itself live — that would match the strength of the within-grace test.

brendanbyrne (commented):
## Code review — refresh-token rotation with reuse detection

**Verdict: Approve** (suggestions only; one concurrency point worth a look before/after merge). Approve/request-changes left to Brendan.

**Areas covered:** `rust.md`, `seaorm.md`, `tokio.md`, `api-contract.md`, `data-model.md` / ADRs.

**Automated checks — all pass:** `cargo fmt --check` ✓ (only nightly-feature warnings), `cargo clippy --all-targets --all-features -- -D warnings` ✓, `cargo test` ✓ (full suite green). Test-quality pass over the new tests found no Critical/Important issues.

### Design-doc accuracy — verified accurate
New `### Refresh Tokens` section in `data-model.md` (columns / indexes / FK-cascade / non-STRICT all match the migration), `api-contract.md` § 4 sub-section + `token_reuse_detected` in the § 7 error table, ADR-0040 added, ADR-0031 marked superseded-in-part **both directions**. No gaps.

### What's good
- The atomic `UPDATE … WHERE used_at IS NULL` single-winner primitive — correct and elegant.
- Minting the access token **before** `txn.commit()` so a JWT failure rolls back the rotation — nice defensive ordering.
- Reuse path commits the family-delete **before** returning the error — revocation persists.
- `db_query` / `db_txn` timeout wrapping applied to **every** new DB call (`tokio.md` § 12).
- Prune task is textbook `tokio.md` § 8 / § 13: `biased` select on cancel, `MissedTickBehavior::Skip`, supervised, logged, shutdown-safe (single idempotent DELETE).
- Entity + behavior + migration + `schema_drift.rs` wired atomically in one PR (`seaorm.md` § 6); behavior file mirrors `notifications_behavior` exactly. Migration edited in place per § 5 with the append-only deviation explicitly flagged.
- No token/secret leakage in logs — logs `family_id` + `user_id` only (`rust.md` § 10).

Inline comments below carry the 5 findings (1 Important, 4 Suggestion).

backend/src/routes/auth.rs
brendanbyrne: Documented (your option a) in fafb611. Agreed the 0-rows branch is really the Postgres/READ-COMMITTED path — on SQLite the loser hits `SQLITE_BUSY_SNAPSHOT` (busy_timeout doesn't apply) → 500. Added an engine-note comment at the branch spelling out that SQLite's real backstop is the client retry landing in the grace window, plus the frontend single-flight from the follow-up PR. Skipped `BEGIN IMMEDIATE` deliberately: it serializes *every* refresh on the write lock (against seaorm.md §3's minimum-scope rule) to remove a 500 that only fires on a truly-concurrent same-token double-write, which single-flight already prevents. Easy to revisit if you'd rather serialize.

backend/src/routes/auth.rs
brendanbyrne: Added a comment on `reissue_from_family_tip` in fafb611: the reissued `exp` (now+TTL) can briefly outlive the tip row's `expires_at`, and it's harmless because the row is authoritative (the `expires_at < now` check) — a pruned-row token surfaces as `token_invalid` not `token_expired`, both clean 401s. Left clamping out to avoid threading an explicit `exp` through `create_refresh_token` for a cosmetic edge.

backend/src/services/auth.rs
brendanbyrne: Leaving as raw `&str` for this PR, with your counter-precedent as the reason: the claims layer is uniformly stringly-typed — `RefreshClaims.sub` is a raw `String`, not even `UserId` — so newtyping only jti/family_id would be locally inconsistent. The adjacent-`&str` swap risk is already guarded by `test_create_and_validate_refresh_token`, which round-trips distinct jti/family_id values. If we newtype the claims layer wholesale (sub → UserId-in-claims) later, `FamilyId`/`Jti` come along then.

backend/src/services/auth.rs
brendanbyrne: Done in fafb611 — `prune_refresh_tokens(conn: &impl ConnectionTrait)`. Matches seaorm.md §4 and the `mint_refresh_in_family` / `reissue_from_family_tip` helpers; the main.rs call site (pool handle) is unchanged.

backend/tests/auth_integration.rs
brendanbyrne: Done in fafb611. The test now decodes cookie2 (the rotation's successor) and the reissued cookie via `validate_refresh_token` and asserts their `jti` matches — pinning 'reissue the *existing* successor', not just 'a working tip'. Kept the follow-on refresh as the liveness proof.

backend/tests/auth_integration.rs
brendanbyrne: Done in fafb611 — dropped the `is_string()` and chained a refresh on family B's rotated successor (`b2`), proving the family is fully intact rather than that b1 worked once.

brendanbyrne (commented):
## Re-review — fafb611 ("address PR #230 review")

**Verdict: Approve.** All four actionable findings from the prior review are resolved; nothing new surfaced. Approve/request-changes left to Brendan.

**Automated checks — all pass on `fafb611`:** `cargo fmt --check` ✓, `cargo clippy --all-targets --all-features -- -D warnings` ✓, `cargo test --test auth_integration` ✓ (23 passed), `cargo test --lib services::auth` ✓ (13 passed).

### Findings resolution
- 🟡 **Lost-race branch / SQLite `SQLITE_BUSY_SNAPSHOT`** (was Important) — **resolved by documentation** (option a). The new comment at the `rows_affected == 0` branch accurately frames it as the Postgres (READ COMMITTED) behavior, notes the single-winner guarantee is unaffected, and names the SQLite backstop (client retry → grace window + frontend single-flight). Keeping deferred `BEGIN` rather than `BEGIN IMMEDIATE` is a fine call now that the behavior is written down.
- 🔵 **Dual-expiry drift** — resolved: doc comment on `reissue_from_family_tip` explains the fresh-`exp`-vs-unchanged-`expires_at` drift, that the row is authoritative, and that a pruned row surfaces as a clean `token_invalid`.
- 🔵 **`prune_refresh_tokens` connection type** — resolved: now `conn: &impl ConnectionTrait` (`seaorm.md` § 4).
- 🔵 **Weak test matchers** — resolved well:
  - `test_refresh_within_grace_window…` now decodes both cookies and asserts `reissued_jti == successor_jti`, pinning the actual grace contract ("reissue the *existing* successor, not a fresh mint") — stronger than the non-empty tweak I suggested.
  - `test_two_families_coexist…` now chains a second refresh on family B's rotated cookie, proving B is fully intact rather than that `b1` worked once.

### Not changed (acceptable)
- 🔵 **`family_id` / `jti` as raw `&str`** — left as-is. Fine to defer: it was the lowest-priority suggestion and `RefreshClaims.sub` already establishes a stringly-typed claims layer here, so this stays locally consistent.

Design docs remain accurate (no schema/contract changes in this commit). Good, tightly-scoped revision — nothing further from me.

brendanbyrne:
## Addressed the SQLite concurrency finding (commit f308b5b)

The reviewer was right that the lost-race branch wasn't graceful on SQLite — and the empirical follow-up showed it was **worse than my original "near-impossible edge" framing**: it's a routine spurious logout under this app's polling client. I've corrected both the code and the record.

### What was actually happening
The handler took a read snapshot (`find_by_id`) **before** the conditional `used_at` UPDATE. Under WAL + deferred `BEGIN`, a losing concurrent refresh upgrades a now-stale read snapshot to a write and fails **immediately** with a `SQLITE_BUSY`-family error — `busy_timeout` does *not* retry a snapshot/upgrade conflict — which mapped to a 500. The frontend turns any non-2xx refresh into a hard logout, and the polling client (`useSession` 2.5s, `useSessions` ×2 5s, `refetchOnWindowFocus`) fires concurrent same-cookie refreshes routinely. So this was a likely-recurring mid-session logout, introduced by this PR's rotation.

### The fix
Make the conditional claim the **first statement** in the transaction, so SQLite takes the write lock up front (the `BEGIN IMMEDIATE` shape). A loser then *waits* on `busy_timeout` for the winner to commit, re-evaluates `WHERE used_at IS NULL` against committed state, and cleanly matches 0 rows → grace/reissue → 200. No error-classification code, and it also fixes a latent stale-snapshot read in the lost-race branch (the post-write-lock read now sees the winner's committed successor).

### Measured (real file-backed WAL DB, production pragmas, 4 racers × 20 rounds)
| pattern | winners | clean 0-row losers | **errored losers (→500)** |
|---|---|---|---|
| read-first (before) | 20/20 | 1/60 | **59/60** |
| write-first (after) | 20/20 | 60/60 | **0/60** |

### Corrections to my earlier comments
- I'd written that the loser "always returns `SQLITE_BUSY_SNAPSHOT`" — under genuine concurrency it's usually plain `SQLITE_BUSY` (code 5); the barriered worst case is 517. Write-first eliminates both, so the distinction no longer matters in the code, and I removed the inaccurate comment.
- `BEGIN IMMEDIATE` via a SeaORM/sqlx knob was a dead end (`begin_with_config` is a no-op for SQLite; no sqlx connect-option); write-first is the idiomatic equivalent on this stack.

The frontend hardening (single-flight + treating 5xx as transient instead of logging out) lands in the follow-up frontend PR per ADR-0040 — it's the complementary client-side defense, but this PR no longer depends on it to avoid the spurious logout.

PR #231, opened 2026-06-13
226: single-flight refresh + transient-aware auth handling (ADR-0040 frontend)

brendanbyrne:
Frontend follow-up to **#230** (backend rotation + reuse detection), completing the ADR-0040 client-side mitigation. Branches off `main` (frontend-only — no overlap with #230's backend files, so no stacking).

## Problem

`apiFetch`'s refresh handling collapsed every non-2xx into "session expired → hard logout," and fired a separate refresh per concurrent 401:

1. **5xx / network during refresh evicted the user mid-session** — even though a server error says nothing about the cookie's validity. With #230's rotation, contended refreshes can briefly 500, and this app polls aggressively (`useSession` 2.5s, `useSessions` ×2 5s, `refetchOnWindowFocus`), so a transient hiccup → spurious logout was a real, recurring failure mode. Logging a user out on a server problem is an availability harm with **no security benefit** — the client logout revokes nothing server-side.
2. **No single-flight** — parallel refreshes of the same cookie race the backend's rotation and can trip a false-positive reuse detection.

## Change

`tryRefresh` is now **single-flight** (concurrent 401s share one in-flight refresh) and returns a classified `RefreshOutcome`:

| Outcome | Trigger | Action |
|---|---|---|
| `refreshed` | 2xx with token | retry the original request |
| `transient` | **5xx / network error** | **keep the session**, surface the original 401, let a later refresh retry |
| `expired` | 401 (or a malformed 2xx body) | ordinary sign-out |
| `reuse` | 401 `token_reuse_detected` | sign out + **"signed out for security"** notice on the login screen |

Only `401` (the cookie itself being rejected) ends the session; a `5xx` never does. The reuse case is distinguished from ordinary expiry so the login screen can explain the security sign-out.

### Files
- `api/result.ts` — `token_reuse_detected` added to the wire error-code registry.
- `api/client.ts` — single-flight + `RefreshOutcome` classification; `AuthFailureReason` threaded through `onAuthFailure`.
- `hooks/useAuth.tsx` — `authNotice` on the context; reuse → security notice, cleared on logout / re-login.
- `pages/Login.tsx` — renders `authNotice` as a `role="status"` banner.

## Tests (246 pass; +10 new)
- single-flight: concurrent 401s trigger exactly one refresh
- 5xx refresh → no logout, token retained, original 401 surfaced
- network error → no logout, token retained
- `token_reuse_detected` → sign out with reason `reuse`, token cleared
- ordinary 401 → sign out with reason `expired`, **no** security notice
- notice cleared on re-login; Login banner renders / hides on `authNotice`
- recognition of `token_reuse_detected` in `parseApiError`

## Manual test
1. `bun run dev` (root) + backend running.
2. Log in. In devtools, throttle/stop the backend briefly so `/api/v1/auth/refresh` returns 5xx, then let an access token lapse (or force a 401). **Expect:** you stay logged in (no bounce to `/login`); the next successful refresh recovers silently.
3. Restore the backend. Replay a stale refresh cookie past the grace window (or simulate a `401 token_reuse_detected` from the refresh endpoint). **Expect:** you're returned to `/login` with the "You were signed out for your security. Please log in again." banner.
4. Open two tabs logged into the same account and let both access tokens lapse near-simultaneously. **Expect:** no spurious logout; rotation proceeds (with #230's backend, the loser reissues within grace).

## Docs
The ADR-0040 "Implementing PRs" line and the api-contract.md "deferred to the follow-up PR" notes will be flipped to reference this PR on merge (docs-direct-to-main per the post-merge convention), keeping this PR scoped to code + tests and decoupled from the repo-wide markdown link-check.

### DIFF
--- a/frontend/src/api/client.test.ts
+++ b/frontend/src/api/client.test.ts
@@ -1,7 +1,12 @@
-import { http, HttpResponse } from 'msw';
+import { delay, http, HttpResponse } from 'msw';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 import { server } from '../mocks/server';
-import { apiFetch, setAccessToken, setOnAuthFailure } from './client';
+import {
+  apiFetch,
+  getAccessToken,
+  setAccessToken,
+  setOnAuthFailure,
+} from './client';
 
 // Verifies the access-token + 401-refresh behavior of `apiFetch`. The
 // user-visible promise: an expired access token is refreshed and the request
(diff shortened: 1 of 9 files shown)


frontend/src/hooks/useAuth.tsx
```diff
@@ -37,25 +42,43 @@ interface AuthContextValue {
...
+      clearAuth();
+      if (reason === 'reuse') setAuthNotice(SECURITY_LOGOUT_NOTICE);
+    },
+    [clearAuth],
+  );
+
   // On mount: attempt silent refresh to restore session from the HttpOnly cookie.
   useEffect(() => {
-    setOnAuthFailure(clearAuth);
+    setOnAuthFailure(handleAuthFailure);
```
brendanbyrne, line 81: 🔵 **Suggestion** — The mount `silentRefresh` below does its own raw `fetch('/api/v1/auth/refresh')` rather than going through `tryRefresh`, so it sits outside the single-flight guard this PR adds. In production this is safe: at mount `accessToken` is `null`, so `apiFetch`'s `res.status === 401 && accessToken` guard never fires a concurrent `tryRefresh` — exactly one refresh runs. But under React StrictMode (dev) the effect double-invokes, firing **two** parallel mount refreshes — the precise race this PR exists to prevent. The backend's ~10s grace window (ADR-0040) absorbs it (returns the successor, no false-positive revoke), so not a correctness bug. Worth a one-line comment noting the mount path is deliberately outside single-flight and relies on the grace window, so a future edit doesn't assume the guard covers it.

frontend/src/api/client.ts
```diff
@@ -24,36 +36,81 @@ export function getAccessToken(): string | null {
...
+      // an expired session. Surface it (§ 8); end the session so re-login (a
+      // different endpoint) can recover rather than loop on the broken body.
+      logIfResponseShapeMismatch(e, '/api/v1/auth/refresh');
+      return 'expired';
+    }
   }
+
+  // A non-401 failure (5xx, proxy error) is a server problem, not a verdict on
+  // the cookie — keep the session and let a later refresh retry.
+  if (res.status !== 401) return 'transient';
```
brendanbyrne, line 110: 🔵 **Suggestion** — This `res.status !== 401` check routes *all* non-401, non-2xx responses to `transient`, including 4xx (429/400/403), not just the 5xx/proxy errors the comment names. For the documented contract (refresh returns 2xx / 401 / 5xx) this never arises, and "keep the session" is the right default for an unexpected status per this PR's philosophy. Suggest broadening the comment to "any non-401 (5xx, 4xx, proxy error)" so the catch-all is explicit rather than reading as 5xx-only.

frontend/src/api/client.test.ts
```diff
@@ -133,4 +138,118 @@ describe('apiFetch', () => {
...
+      http.get('/api/v1/me', () => new HttpResponse(null, { status: 401 })),
+      http.post('/api/v1/auth/refresh', () =>
+        HttpResponse.json(
+          { error: 'Refresh token has been revoked', code: 'token_invalid' },
+          { status: 401 },
+        ),
+      ),
+    );
+
+    await apiFetch('/api/v1/me');
```
brendanbyrne: 🔵 **Suggestion** — Unlike its four sibling tests, this one discards the response. Capture `const res = await apiFetch(...)` and add `expect(res.status).toBe(401)` for parity with the `reuse` test directly above. The test is otherwise sound — it pins `onFailure` called with `'expired'` and `getAccessToken()` nulled, which are the load-bearing claims.

brendanbyrne (commented):
## Code Review — single-flight refresh + transient-aware auth handling

**Verdict: Approve** (minor suggestions only — nothing blocks merge; final approval is Brendan's).

Makes the 401→refresh→retry interceptor single-flight and replaces "any non-2xx → hard logout" with a classified `RefreshOutcome` so a 5xx/network hiccup keeps the session and only a genuine 401 ends it — the client half of ADR-0040.

**Standards consulted:** `api-contract.md` §§ 2/7, ADR-0040, `react.md` §§ 6/7, `testing.md`, `typescript.md` § 6, plus the backend wire source.

**Automated checks:** `bun run lint` 0 errors (106 warnings, all pre-existing in untouched `Session.tsx`/`time.ts`); `bun run typecheck` clean; `bun run test` 246 passed (+10 new).

**Contract validation (not just the mock):** `backend/src/error.rs` has `#[serde(rename_all = "snake_case")]` on `ErrorCode`, so `TokenReuseDetected` → `token_reuse_detected`, and `Error::token_reuse_detected()` maps to 401. The test fixtures (`{ error, code: 'token_reuse_detected' }` @ 401) match the real backend exactly.

**Test quality:** a `test-quality-review` pass cleared all added tests — each pins the *specific* reason/code (not just status), drives real outcomes through MSW, and awaits async work; the reuse-vs-expired pair asserts the differentiator (`authNotice` set vs `null`). Per `testing.md` §§ 1–2.

**Design-doc accuracy:** ADR-0040's "Implementing PRs … frontend TBD" line and api-contract.md's "deferred to the follow-up PR" notes (§§ 4/7) still mark this work pending. The PR body flips them via a direct-to-main docs commit on merge (docs-only skips review; also keeps this code PR off the repo-wide link-check). A legitimate, documented deferral — `design.md` itself needs no change (it doesn't describe the interceptor). Just confirm that follow-up commit lands so the contract docs don't stay stale.

**What's good:**
- `inFlightRefresh ??= refreshOnce().finally(...)` is a clean single-flight idiom, and `refreshOnce` is *total* — every path returns (network reject caught, `parseApiError` never throws, `parseBody` throw caught in the `res.ok` branch), so the shared promise never rejects onto awaiters.
- Correct `useCallback` retention for `clearAuth`/`handleAuthFailure` — the exact `react.md` § 7(d) correctness carve-out, with a comment citing the rule instead of silencing exhaustive-deps.
- Sound, well-documented security reasoning: a client logout revokes nothing server-side, so evicting on a 5xx is pure availability harm.
- `frontend.yml` codecov workaround mirrors backend.yml exactly (same v6.0.1 SHA pin, `skip_validation: true`, REVERT condition).

No critical or important findings. Three suggestions inline.

brendanbyrne (commented):
## Re-review — `f998ceb` (address review)

All three 🔵 suggestions from the prior review are addressed correctly:

- **Mount `silentRefresh` outside single-flight** — `useAuth.tsx` now documents the deliberate raw-fetch choice, why it's safe (null `accessToken` at mount → no competing `tryRefresh`), the StrictMode double-fire, and the ADR-0040 grace-window backstop, with a "reconsider if a token can ever be set before mount" guard. 
- **Non-401 → transient comment** — `client.ts` broadened to "any non-401 (5xx, 4xx, proxy error)… only a 401 ends the session." 
- **Test parity** — the `expired` test captures `res` and asserts `status).toBe(401)`, matching its `reuse` sibling.

**Checks on this HEAD:** `tsc -b` clean; the 4 affected test files pass (39/39); lint 0 errors (remaining warnings on `useAuth.tsx` are pre-existing repo-wide style warnings, none introduced here).

Nothing outstanding. Reminder for merge: flip the ADR-0040 "Implementing PRs … frontend TBD" line and the api-contract.md §§ 4/7 "deferred to follow-up PR" notes to point at #231 (direct-to-main docs commit per the PR plan).

PR #233, opened 2026-06-15
227: flip no-conditional-expect + no-node-access lint rules to error

### LINKED ISSUE #227: Flip eslint `vitest/no-conditional-expect` and `testing-library/no-node-access` from warn to error
## Context

PR #225 (Issue #217) adopted the `@vitest/eslint-plugin` + `eslint-plugin-testing-library`
presets in `frontend/eslint.config.js`. Two rules start at **`warn`** because
they fire on legitimate existing patterns rather than real defects:

- **`vitest/no-conditional-expect`** (8 offenders) — the
  `try { …; expect.unreachable() } catch { expect(err) }` shape used to assert on
  a thrown error's nested fields (`result.test.ts`, `runs.test.ts`).
- **`testing-library/no-node-access`** (15 offenders) — the raw-node drag
  simulation for the custom `<div>`-based DQ slider (`RunEntrySheet.test.tsx`),
  which has no Testing Library query.

Per `eslint.config.js`'s ratchet convention, a rule flips to `error` once its
offender count reaches zero. PR #225 already did this for 8 type-checked rules
that were clean (commit `57d7659`); these two still have offenders.

## Work

1. **`no-conditional-expect`:** convert the `try/expect.unreachable()/catch`
   tests to `await expect(promise).rejects.toMatchObject({ … })` (or `.rejects`
   + the relevant matcher) so the assertion isn't inside a `catch`.
2. **`no-node-access`:** rework the DQ-slider drag to drive a queryable element
   via Testing Library (`fireEvent`/`pointer` on a role-bearing node), or — if
   the raw-node access is genuinely unavoidable — add a justified inline
   `// eslint-disable-next-line testing-library/no-node-access`.
3. When each reaches **zero** offenders, move it from the "still at `warn`" block
   to `error` in `frontend/eslint.config.js` and drop the warn-explaining comment.

## Acceptance

- `bunx eslint .` reports 0 warnings for both rules.
- Both rules set to `error`; `bun run lint` stays green.

_Follow-up from PR #225 review._

brendanbyrne:
Closes #227. Follow-up from PR #225 review (the frontend-standards lint-preset adoption).

## What & why

PR #225 adopted the `@vitest/eslint-plugin` and `eslint-plugin-testing-library` presets but started two rules at **`warn`** because each fired on a legitimate existing pattern rather than a real defect. This PR reworks those patterns to zero offenders, then ratchets both rules to **`error`** per `eslint.config.js`'s ratchet convention.

Both rules are already `error` in their upstream presets — the config explicitly *downgraded* them to `warn`. So the flip is just restoring `error`, now that there are zero violations.

### `vitest/no-conditional-expect` (was 8 offenders)
- **`src/api/result.test.ts`** — the `try { …; expect.unreachable() } catch { expect(err) }` shape put assertions inside a `catch`. Reworked to a **capture-once** pattern: `const err = await parseBody(...).catch((e) => e)` then assert at the top level. (Capture-once, not `await expect(...).rejects` repeated, because a `Response` body can only be read once — re-awaiting would throw "body already read" instead of the schema error.)
- **`src/api/runs.test.ts`** — the discriminated-`Result` narrowing put assertions inside `if (result.ok) { … }`. Reworked to a **throwing guard** (`if (!result.ok) throw …`) that narrows the union so the assertions sit at the top level.

### `testing-library/no-node-access` (was 15 offenders)
- **`src/components/RunEntrySheet.test.tsx`** — the DQ-slider drag tests walked `.parentElement` from the `»` thumb text to reach the thumb and track DOM nodes. The **track** is the `role="button"` element that carries *every* drag handler (`onMouseDown/Move/Up`, `onTouch*`); the thumb is `aria-hidden` and presentational (a press on it just bubbles to the track). So the tests now query the track directly via its accessible name (`findByRole('button', { name: /disqualify this run/i })`) and fire all events on it — behaviorally identical (the geometry mock is on `HTMLElement.prototype`), zero node walking, no `eslint-disable` needed.

No production code changes — only test files and `eslint.config.js`.

## Testing

No user-visible behavior changes; verification is lint + the existing suite:

```bash
cd frontend
bun install --frozen-lockfile

# 1. The two target rules report ZERO offenders:
bunx eslint . | grep -E "no-conditional-expect|no-node-access"   # → no output

# 2. Lint passes with both rules now at `error` (remaining warnings are other
#    rules still intentionally at warn, owned by later compliance PRs):
bun run lint        # → "0 errors", exit 0

# 3. Full suite still green (the reworked DQ-slider + Result/parseBody tests):
bun run test        # → 30 files, 246 tests passed

# 4. Types still check (the throwing-guard narrowing relies on TS):
bun run typecheck   # → clean
```

To confirm the ratchet actually bites: temporarily re-introduce a `expect` inside a `catch` (or a `.parentElement` access) in any test file and `bun run lint` should now **fail** with an error, not a warning.

### DIFF
--- a/frontend/src/api/runs.test.ts
+++ b/frontend/src/api/runs.test.ts
@@ -143,10 +143,11 @@ describe('getRunDefaults', () => {
     );
     const result = await getRunDefaults();
     expect(result.ok).toBe(true);
-    if (result.ok) {
-      expect(result.value.source).toBe('previous_run');
-      expect(result.value.character_id).toBe(1);
-    }
+    // Narrow the discriminated Result with a throwing guard so the assertions
+    // sit at the top level, not inside an `if` (vitest/no-conditional-expect).
+    if (!result.ok) throw new Error('expected an ok Result');
+    expect(result.value.source).toBe('previous_run');
+    expect(result.value.character_id).toBe(1);
   });
 
   it('returns an error Result carrying the typed code when the request fails', async () => {
@@ -160,8 +161,7 @@ describe('getRunDefaults', () => {
     );
     const result = await getRunDefaults();
     expect(result.ok).toBe(false);
-    if (!result.ok) {
-      expect(result.error.code).toBe('internal');
-    }
+    if (result.ok) throw new Error('expected an error Result');
+    expect(result.error.code).toBe('internal');
   });
 });
--- a/frontend/src/api/runs.test.ts
+++ b/frontend/src/api/runs.test.ts
@@ -163,5 +163,7 @@ describe('getRunDefaults', () => {
     expect(result.ok).toBe(false);
     if (result.ok) throw new Error('expected an error Result');
     expect(result.error.code).toBe('internal');
+    // Also pin the backend `error` field threaded through parseApiError.
+    expect(result.error.message).toBe('Defaults unavailable');
   });
 });
--- a/docs/coding-standards/testing.md
+++ b/docs/coding-standards/testing.md
@@ -125,10 +125,11 @@ These rules are partly enforced so they can't silently reappear:
   flat React config, scoped to test files in [`eslint.config.js`](../../frontend/eslint.config.js).
   Together they lock in `valid-expect`, `valid-expect-in-promise`,
   `expect-expect`, `no-focused-tests`, `no-disabled-tests`, and the async-query
-  hygiene rules. Two rules that fire on existing valid patterns
-  (`vitest/no-conditional-expect`, `testing-library/no-node-access`) start at
-  `warn` per that file's preamble. Rule 2 (outcome over choreography) is
-  review-enforced.
+  hygiene rules. Two rules that fired on existing valid patterns
+  (`vitest/no-conditional-expect`, `testing-library/no-node-access`) started at
+  `warn` and were ratcheted to `error` once their offenders were reworked to
+  unconditional assertions / accessible-name queries (#227). Rule 2 (outcome
+  over choreography) is review-enforced.
 
 ## Document history
 
--- a/frontend/eslint.config.js
+++ b/frontend/eslint.config.js
@@ -174,13 +174,11 @@ export default defineConfig([
     files: ['**/*.test.{ts,tsx}', 'src/__tests__/**/*.{ts,tsx}'],
     rules: {
       ...vitest.configs.recommended.rules,
-      // Started at `warn`, not `error`, per this file's preamble: it fires on
-      // two legitimate shapes for asserting on a failure — result.test.ts's
-      // `try { …; expect.unreachable() } catch { expect(err) }`, and
-      // runs.test.ts's discriminated-`Result` narrowing
-      // (`expect(r.ok).toBe(false); if (!r.ok) expect(r.error.code)…`). Flip to
-      // `error` once both are reworked to unconditional assertions (#227).
-      'vitest/no-conditional-expect': 'warn',
+      // Enforced once its offender count hit zero (#227): result.test.ts's
+      // `try/expect.unreachable()/catch` and runs.test.ts's in-`if` Result
+      // narrowing were reworked to unconditional assertions. (The preset already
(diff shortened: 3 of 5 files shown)


frontend/src/api/runs.test.ts
```diff
@@ -160,8 +161,7 @@ describe('getRunDefaults', () => {
     );
     const result = await getRunDefaults();
     expect(result.ok).toBe(false);
-    if (!result.ok) {
-      expect(result.error.code).toBe('internal');
-    }
+    if (result.ok) throw new Error('expected an error Result');
+    expect(result.error.code).toBe('internal');
```
brendanbyrne, line 165: 🔵 **Suggestion (optional)** — this correctly pins `result.error.code === 'internal'` (the typed code the test name foregrounds, `testing.md` § 1). The MSW handler also returns `error: 'Defaults unavailable'`, which `getRunDefaults` threads through `parseApiError`. Adding `expect(result.error.message).toBe('Defaults unavailable');` would also pin the message-passthrough. Belt-and-suspenders, not required.

brendanbyrne (commented):
## Code review — flip `no-conditional-expect` + `no-node-access` to error

**Verdict: Approve.** Clean, well-scoped ratchet flip. No production code changes; the two reworked assertion patterns are sound and (for `runs.test.ts`) slightly *stronger* than what they replaced. One doc-accuracy gap found and fixed on this branch; one optional suggestion inline.

### What was verified
- `bun run lint` → **0 errors** (83 warnings, all unrelated rules still intentionally at `warn`, owned by later compliance PRs).
- `bunx vitest run` on the 3 changed files → **35 tests pass**.
- Offender scan for both target rules → **zero**.
- Behavioral-equivalence claim verified against `RunEntrySheet.tsx:672-719`: the `role="button"` track carries *every* drag handler (`onMouseDown/Move/Up/Leave`, `onTouch*`, `onKeyDown`); the thumb is `aria-hidden` with zero handlers. `findByRole('button', { name: /disqualify this run/i })` resolves to that exact track, so firing events on it is behaviorally identical to (and more robust than) the old `.parentElement` walk.
- `test-quality-review` run on all three files: all reworks sound. The `result.test.ts` capture-once `.catch((e) => e)` does **not** open a non-throwing-resolution hole — on success `parseBody` returns the validated plain object, which is type-disjoint from `ApiErrorException`, so `toBeInstanceOf(ApiErrorException)` is a sufficient guard. `code`/`cause` are still pinned (`testing.md` § 1). The `runs.test.ts` throwing guard is strictly stronger than the old in-`if` form, which would have passed green on a wrong discriminant.

### 🟡 Important (resolved on this branch) — `testing.md` § Tooling was stale
`docs/coding-standards/testing.md` § Tooling still described these two rules as starting at `warn`, which this PR makes false. Per the same-PR-doc-update convention I folded the fix into this branch (commit dc5558b): the sentence now reads that both *were ratcheted from `warn` to `error`* (#227). No `## Document history` entry added — a line there would just restate the diff, which the decisions-not-changelog rule says to skip.

### Design-doc accuracy
`design.md`: **no changes required** — no architecture / data-model / API-surface change. The only doc drift was the `testing.md` § Tooling sentence above.

### What's good
- The capture-once rationale (a `Response` body reads once, so `.rejects` can't be re-awaited) is correct and documented inline — a subtle trap avoided.
- Replacing `.parentElement` walking with an accessible-name query exceeds what the rule strictly requires: it reaches the handler element by *contract*, not DOM structure.
- `eslint.config.js` comments correctly note the preset already sets `error` and that the explicit listing documents the ratchet — good provenance for the next reader.

brendanbyrne:
Thanks for the review. Applied the optional suggestion in 626f7fa — `runs.test.ts` now also pins `result.error.message === 'Defaults unavailable'` (verified against `result.ts:123-126`, where the wire `error` field maps to `message`). The `testing.md` § Tooling sync (dc5558b) is pulled into my local branch. No production code touched; full suite still green (246 tests).

PR #234, opened 2026-06-15
222: focus management for RunEntrySheet drink/setup sub-pickers

### LINKED ISSUE #222: Frontend: focus management for RunEntrySheet's drink/setup sub-pickers
## Context

Follow-up surfaced in the [PR #221](https://github.com/brendanbyrne/beerio-kart/pull/221) review (PR-G2, the accessibility sweep). PR-G2 added modal a11y to the two modals named in its acceptance criteria — the Home create-session modal and the RunEntrySheet bottom sheet — but **deliberately left the RunEntrySheet's nested drink/setup sub-pickers out of scope**.

Those sub-pickers ([`RunEntrySheet.tsx`](../blob/main/frontend/src/components/RunEntrySheet.tsx)) render as opaque `fixed inset-0 z-50` overlays **on top of the still-mounted sheet**. `useModalA11y`'s trap is *suspended* while a sub-picker is open (via the `active` arg) so it doesn't fight the picker for focus — a reasonable interim, but it leaves two gaps:

1. **On open**, focus stays on the now-covered trigger button and nothing traps focus inside the picker, so a keyboard / screen-reader user can Tab into the occluded sheet behind it.
2. **On close**, the sheet's trap re-runs and re-seats focus to the *first* control in the sheet rather than the drink/setup button the user came from.

## Acceptance criteria

- [ ] Opening a drink/setup sub-picker moves focus into the picker.
- [ ] While a sub-picker is open, the sheet behind it cannot receive focus (keyboard or SR). Recommended: mark the sheet `inert` while a picker is open — this also closes the SR-leak in (1) without a second trap.
- [ ] Closing a sub-picker restores focus to the drink/setup button that opened it.
- [ ] Escape closes the sub-picker (returning to the sheet), consistent with the modal contract.
- [ ] Tests cover open-focus, the `inert`/no-leak guarantee, and close-restore.

## References

- Review thread: PR #221 (inline comment on `RunEntrySheet.tsx`).
- Standards: `docs/coding-standards/react.md` § 10 (modal focus contract).
- Reusable utility: `frontend/src/hooks/useModalA11y.ts` (the sheet/modal trap; sub-pickers may need a variant or the `inert` approach).

brendanbyrne:
Closes #222.

## What & why

PR-G2 added modal a11y to the RunEntrySheet bottom sheet but left its nested drink/setup sub-pickers out of scope. Those overlays mounted on top of the still-focusable sheet with no focus management, so a keyboard/SR user could Tab into the occluded sheet, and closing a picker dumped focus on the sheet's first control instead of the button they came from.

Each sub-picker is now a `SubPickerOverlay` running its own `useModalA11y`, and the sheet behind is marked `inert` while a picker is open.

## How

- **`SubPickerOverlay`** — extracts the two inline picker overlays into one component that calls `useModalA11y(onClose, …)`: seats focus into the picker on open, traps Tab, closes on Escape + Back, and restores focus to the opening control on close. Takes its accessible name from the visible heading via `aria-labelledby` (matching `Home.tsx`'s `CreateSessionModal`).
- **`inert` on the sheet** — the sheet dialog + backdrop get `inert={subPickerOpen}`, and the sheet's own trap suspends (`active=false`) while a picker is open. `inert` removes the sheet from both Tab order and the accessibility tree, sealing the keyboard + SR leak; suspending the sheet trap stops it fighting the picker for focus.
- **Explicit trigger capture (the subtle bit)** — opening a picker flips the sheet to `inert` in the *same commit* that mounts the overlay. In real browsers, making an ancestor `inert` synchronously **blurs** the focused trigger to `<body>` before the overlay's mount effect runs, so reading `document.activeElement` would capture the wrong element and restore focus to the wrong place on close (Safari compounds this — it never focuses a button on click). So the drink/setup buttons capture `e.currentTarget` at click time, and `useModalA11y` gained an optional `restoreFocusRef` to restore to it.
- **Trap re-seat guard** — `useModalA11y`'s resumed trap now re-seats focus only when focus has escaped the dialog (`!node.contains(document.activeElement)`), so resuming the sheet after a picker closes doesn't steal focus back from the just-restored trigger.

> This was caught by an adversarial pre-PR review: the original draft captured `document.activeElement` and passed every jsdom test, but would have shipped broken in Firefox/Safari/Chrome because jsdom doesn't implement `inert`'s blur. The fix is verified by a test that opens the picker *without* focusing the trigger (reproducing the blurred state) — it fails on the `document.activeElement` approach.

## Acceptance criteria

- [x] Opening a drink/setup sub-picker moves focus into the picker.
- [x] While a sub-picker is open, the sheet behind cannot receive focus (keyboard or SR) — via `inert` + suspended trap.
- [x] Closing a sub-picker restores focus to the drink/setup button that opened it.
- [x] Escape closes the sub-picker (returning to the sheet), for both pickers.
- [x] Tests cover open-focus, the `inert`/no-leak guarantee, and close-restore.

## Tests

`frontend` suite: **254 passing**. New tests:
- `RunEntrySheet.test.tsx` — open moves focus into the drink picker; sheet is `inert` while open and not after close; close restores focus to the trigger; **a `fireEvent`-open case that pins the explicit-trigger fix** (fails on a `document.activeElement` restore); Escape closes each picker without closing the sheet; setup-picker open-focus + restore.
- `useModalA11y.test.tsx` — restores focus to an explicit `restoreFocusRef` over the element focused at mount.

Patch coverage: the new product lines are fully covered (`useModalA11y.ts` 100%; the new `RunEntrySheet` lines all hit). Typecheck clean; 0 new lint errors.

## Manual test steps

Run the app from the repo root: `bun run dev`, then open the run-entry sheet for a race (Session → a race → Log a run).

**Keyboard (Chrome + Firefox):**
1. Tab to the **Change** (drink) button and press Enter → the drink picker opens and focus lands on its **Back** button.
2. Press **Tab** repeatedly → focus cycles only within the picker; it never reaches the sheet behind.
3. Press **Escape** → the picker closes and focus returns to the **Change** button (not the Total Time field). Repeat with Enter-then-Back.
4. Repeat 1–3 with the **Edit** (race setup) button.

**`inert` seal (verify in Firefox + a real browser — jsdom can't):**
5. Open the drink picker, open DevTools, and inspect the sheet `<div role="dialog" aria-label="Log your run">` → it has the `inert` attribute while the picker is open and loses it on close.
6. With the picker open, try Tab/Shift+Tab → you cannot focus anything in the sheet behind. With a screen reader (VoiceOver/NVDA), the sheet's controls are not announced while the picker is open.

**Mouse/touch (regression):**
7. Open each picker, pick a drink / complete the setup → the sheet shows the new selection; the DQ slider, time inputs, and defaults pre-fill all still work.


brendanbyrne (commented):
## Review — focus management for RunEntrySheet drink/setup sub-pickers

**Verdict: ✅ Approve** — no blockers. Strong, careful PR; the findings below are all minor/optional. (Posting as a comment; `APPROVE` is reserved for Brendan.)

**What this PR does:** Extracts the two inline full-screen pickers into a shared `SubPickerOverlay` that runs its own `useModalA11y`, marks the sheet behind `inert` + suspends its trap while a picker is open, and captures the opening trigger at click time so focus restores correctly despite `inert` blurring it. Adds an optional `restoreFocusRef` param to `useModalA11y` and a re-seat guard on the resumed trap.

**Verified:**
- `bun run lint` → **0 errors**, 82 warnings (all pre-existing patterns; this PR adds one `interface` warning of a category already present 4× in the file — inline comment on line 483).
- Tests → **28 passing** across the two changed files.
- Standards consulted: `react.md` (§ 2, § 6, § 10, § 13), `typescript.md` (§ 2), `testing.md`, `design.md`.

### Design-doc accuracy
- **`design.md`: no changes required** — UI focus-management behavior, no data-model / API-surface / architecture change.
- **`react.md` § 10 (unanchorable — file not in this diff): 🔵 Suggestion.** The § 10 "How:" paragraph documents the `useModalA11y` contract and was updated when the hook landed (PR #221). This PR extends that contract — a new third param (`restoreFocusRef`) and the *why* behind it (a parent going `inert` on the same commit synchronously blurs the trigger before the child's mount effect, so `document.activeElement` is the wrong restore target). The existing text stays accurate, just incomplete — hence a suggestion, not a blocker. Proposed addition to the "How:" paragraph:
  > For an overlay layered on top of another overlay (e.g. `RunEntrySheet`'s drink/setup pickers over the sheet), mark the parent `inert` and pass `active={false}` to suspend its trap while the child owns focus. Because making the parent `inert` synchronously blurs the focused trigger before the child mounts, capture the trigger at click time (`e.currentTarget`) and pass it as `useModalA11y`'s `restoreFocusRef` so focus restores to the real opener rather than `<body>`.

### Test quality
A `test-quality-review` mutation-test pass (broke the implementation four ways, confirmed the right tests went red each time) returned **no findings** — notable given the jsdom-`inert` hazard. The two highest-risk tests deliberately use `fireEvent.click` (which doesn't move focus in jsdom) to reproduce the inert-blurred state *without* relying on jsdom implementing `inert`, and they fail when `restoreFocusRef` handling is stripped — so they pin the explicit-ref path rather than passing by jsdom luck. The `inert` test asserts the *attribute* (the testable contract that causes the real-browser blur), both seal and lift.

### What's good
- **The load-bearing correctness point: the pickers render as *siblings* of the inert sheet, not descendants** (the sheet dialog div closes before the `SubPickerOverlay` blocks, under the non-inert root). If they'd been nested inside the `inert` sheet, `inert` would inherit and the picker itself couldn't take focus — and jsdom wouldn't catch it. Getting the DOM placement right here is the whole game, and it's correct.
- The re-seat guard (`!node.contains(document.activeElement)`) is the right fix for close-stealing-focus, and the effect-ordering reasoning (cleanups run before effects, so the trigger is refocused before the resumed trap checks) holds.
- Thorough JSDoc on the new param + inline comments explaining the inert-blur / Safari-click *why*; `aria-labelledby` from the visible heading via `useId()`, matching `CreateSessionModal`.

frontend/src/components/RunEntrySheet.tsx
```diff
@@ -418,65 +437,96 @@ export function RunEntrySheet({
...
+          />
+        </SubPickerOverlay>
       )}
     </div>
   );
 }
 
+// ── Sub-picker overlay (full-screen drink / race-setup picker) ──────
+
+interface SubPickerOverlayProps {
```
brendanbyrne, line 483: 🔵 **Suggestion** — `SubPickerOverlayProps` uses `interface`; `typescript.md` § 2 defaults to `type` (interface only for declaration merging, which Props don't need) and specifies `consistent-type-definitions: ["error", "type"]`. It trips that lint rule (currently at `warn`).

That said, it matches the file's 4 existing Props interfaces (`TimeFields`, `RunEntrySheetProps`, `TimeInputGroupProps`, `SlideToConfirmProps`), all pre-dating this PR — so the clean fix is a file-wide `interface`→`type` pass, not a one-off here, and matching local style was a defensible call. Flagging for the record, not as a blocker.

frontend/src/components/RunEntrySheet.tsx
```diff
@@ -418,65 +437,96 @@ export function RunEntrySheet({
...
+  return (
+    <div
+      ref={ref}
+      role="dialog"
+      aria-modal="true"
+      aria-labelledby={headingId}
+      className="fixed inset-0 z-50 bg-white flex flex-col"
+    >
+      <div className="flex items-center justify-between px-4 py-3 border-b border-gray-200">
+        <button onClick={onClose} className="text-sm text-brand-primary">
```
brendanbyrne: 🔵 **Suggestion** — The `Back` button has no vertical padding of its own (the `py-3` is on the parent flex row, not the button's hit area), so its tap target is ~20px tall vs. the ≥44×44px floor in `react.md` § 10.

Pre-existing — extracted with its className unchanged — so not strictly this PR's regression. But it's the primary control of the dialog this PR is hardening for a11y, so `min-h-[44px] px-2 -mx-2 flex items-center` (or similar) would round out the accessibility story. Optional.

brendanbyrne:
Thanks for the review! Addressed in 77f38c0:

1. **react.md § 10 contract** ✅ — extended the modal "How:" paragraph with the overlay-over-overlay pattern (parent `inert` + `active={false}`; capture the trigger at click time → child's `restoreFocusRef`, since the inert flip blurs the trigger before the child mounts) and added a `## Document history` entry. Folded into this branch so the doc + the `restoreFocusRef` it documents land together.

2. **`SubPickerOverlayProps` `interface`→`type`** — left as-is, per your note: matching the file's 4 existing Props interfaces is the local-consistent choice, and the clean fix is the file-wide pass that #185 (compliance) owns. Flagged in the commit message so it's not forgotten.

3. **Back button tap target** ✅ — now `min-h-touch` (44px) and `inline-flex items-center px-2 -mx-2`. Made it **layout-neutral**: dropped the parent row's `py-3` so the row height is driven by the 44px button instead — the header keeps its height and the full button area is tappable, rather than growing the header by ~24px.

Good catch on the sibling-vs-descendant DOM placement being load-bearing — agreed that's the crux, and it's why the pickers render outside the inert sheet div.

PR #235, opened 2026-06-15
185: frontend final lint cleanup, DTO drift-check CI, coverage backfill (PR-H1)

### LINKED ISSUE #185: Frontend final lint cleanup, DTO drift-check CI, and test-coverage backfill

brendanbyrne:
Closes #185. **PR-H1 — the frontend compliance-plan capstone** ([Milestone 9](https://github.com/brendanbyrne/beerio-kart/milestone/9)). With this in, the [frontend compliance plan](https://github.com/brendanbyrne/beerio-kart/blob/main/docs/designs/2026-05-16-frontend-compliance-plan.md) is archive-eligible.

## What & why (by AC)

**Lint cleanup + ratchet flip (AC1–2, AC2b).** Fixed the last **82 warn-level offenders** across 13 files, then flipped **9 rules `warn`→`error`** so they can never re-enter:
- `consistent-type-definitions` (every remaining `interface`→`type`), `consistent-type-specifier-style` (hoisted the last inline-type imports), `no-floating-promises` / `no-misused-promises` (`void`-marked or wrapped fire-and-forget calls), `no-confusing-void-expression` (braced void-returning arrows), `prefer-nullish-coalescing`, `no-unnecessary-type-arguments`, `restrict-template-expressions`, and `no-unnecessary-condition`.
- The `no-unnecessary-condition` cases were the only judgement calls: `RunEntrySheet`'s `handleSubmit` guard and the lap-sum warning were genuinely redundant — TypeScript's **aliased-condition narrowing** already proves `parsedTotal`/etc. non-null from the single `canSubmit` check (verified by typecheck), so collapsing them is safe *and* removes dead branches. The three `TimeInputGroup` field checks became exhaustive `else`s.
- **AC1/AC2b confirmed, no code change:** `api/types.ts` already models every status/kind/ruleset/source as a literal union (PR-B1), and there are **zero TS `enum`s** — the `no-restricted-syntax` ban (now at `error`, lint passing) enforces it.

**DTO drift-check CI (AC3–4).** New [`DTO Drift`](.github/workflows/dto-drift.yml) workflow + [script](.github/scripts/dto-drift-check.sh) fail any PR that edits a backend **wire-contract** file without also touching `frontend/src/api/types.ts` (the hand-maintained Zod mirror, ADR-0039). There's no single `backend/src/dtos/` dir, so the check watches an **explicit path list** — `routes/`, `domain/enums.rs`, and the serialized `services/` DTO modules — duplicated in the workflow `paths:` and the script regex (extend both when a new DTO module lands). False positives clear with a one-line touch. Documented in `typescript.md` §11; `rust-to-ts-codegen.md` §"Cheap drift detection" marked implemented.

**Coverage backfill + gate flip (AC5–6).** Backfilled the stragglers and flipped the Codecov **frontend project status from informational to pass/fail** (`design.md` §Coverage & CI updated). Note: with no branch-protection rule on `main`, this status reports a red ✗ on a regression but doesn't itself block merge — enforcement is a follow-up (a `main` branch-protection rule listing these checks as required).

| File | Stmts before → after |
|---|---|
| Profile.tsx | 48% → **88%** |
| Home.tsx | 79% → **91%** |
| Onboarding.tsx | 79% → **86%** |
| **Overall** | 84.8/74.6 → **89.0% stmts / 80.9% branch** |

Patch coverage on changed product lines ≈ **90.6%** (gate: 80%). Remaining uncovered changed lines are deliberately left: `RunEntrySheet`'s slide-to-confirm drag math and time-field auto-advance edge sequences (jsdom-hostile — no real layout/IME), plus a few cosmetic toggles (header/history expand, image-error). Per `typescript.md` §12 these are coverage-theater risks, not silent behavior. The backfill was a mop-up, not multi-day — earlier PRs largely held the standard; Profile was the one notable gap.

## Manual test steps

```bash
cd frontend
bun install
bun run lint        # 0 problems (9 newly-flipped rules now at error)
bun run typecheck   # passes
bun run test        # 270 pass (was 254)
bun run test:coverage   # ~89% stmts / ~81% branch
```

Smoke-test the touched promise handlers (all changes are behavior-preserving) — from repo root `bun run dev`, then:
- **Login** → redirects home; **Register** → onboarding.
- **Onboarding:** complete *or* skip race setup → drink phase; skip drink → home.
- **Home:** "Start a Session" → create → lands on the session; with an active session the button reads "Jump to Current Session" and jumps; clicking a session in the list opens it.
- **Session:** Next Track / Skip Track (as host), Join / Leave, Submit Time.
- **Profile:** edit race setup, edit drink, change password, **Log Out** → login.

Confirm no console errors and every navigation works.

**DTO Drift check:** this PR touches no backend DTO, so the new workflow runs and passes ("No backend DTO files changed"). To see it fail, a later PR that edits e.g. `backend/src/domain/enums.rs` without updating `types.ts` will be blocked.


brendanbyrne (commented):
## Review: lint cleanup, DTO drift CI, coverage backfill

**Verdict: Request Changes** — one watch-list gap (inline on the script + workflow) that lets real wire-contract drift slip past the new check; everything else is suggestion-grade.

**Verified:** `bun run lint` ✅ clean · `bun run typecheck` ✅ clean · `bun run test` ✅ 268 pass (PR body says 264 — 4 added since; worth refreshing). Action SHA pin `df4cb1c…` confirmed to dereference from the annotated `v6.0.3` tag, consistent across all 4 workflows. `RunEntrySheet`'s `no-unnecessary-condition` removals verified behavior-preserving — the passing typecheck on the `CreateRunRequest` body assignment is the proof that `canSubmit`'s aliased-condition narrowing holds.

### 🔵 Discussion (not line-anchored)

**"Blocking" gates are advisory until `main` is protected.** `main` has no branch protection (the API returns 404 "Branch not protected"). Flipping codecov `informational: false` and writing "all four gates now block" / "fails any PR" all produce a red ✗ *status*, but with no rule requiring these as *required status checks*, nothing prevents a merge. (Pre-existing for the backend + frontend-patch gates; this PR extends the claim to "all four," so noting it now.) If gating is the intent, the missing piece is a branch-protection rule on `main` — in GitHub, a check reporting failure ≠ a check blocking merge.

**Path-filtered + required-check gotcha (forward-looking).** If `DTO Drift` is later made a *required* status check, the `paths:` filter means PRs that touch none of those files never report it → stuck "pending" → unmergeable. Standard fix when you get there: an always-runs job that conditionally skips. Not actionable today (no branch protection yet).

### Design-doc accuracy

design.md § Coverage & CI, typescript.md § 11, and rust-to-ts-codegen.md are all updated in-PR, accurately, with proper history entries — best case. Only wrinkle is the "all four gates now block" wording vs. the no-branch-protection reality above.

### What's good

The ratchet-only-tightens eslint pattern with per-rule "why it's now zero" comments; the safe `drinkType ?? (…)` fix (object-or-null, so truthy ≡ nullish — the case `prefer-nullish-coalescing` can get wrong, and didn't); drift-script hygiene (least-privilege `permissions: contents: read`, `persist-credentials: false`, concurrency cancellation, three-dot `BASE...HEAD` diff with an explaining comment, explicit "extend BOTH" notes on the duplicated lists); and the coverage test files (Home / Onboarding / Session / DrinkTypeSelector / RaceSetupPicker) that assert real user-visible outcomes through MSW/DOM rather than mock choreography.

.github/scripts/dto-drift-check.sh
```diff
@@ -0,0 +1,57 @@
...
+# Inputs (env): BASE_SHA, HEAD_SHA — the PR's base-branch tip and head commit.
+set -euo pipefail
+
+TYPES_FILE='frontend/src/api/types.ts'
+
+# Backend files whose change implies a possible wire-contract change:
+#   - routes/**         — the request/response structs returned by handlers
+#   - domain/enums.rs   — SessionStatus / SessionRuleset (serialized enums)
+#   - the services/ modules that define serialized DTOs
+DTO_PATHS_REGEX='^backend/src/(routes/|domain/enums\.rs$|services/sessions/types\.rs$|services/sessions/detail\.rs$|services/runs/read\.rs$|services/users\.rs$|services/notifications\.rs$)'
```
brendanbyrne: 🟡 **Important** — This watch-list is already incomplete. Two *existing* serialized wire DTOs aren't matched here:

- **`services/sessions/lifecycle.rs` → `SessionSummary`** — the `GET /sessions` list response (returned at `routes/sessions.rs:72` as `Json<Vec<SessionSummary>>`), mirrored field-for-field at `frontend/src/api/types.ts:144-146`. Its siblings `sessions/types.rs` and `sessions/detail.rs` *are* watched — the asymmetry is the tell. `SessionSummary` is *defined* in `services/` and only *referenced* from `routes/`, so the "routes/** catches response structs" assumption misses it: a `participant_count`→`player_count` rename here touches no watched path → check stays green → frontend Zod parse breaks at runtime (exactly what this check exists to prevent).
- **`services/auth.rs` → `AccessClaims`** — the JWT payload the frontend decodes via `AccessTokenPayloadSchema` (`types.ts:337`, reads `sub`/`username`). Narrower but the same class of gap. (The `{access_token}` envelope itself is fine — it's in `routes/auth.rs`, which is watched.)

**Fix:** add `services/sessions/lifecycle.rs` and `services/auth.rs` to this regex **and** the workflow `paths:`. If `auth.rs` is deemed out of scope, document that in typescript.md § 11 — and while there, note why `domain/strings.rs` newtypes are excluded but `domain/enums.rs` is watched (that asymmetry is currently undocumented). Verified by grepping every `Serialize` derive under `services/`+`domain/`: the only unwatched ones are `auth.rs`, `sessions/lifecycle.rs`, `domain/strings.rs`.

.github/workflows/dto-drift.yml
```diff
@@ -0,0 +1,54 @@
...
+# DTO_PATHS_REGEX in .github/scripts/dto-drift-check.sh. When you add a new
+# DTO-bearing backend module, add it to BOTH. See
+# docs/coding-standards/typescript.md § 11.
+on:
+  pull_request:
+    paths:
+      - 'backend/src/routes/**'
+      - 'backend/src/domain/enums.rs'
+      - 'backend/src/services/sessions/types.rs'
+      - 'backend/src/services/sessions/detail.rs'
```
brendanbyrne, line 29: 🟡 **Important** — Same watch-list gap as the script's `DTO_PATHS_REGEX` (see the inline comment there): add `backend/src/services/sessions/lifecycle.rs` (holds `SessionSummary`, the `GET /sessions` response mirrored at `types.ts:144`) and `backend/src/services/auth.rs` (holds `AccessClaims`, mirrored by `AccessTokenPayloadSchema`) here too. Keep this `paths:` list and the script regex in lock-step per the "extend BOTH" comment above.

.github/scripts/dto-drift-check.sh
```diff
@@ -0,0 +1,57 @@
...
+
+# Three-dot diff: changes on the PR head since its merge base with the target
+# branch (ignores commits that landed on the base after the branch point).
+changed="$(git diff --name-only "${BASE_SHA}...${HEAD_SHA}")"
+
+dto_changed="$(printf '%s\n' "${changed}" | grep -E "${DTO_PATHS_REGEX}" || true)"
+types_changed="$(printf '%s\n' "${changed}" | grep -Fx "${TYPES_FILE}" || true)"
+
+if [[ -n "${dto_changed}" && -z "${types_changed}" ]]; then
```
brendanbyrne, line 47: 🔵 **Suggestion** — This failure branch is the core of the check and it's never exercised: this PR only hits the "no DTO changed → pass" path. The logic reads correctly (`set -euo pipefail`, `|| true` on the no-match greps, `grep -Fx` exact match, three-dot `BASE...HEAD` with `fetch-depth: 0`), but a one-time local smoke test with crafted `BASE_SHA`/`HEAD_SHA` — or a deliberate trip in a later PR — would confirm both the fail and the alongside-pass branches before you rely on it as a guard.

frontend/src/pages/Profile.test.tsx
```diff
@@ -152,3 +224,111 @@ describe('Profile password change form', () => {
...
+    const user = userEvent.setup();
+    renderProfile();
+
+    await user.click(screen.getByRole('button', { name: /race setup/i }));
+    await user.click(screen.getByRole('button', { name: 'complete-setup' }));
+
+    await waitFor(() => {
+      expect(apiFetch).toHaveBeenCalledWith(
+        '/api/v1/users/u1',
+        expect.objectContaining({ method: 'PUT' }),
```
brendanbyrne: 🟡 **Important** — This asserts only `{ method: 'PUT' }` to `/api/v1/users/u1` — identical to the drink-save test below. The request **body** is the one thing that distinguishes "save race setup" from "save drink," and it's unchecked: a handler that sent the wrong fields (or even the drink payload) would still pass this test. Pin it (deterministic — the mocked picker fires fixed values):
```js
expect.objectContaining({
  method: 'PUT',
  body: JSON.stringify({
    preferred_character_id: 1, preferred_body_id: 2,
    preferred_wheel_id: 3, preferred_glider_id: 4,
  }),
})
```
Per testing.md (assert the observable outcome / pin the wire payload).

frontend/src/pages/Profile.test.tsx
```diff
@@ -152,3 +224,111 @@ describe('Profile password change form', () => {
...
+    const user = userEvent.setup();
+    renderProfile();
+
+    await user.click(screen.getByRole('button', { name: /preferred drink/i }));
+    await user.click(screen.getByRole('button', { name: 'select-drink' }));
+
+    await waitFor(() => {
+      expect(apiFetch).toHaveBeenCalledWith(
+        '/api/v1/users/u1',
+        expect.objectContaining({ method: 'PUT' }),
```
brendanbyrne: 🟡 **Important** — Same as the race-setup test above: assert the body, not just `method: 'PUT'`. `handleSaveDrinkType` sends a different single-field shape to the same URL + method, so without a body assertion the two "saves" tests are indistinguishable:
```js
expect.objectContaining({ method: 'PUT', body: JSON.stringify({ preferred_drink_type_id: 'd1' }) })
```
🔵 Also: neither save handler has an error-path test (`!res.ok → setSaveError`). If you add one, mock a realistic failure `Response` (status + `{ code }` per api-contract.md § 7), not a bare `{ ok: false }`, so `parseApiError` runs against the real error shape.

brendanbyrne:
Thanks — thorough review. Verified each finding myself; all addressed in `b2dafd5`.

### 🟡 Watch-list gap (the blocker) — fixed
You're right, and it's the exact failure mode the check exists to prevent. I grepped every `Serialize` derive under `services/` + `domain/` and cross-checked against `types.ts`: **`SessionSummary` (`services/sessions/lifecycle.rs`) and `AccessClaims` (`services/auth.rs`) were the only two gaps** — both mirrored on the wire but *defined away from the route that returns them*, so "watch `routes/`" missed them. Added both to `DTO_PATHS_REGEX` **and** the workflow `paths:`, and **end-to-end-tested it**: a `participant_count → player_count` rename in `lifecycle.rs` with no `types.ts` touch now trips the check (previously green). Documented in typescript.md § 11: the full list, the "defined away from the route" subtlety, and the `domain/enums.rs`-watched-but-`strings.rs`/`numeric.rs`-not rationale (an enum variant rename moves the wire string without touching any struct's field list; a newtype serializes transparently and only reaches the wire through a watched DTO field).

### 🟡 Profile save tests asserted only `method:` — fixed
Pinned the request **body** in both saves (deterministic from the mocked picker values), so race-setup vs drink are now distinguishable. Also added the missing **error-path test for each handler** — mocking a contract-shaped failure `Response` (`{ error, code }` per api-contract § 7, status 4xx/5xx) so `parseApiError` runs against the real shape and the `setSaveError` branches are covered. (13 Profile tests now.)

### 🔵 "Blocking" gates are advisory without branch protection — reworded
Correct, and I overclaimed. Confirmed `main` has no protection rule. Reworded codecov.yml + design.md + this PR's body to "pass/fail status, **advisory until `main` carries a branch-protection rule requiring these checks**" — the ✗ surfaces a regression but GitHub won't block the merge without that rule. Setting up branch protection is your repo-admin call; flagging it as the follow-up rather than claiming enforcement.

### 🔵 Script fail-branch never exercised
Acknowledged — and I *did* smoke-test all three branches locally before the first push (DTO-changed-without-types → exit 1; DTO + types together → pass; no-DTO → pass), plus the new end-to-end `lifecycle.rs` trip above. No in-repo shell-test harness, so it's manual verification for now. Added a forward-looking note in `dto-drift.yml`: if it's ever made a *required* check, the path filter needs an always-runs-and-skips pattern (same gotcha you flagged re: path-filtered required checks).

### Nits
- PR body test count refreshed (now 270).
- Action SHA pin / `df4cb1c…` ↔ `v6.0.3` — thanks for re-verifying; reused the repo's existing pin deliberately.

Re-pushed; CI re-running.

PR #236, opened 2026-06-15
195: CI lint/typecheck/clippy/fmt safety-net + branch-protection readiness

### LINKED ISSUE #195: Add CI workflow running frontend and backend lint/typecheck/clippy checks

brendanbyrne:
## Reviewer notes

- **This PR is its own first live test.** It touches CI plumbing
  (`checks.yml` + `detect-changes.sh`), which `detect-changes.sh` classifies as
  *both* halves — so on this PR the `Checks` jobs run frontend lint/typecheck
  **and** backend clippy/fmt for real (not the skip path). `frontend.yml` now
  runs unconditionally; `dto-drift.yml` runs and early-passes (no DTO changed).
  Watch this PR's checks to see the design exercised end-to-end.
- **The ruleset is *not* applied by this PR.** Enabling branch protection is a
  manual `gh api` POST done **after merge** (so the new check names exist on
  `main`). Procedure + rollback are in [`docs/branch-protection.md`](docs/branch-protection.md).
- **Two decisions made with Brendan** (drove the shape): (1) require the Codecov
  patch checks → `frontend.yml` is always-run so an upload always lands and the
  `codecov/patch/*` checks always post; (2) require a PR for **all** changes,
  no bypass → the docs-only direct-to-`main` carve-out is retired here.
- **`codecov/project/*` is deliberately *not* in the required set** — Codecov
  isn't currently emitting it as a check on this repo (only `codecov/patch/*`
  appears on PRs/pushes). Requiring a context that never posts would deadlock
  every PR. Promoting project coverage to a gate is a documented follow-up.
- **Flake caveat:** always-run frontend + required Codecov means a genuinely
  flaky frontend test would block merges (a red suite leaves both
  `codecov/patch/*` checks *pending*, not red — see the Operational note in the
  doc). The fix is to fix the flake, not to relax the gate.
- A review pass (deadlock-safety / correctness / ruleset+docs) caught one real
  bug: `detect-changes.sh` originally failed *closed* (green "all-clear") on an
  unreachable diff range — a required check could pass green having validated
  nothing. Fixed to fail **open** (`run=true`) and covered by a test below.

## Summary

CI had no job running frontend `lint`/`typecheck` or backend `clippy`/`fmt` —
those lived only in the local, `--no-verify`-bypassable lefthook hooks. This adds
a `Checks` workflow that runs them, and converts every merge-gating workflow to an
**always-run + internal-skip** shape so each always posts a status, making them
safe to mark *required*. Ships a ready-to-apply branch-protection ruleset and the
convention updates it implies.

## Linked work

Closes #195.

## How to verify

**A. Watch this PR's checks (the design, exercised).** Confirm these are green:
`Frontend lint & typecheck`, `Backend clippy & fmt`,
`Backend DTO ↔ frontend types sync`, `codecov/patch/frontend`,
`codecov/patch/backend`. (Backend *coverage* `backend.yml` skips — this PR
touches no backend code — and `backend` carries forward; that's expected.)

**B. `detect-changes.sh` discrimination + fail-open (local).**

```bash
cd "$(git rev-parse --show-toplevel)"
emit() { local out; out=$(mktemp); GITHUB_OUTPUT="$out" BASE_SHA="$1" HEAD_SHA="$2" \
  bash .github/scripts/detect-changes.sh "$3" >/dev/null 2>&1; grep '^run=' "$out"; rm -f "$out"; }
echo "frontend src  -> $(emit b5dd75c^ b5dd75c frontend) / $(emit b5dd75c^ b5dd75c backend)  # expect run=true / run=false"
echo "backend only  -> $(emit 2a306fe^ 2a306fe frontend) / $(emit 2a306fe^ 2a306fe backend)  # expect run=false / run=true"
echo "docs only     -> $(emit e549a8a^ e549a8a frontend) / $(emit e549a8a^ e549a8a backend)  # expect run=false / run=false"
# Unreachable range must fail OPEN (not green-pass):
out=$(mktemp); GITHUB_OUTPUT="$out" BASE_SHA=deadbeefdeadbeefdeadbeefdeadbeefdeadbeef HEAD_SHA=HEAD \
  bash .github/scripts/detect-changes.sh frontend; echo "unreachable   -> $(grep '^run=' "$out")  # expect run=true"; rm -f "$out"
```

**C. The four gates pass on the current tree (what `Checks` runs).**

```bash
( cd frontend && bun install --frozen-lockfile && bun run lint && bun run typecheck )
cargo +stable clippy --all-targets -- -D warnings   # from repo root
cargo +nightly fmt --check                          # from repo root
```

**D. Ruleset payload is valid JSON (full apply is post-merge, per the doc).**

```bash
python3 -c "import json; json.load(open('.github/branch-protection-ruleset.json')); print('ok')"
```

After merge, follow [`docs/branch-protection.md`](docs/branch-protection.md) §
"How to apply": confirm the five context names appear on a fresh PR, `POST` the
ruleset, then smoke-test that merge is blocked until checks pass and a direct
`git push origin main` is rejected.

## Author checklist

- [x] Issue linked (`Closes #195`).
- [x] Verified locally: lint, typecheck, clippy (`--all-targets -D warnings`),
  `fmt --check`, frontend suite (270/270), `detect-changes.sh` matrix + fail-open,
  lychee `--offline` (0 errors), YAML/JSON parse.
- [x] Docs updated: `branch-protection.md` (new), `design.md` § Coverage & CI,
  `README` index, and the retired-carve-out edits across CLAUDE.md /
  project-workflow.md / post-merge skill.
- [x] No schema/data change.
- [x] Adversarial review pass; the one real finding (fail-closed bug) is fixed.


brendanbyrne:
### CI surfaced a real issue — clippy toolchain drift (fixed in 4e2c5c3)

This PR's own `Backend clippy & fmt` job (the first time clippy has *ever* run in CI) went red on its first run — exactly the kind of gap #195 exists to close. Root cause: **CI runs clippy on latest-stable (rust 1.96)**, which has the `clippy::duration_suboptimal_units` lint; **local stable is 1.94** (clippy 0.1.94), which doesn't. So 7 pre-existing `Duration::from_secs(60|900|3600)` sites passed the lefthook pre-commit hook locally but failed in CI.

Fixed by switching them to `from_mins`/`from_hours` (verified both compile on 1.94 *and* 1.96; semantically identical — 60s=1min, 900s=15min, 3600s=1hr). The rest of CI was already green, including the key validation: **`codecov/patch/backend` passed via carryforward even though `backend.yml` skipped** (no backend code on the first push), confirming the frontend-always-run → codecov-always-posts design.

### Recommended follow-up: pin the Rust toolchain

Now that clippy is about to become a **required** check, the unpinned `dtolnay/rust-toolchain stable` means *any* future stable release that adds a default lint will turn the required clippy check red on **every open PR** until someone fixes the new violations — blocking all merges, the same class of surprise-breakage this drift just demonstrated. The robust fix is a repo-root `rust-toolchain.toml` pinning a known version so local + CI clippy always agree and the team adopts new lints deliberately (with a matching update to the rust workflows). I kept that **out of this PR** to stay focused — worth a small follow-up issue. Happy to file it.

brendanbyrne:
### Pinned the Rust toolchain (done in this PR, per request)

Added `rust-toolchain.toml` (`channel = "1.96.0"`, components `clippy` + `llvm-tools-preview`, minimal profile) and dropped the explicit `dtolnay/rust-toolchain stable` step from both `checks.yml` and `backend.yml` — bare `cargo clippy` / `cargo llvm-cov` now resolve the pinned stable, and so does the local lefthook hook (rustup walks up from `backend/` to the repo-root file). Local and CI can no longer disagree, and the about-to-be-required clippy check can't surprise-break when a runner bumps stable. Bump the pin deliberately; `detect-changes.sh` now routes a `rust-toolchain.toml` change to the backend clippy job so a bump is validated.

`fmt` stays on floating nightly (`cargo +nightly fmt`) — `rustfmt.toml` uses nightly-only options — matching lefthook. That's a smaller, separately-addressable surface; flag if you want it pinned too.

**CI confirms it works:** clippy and `cargo llvm-cov` both ran green on the pinned 1.96.0 (the toolchain + components auto-install on first `cargo` use). **Heads-up:** your next local backend `cargo` command (or pre-commit hook) will have rustup auto-install 1.96.0 — a one-time ~download to bring your local stable (was 1.94) in line with the pin.

.github/scripts/detect-changes.sh
```diff
@@ -0,0 +1,73 @@
...
+
+AREA="${1:?usage: detect-changes.sh <frontend|backend>}"
+
+# Files that, when changed, mean this area's checks must run. The CI plumbing
+# itself (this script + checks.yml) is in BOTH lists, so a change to it is
+# validated by running both halves.
+CI_PLUMBING='\.github/workflows/checks\.yml$|\.github/scripts/detect-changes\.sh$'
+case "${AREA}" in
+  frontend) AREA_REGEX="^(frontend/|${CI_PLUMBING})" ;;
+  backend)  AREA_REGEX="^(backend/|Cargo\.toml$|Cargo\.lock$|rust-toolchain\.toml$|${CI_PLUMBING})" ;;
```
brendanbyrne: 🟡 **Important** — the backend area regex omits `rustfmt.toml` and `.cargo/config.toml`, both of which exist at the repo root. This is the same bug *class* this PR's fail-open fix addressed: **a required check passing green having validated nothing.**

Confirmed a config-only change to either short-circuits the `Backend clippy & fmt` job:
```
$ printf 'rustfmt.toml\n' | grep -Eq '^(backend/|Cargo\.toml$|Cargo\.lock$|rust-toolchain\.toml$|...)'
NO MATCH (backend clippy/fmt would short-circuit to green)
```
Scenario: a PR edits `rustfmt.toml` → the required check passes green *without* running `cargo +nightly fmt --check` → merges → `main`'s code now disagrees with its own fmt config, and the **next** unrelated backend PR fails fmt with a large diff blaming the wrong author. `.cargo/config.toml` is the same risk for `RUSTFLAGS`/clippy-affecting build settings. (`rust-toolchain.toml` is already covered — good.)

Fix (one line):
```bash
backend)  AREA_REGEX="^(backend/|Cargo\.toml$|Cargo\.lock$|rust-toolchain\.toml$|rustfmt\.toml$|\.cargo/config\.toml$|${CI_PLUMBING})" ;;
```

brendanbyrne (commented):
## Code review — Issue #195 CI safety-net + branch-protection readiness

**Verdict: Approve** (leaving APPROVE for Brendan) — with one 🟡 worth folding into this PR (one-line fix) and two 🔵 doc nits. Inline comments below.

**Verified locally on the pinned 1.96.0 toolchain** (which auto-installed from the new `rust-toolchain.toml` — the mechanism works):
- `cargo clippy --all-targets -- -D warnings` → clean
- `cargo +nightly fmt --check` → clean
- `bun run lint` / `bun run typecheck` → clean (exit 0)
- `cargo test --workspace` → 380+ tests pass, 0 failed
- `detect-changes.sh` discrimination matrix + fail-open → all cases match expected, including unreachable-range → `run=true`
- ruleset JSON valid; the three workflow-sourced required contexts are byte-identical to their job names (incl. the unicode `↔`)

**Design-doc accuracy:** needed updates, they were made, verified accurate against the workflow files — the only inaccuracy is the 🔵 over-statement noted inline on `design.md`.

**Notes (not blocking):**
- The unicode `↔` in the required context is correct today (verified byte-identical to the dto-drift job name); the apply procedure's "confirm the five context names" step covers it. Just a copy-paste/encoding footgun for future edits.
- Test changes (`shutdown.rs`, `timeout.rs`) are value-preserving `from_secs(60)`→`from_mins(1)` swaps in test setup — no assertion/logic change — so no full test-quality pass warranted; reviewed inline.
- `codecov/patch/backend` via carryforward on a *frontend-only* PR is the one path this PR can't self-test (it touches backend code, so `backend.yml` runs here). Worth eyeballing on the first frontend-only PR after the ruleset is applied.

**What's good:** the fail-open fix is genuinely correct (capturing `git diff` in an assignment to distinguish a failed range from "no match" — verified the unreachable path emits `run=true`); the always-run + internal-skip pattern is the right fix for the skipped-required-check deadlock and `branch-protection.md` documents it unusually well, honest about its limits; the toolchain pin closes a real surprise-breakage hole; security hygiene is solid (`contents: read`, `persist-credentials: false`, SHA-pinned actions, no shell-injection surface).

docs/design.md
```diff
@@ -53,9 +53,9 @@ The two sides share two load-bearing design surfaces — the contracts between t
...
+- **CI:** GitHub Actions. The `Checks` workflow runs frontend `lint`/`typecheck` and backend `clippy`/`fmt --check` — a CI backstop for the lefthook pre-commit gates (which are local-only and `--no-verify`-bypassable). The `Backend` workflow runs `cargo-llvm-cov` and the `Frontend` workflow runs `bun run test:coverage`, both uploading to Codecov (`backend` / `frontend` flags). `Backend` is path-filtered to its subtree (Codecov carries the untouched flag forward when it skips); `Frontend` and `Checks` run on **every** PR — `Frontend` so Codecov always gets an upload and posts its checks, `Checks` so its required statuses always post. A `DTO Drift` workflow fails any PR that changes a backend wire-contract file without also updating `frontend/src/api/types.ts`, the hand-maintained Zod mirror (see [`coding-standards/typescript.md`](./coding-standards/typescript.md) § 11). Every merge-gating workflow uses an **always-run + internal-skip** shape (run on every PR, skip the work internally when nothing relevant changed, but still post a status) so it's safe to mark required — see [`branch-protection.md`](./branch-protection.md).
```
brendanbyrne: 🔵 **Suggestion** — "Every merge-gating workflow uses an always-run + internal-skip shape" over-states it: the backend coverage gate (`codecov/patch/backend`) is *not* always-run — `backend.yml` stays path-filtered and the check posts via Codecov flag carryforward. This contradicts the same bullet's earlier "`Backend` is path-filtered." `branch-protection.md` gets it right ("the coverage workflows are the exception that proves the rule").

Suggest: "Every *required-status-posting* workflow uses an always-run + internal-skip shape (backend coverage excepted — it relies on Codecov carryforward; see `branch-protection.md`)."

.claude/skills/post-merge/SKILL.md
```diff
@@ -4,7 +4,7 @@ description: >
   Clean up after a PR has been merged. Determines the branch to delete, checks out
   main, pulls the latest changes, and deletes the merged branch locally and remotely,
   then checks whether the merge now makes any follow-up documentation updates necessary
-  (e.g. design-record or compliance-plan sign-off) and commits those directly to main.
+  (e.g. design-record or compliance-plan sign-off) and opens a small follow-up PR for those.
```
brendanbyrne: 🔵 **Suggestion** — the description drifts from the body. The body (correctly) says the *normal* path is to **confirm the sign-offs already landed in the triggering PR and stop** — a follow-up PR is only the fallback when one was genuinely missed. The description's "opens a small follow-up PR for those" reads as the default, and since the description is what a future Claude Code uses to decide behavior, it could induce exactly the churn the body warns against.

Suggest: "…confirms those sign-offs landed in the triggering PR, or opens a small follow-up PR if one was missed."

.github/scripts/detect-changes.sh
brendanbyrne: Fixed in 271dad6 — added `rustfmt\.toml$` to the backend regex. Good catch; it's the same "validated nothing" class as the fail-open fix.

One adjustment on the other half: `.cargo/config.toml` is **gitignored** (`.gitignore:11:.cargo/` — it's the local-only WSL2 perf config Brendan maintains), so it can never appear in a PR diff and `git diff --name-only` would never list it. Watching it would be dead regex that also misleads a future reader into thinking it's covered. CI also never sees that file, so it can't affect CI clippy/fmt. So I added `rustfmt.toml` only. (No `clippy.toml`/`.clippy.toml` in the repo either.)

docs/design.md
brendanbyrne: Fixed in 271dad6. Reworded to: "Every merge-gating workflow that posts its *own* status uses an always-run + internal-skip shape … the one exception is backend coverage — `Backend` stays path-filtered and its `codecov/patch/backend` check posts via Codecov flag carryforward." Now consistent with the earlier "`Backend` is path-filtered" in the same bullet and with branch-protection.md.

.claude/skills/post-merge/SKILL.md
brendanbyrne: Fixed in 271dad6. Description now reads "…confirms those sign-offs landed in the triggering PR, or opens a small follow-up PR only if one was missed" — the confirm-and-stop default first, follow-up as the fallback, matching the body. Agreed the description is the behavior-driving surface.

brendanbyrne:
### Added: backend coverage always-runs on `main` pushes (sets up `codecov/project/*`) — ec3cf5a

Per discussion, dropped the `paths:` filter from `backend.yml`'s **`push: main`** trigger only (PR runs stay path-filtered, so frontend-only PRs still skip the slow `cargo llvm-cov`). 

**Why:** `codecov/project/backend` is a *base-relative* comparison ("did total coverage regress vs. `main`?"), so it needs a coverage report at the PR's base commit. Many `main` commits had none — a docs- or one-sided commit where the path-filtered coverage workflows skipped uploaded nothing, so Codecov had no base and withheld the project status (while `patch`, an absolute diff threshold, still posted). Now every `main` commit uploads backend coverage (frontend already always-uploads after this PR), so there's always a base; a one-sided PR's missing flag fills via carryforward.

**What this does and doesn't do:**
- ✅ Plumbing so `codecov/project/{frontend,backend}` can start posting once this lands on `main`.
- ❌ Does **not** add `project/*` to the required set — I can't verify it actually posts until base reports exist on `main`, and requiring a not-yet-posting context would deadlock PRs. So *confirm-it-posts → add-to-required* stays a small post-merge follow-up.

Cost: ~1–2 min of `cargo llvm-cov` per merge to `main` (not per PR). CI green on ec3cf5a. `branch-protection.md` + `design.md` updated to match.

PR #240, opened 2026-06-20
209: leave the current session on logout

### LINKED ISSUE #209: Leave the current session on logout so sole-participant sessions close
## Context

When a user logs out while in a session, the frontend's `logout` flow ([useAuth.tsx:110-114](frontend/src/hooks/useAuth.tsx#L110-L114)) POSTs `/auth/logout` and clears tokens, but does not call `leaveSession` first. The user's `session_participants` row stays with `left_at = NULL`, so:

- If they were the **only active participant**, the session stays `active` forever (well, until the hourly stale-session sweeper picks it up — see [`leave_session` in services/sessions/lifecycle.rs](backend/src/services/sessions/lifecycle.rs)).
- If they were **a participant alongside others**, they remain listed as in the session even after logout. Re-logging-in puts them right back into it without an explicit Join action.

The backend side is already in place. `leave_session` calls `transfer_host_or_close`, which returns `HostDisposition::SessionClosed` and invokes `close_session` when no transferable participant remains. So a sole participant calling Leave already closes the session — the fix is purely making logout *call* Leave first.

Surfaced during PR-D2 ([#208](https://github.com/brendanbyrne/beerio-kart/pull/208)) review while smoke-testing logout-from-session.

## Acceptance criteria

- [ ] `useAuth.logout` fetches `getMySession()` and, if a session id comes back, calls `leaveSession(id)` before the existing `/auth/logout` POST + token clear.
- [ ] A `leaveSession` failure does not block logout — wrap in try/catch and continue. Token clear must still happen even if the leave call errors.
- [ ] Add a `useAuth` test covering: (a) sole-participant logout → leaveSession is called and logout completes; (b) leaveSession failure → logout still completes; (c) not-in-a-session logout → leaveSession is *not* called.
- [ ] Manual verification: log in as a single user, join/create a session, log out. The session should disappear from any other open client within one poll cycle (Home list refreshes via the `['sessions']` invalidation; backend marks `ended_at`).
- [ ] No regression in multi-participant case: logging out as one of several participants leaves the session open for the rest, with the host transferring if the logging-out user held the host role.

## Open decisions

- **Always leave vs. only-when-sole:** The simplest semantics is \"logout always leaves your current session.\" This matches typical mobile-app behavior and avoids the extra round-trip to count participants. The alternative — only leave when sole-participant — preserves multi-participant membership across logout/login cycles. Recommend always-leave; flag for confirmation in the PR.
- **What if the leave call hangs (slow network)?** Either time-bound it (e.g., `Promise.race` with a short timeout) or fire-and-forget. Probably fire-and-forget is fine: the worst case is the stale-session sweeper picks it up an hour later.

## References

- Frontend: [useAuth.tsx](frontend/src/hooks/useAuth.tsx), [api/sessions.ts](frontend/src/api/sessions.ts) (`getMySession`, `leaveSession`).
- Backend (no changes expected): [services/sessions/lifecycle.rs:604](backend/src/services/sessions/lifecycle.rs#L604) (`leave_session` + `transfer_host_or_close` choke point).
- Surfaced by: PR #208 smoke test.

brendanbyrne:
## What & why

When a user logged out mid-session, the frontend cleared tokens but never left the session. The `session_participants` row kept `left_at = NULL`, so a **sole participant's session stayed `active`** until the hourly stale-session sweeper closed it (and a multi-participant logout left the user listed as present, auto-rejoining on next login).

`useAuth.logout` now leaves the current session first:

```
getMySession()  →  if (session id) leaveSession(id)  →  POST /auth/logout  →  clearAuth()
```

The backend already does the rest — `leave_session → transfer_host_or_close` closes a session whose last participant departs, and transfers host to the earliest-joined remaining participant otherwise. So this is a purely frontend change that makes logout *call* leave.

### Ordering & error handling
- The leave runs **before** the logout POST and the token clear because `getMySession`/`leaveSession` go through `apiFetch`, which needs the still-live access token (`clearAuth()` nulls it last).
- The leave is **best-effort**: not being in a session, a network error, or a rejected leave is swallowed and sign-out proceeds regardless. The sweeper remains the server-side backstop, so a missed leave self-heals within the hour.

## ⚠️ Open decisions — please confirm

The issue flagged two decisions for confirmation. My choices and reasoning:

1. **Always-leave (chosen) vs. only-when-sole.** Logout leaves whatever session you're in, not just when you're the sole participant. This matches typical mobile-app behavior and avoids a participant-count round-trip. It's safe for the multi-participant case: the backend transfers host / keeps the session open for the others (verified in `services/sessions/lifecycle.rs`). The alternative would preserve multi-participant membership across logout/login — say the word if you'd prefer that and I'll flip both the code and the doc note in this same PR.

2. **Await sequentially, no timeout (chosen) vs. fire-and-forget / `Promise.race` timeout.** I await the leave so it's guaranteed sent before the token is cleared, and I rely on the browser's own fetch timeout rather than adding one — consistent with every other call in `apiFetch` (none are time-bounded). Trade-off: on a dead network the logout (and the redirect to /login) waits for the browser to time out the leave, with no busy-state on the button. That no-feedback-on-slow-network behavior is **pre-existing** (the logout POST already had it; `Profile.tsx` is unchanged here). If you want a snappier logout, a fire-and-forget leave or a short `Promise.race` timeout is a small follow-up.

## Changes

- **`frontend/src/hooks/useAuth.tsx`** — `logout` leaves the current session first (best-effort), imports `getMySession`/`leaveSession`.
- **`frontend/src/hooks/useAuth.test.tsx`** — 4 new tests (below).
- **`docs/user-workflows.md`** — § 1.5 step 6 records the logout-leaves behavior + a history entry.
- **`docs/api-contract.md`** — § 1.5 adds the previously-undocumented `GET /sessions/mine` that the auth flow now depends on.

## Automated tests (`useAuth.test.tsx`)

All four assert against MSW handlers validated against the real contract (`MySessionResponseSchema`, `leaveSession` throwing `ApiErrorException`, 204/500 status codes):

1. **Ordered leave-before-logout** — `calls === ['mine', 'leave', 'logout']`, leave gets the live session id, auth cleared.
2. **Leave failure still completes** — `/leave` 500s; asserts the leave was *attempted first* (`['mine','leave','logout']`) and sign-out still finished.
3. **Not in a session** — `/mine` returns `{ session_id: null }` → leave **not** called, logout POST still fired.
4. **`/mine` errors** — `/mine` 500s → `getMySession` swallows and returns null → leave skipped, sign-out still completes.

`bun run lint` ✓ · `bun run typecheck` ✓ · `bun run test:coverage` ✓ (274 tests; new logout lines fully covered).

## Manual verification

Two clients (or two browsers / one incognito), backend + frontend running (`cargo run` and `bun run dev` from the repo root):

1. Log in as user **A** on client 1; create or join a session. Confirm it appears in the Home session list on client 2 (logged in as a different user, or just polling the public list).
2. On client 1, log out (Profile → **Log Out**).
3. **Expected:** within one poll cycle (~2–3 s) the session disappears from client 2's Home list — the backend marked it `ended_at`. A was the sole participant, so the session closed.
4. **Multi-participant no-regression:** A and B both in one session; A logs out. Session stays open for B; if A was host, host transfers to B. (Verify B's view still shows the session active with B as host.)
5. **Best-effort:** with the network throttled/offline at logout, confirm the sign-out still completes and lands on /login (the leave is swallowed; the sweeper would clean up server-side).

Closes #209

### DIFF
--- a/frontend/src/hooks/useAuth.tsx
+++ b/frontend/src/hooks/useAuth.tsx
@@ -14,6 +14,7 @@ import {
 } from '../api/client';
 import type { AuthFailureReason } from '../api/client';
 import { parseApiError, parseBody } from '../api/result';
+import { getMySession, leaveSession } from '../api/sessions';
 import {
   AccessTokenPayloadSchema,
   AuthSessionSchema,
@@ -168,6 +169,23 @@ export function AuthProvider({ children }: { children: ReactNode }) {
   };
 
   const logout = async () => {
+    // Leave the current session before signing out (#209). A sole participant
+    // who logs out without leaving would keep the session `active` until the
+    // hourly stale-session sweeper closes it; leaving first closes it at once
+    // (the backend's `leave_session` closes a session whose last participant
+    // departs). This runs *before* the logout POST and the token clear because
+    // both `getMySession` and `leaveSession` go through `apiFetch`, which needs
+    // the still-live access token. Best-effort: not being in a session, a
+    // network error, or a rejected leave must not block the sign-out, so any
+    // failure is swallowed and we fall through to clear local state regardless.
+    // The stale-session sweeper remains the server-side backstop.
+    try {
+      const sessionId = await getMySession();
+      if (sessionId) await leaveSession(sessionId);
+    } catch {
+      // Best-effort — sign-out proceeds regardless (see above).
+    }
+
     try {
       // Send the access token so the server can bump refresh_token_version
       const token = getAccessToken();
--- a/frontend/src/hooks/useAuth.tsx
+++ b/frontend/src/hooks/useAuth.tsx
@@ -182,8 +182,13 @@ export function AuthProvider({ children }: { children: ReactNode }) {
     try {
       const sessionId = await getMySession();
       if (sessionId) await leaveSession(sessionId);
-    } catch {
-      // Best-effort — sign-out proceeds regardless (see above).
+    } catch (e) {
+      // Best-effort — sign-out proceeds regardless (see above). getMySession
+      // returns null (no throw) when you're simply not in a session, so a
+      // throw here is abnormal — a network failure or a genuinely-rejected
+      // leave — and otherwise invisible until the sweeper catches it. Log at
+      // debug so a persistently-broken leave is diagnosable without noise.
+      console.debug('logout: best-effort leave failed', e);
     }
 
     try {
--- a/docs/api-contract.md
+++ b/docs/api-contract.md
@@ -68,6 +68,7 @@ A drink type is identified by `(name, alcoholic)`, matched case-insensitively on
 POST   /sessions                   Create a new session (choose ruleset)
 GET    /sessions                   List active sessions (sorted by most recent activity)
 GET    /sessions/:id               Get session details (participants, current race, state)
+GET    /sessions/mine              Get the caller's current session (`{ session_id }`, null if not in one)
 POST   /sessions/:id/join          Join a session (dedicated endpoint — designed for future password support)
 POST   /sessions/:id/leave         Leave a session (if you were the last to leave, the session closes and unresolved pending races are dropped)
 POST   /sessions/:id/next-track    Trigger next track selection (host or chooser, depending on ruleset)
--- a/docs/user-workflows.md
+++ b/docs/user-workflows.md
@@ -62,6 +62,7 @@ Earmarked: the track selection sub-workflow (how the chooser browses/searches fo
 3. If the leaving player is the host: host role transfers to the earliest-joined remaining participant.
 4. Session ends when all participants have left.
 5. Sessions with no activity of any kind for an hour — no new race, run, join, leave, or skip — auto-close on the next sweeper pass. No further run submissions accepted, and any unresolved pending races are dropped.
+6. Logging out leaves your current session first (always — even mid-session), so a sole-participant session closes immediately on logout rather than lingering until the sweeper. Best-effort: a failed leave never blocks the sign-out, with the sweeper as the backstop.
 
 ### 1.6 Checking personal stats
 
@@ -181,3 +182,4 @@ Streamlined compared to standalone entry — the track is already known from the
 - 2026-05-08 — Updated the § 0 cross-reference: `workflow.md` → `project-workflow.md` (operational doc renamed for clarity — see that file's history). Dropped the "note the singular 'workflow'" caveat, no longer needed once the filename is unambiguous.
 - 2026-05-08 — Demoted the broken `[workflow.md](./workflow.md)` markdown link in the 2026-05-06 initial-creation history bullet to plain backticked text. The file no longer exists at that path (renamed to `project-workflow.md` on 2026-05-08); the rename history is documented in the bullet immediately above. PR 5 of the docs restructure (plan deviation — Cowork's PR 5 handoff did not anticipate this link, surfaced when lychee `fail: true` flip would otherwise turn it into a CI failure).
 - 2026-05-16 — Rewrote § 1.5 step 2 for ADR-0037: pending races stay submittable while the session is active and rejoining works as long as someone keeps the session alive; if everyone leaves, the session ends and pending races are dropped (with a notification). The per-race 1-hour window is gone. Step 5 (auto-close) updated for the five-signal sweeper predicate and the drop-on-close behavior. § 1.2 step 6 "(oldest expire first)" → "(oldest shown first)" — no per-race expiry remains. Issue [#51](https://github.com/brendanbyrne/beerio-kart/issues/51).
+- 2026-06-20 — § 1.5 step 6 added: logout now leaves your current session before signing out ([#209](https://github.com/brendanbyrne/beerio-kart/issues/209)). Chose **always-leave** (not only-when-sole) so a sole-participant session closes on logout instead of waiting for the hourly sweeper; best-effort, so a failed leave never blocks the sign-out.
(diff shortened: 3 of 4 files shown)


frontend/src/hooks/useAuth.tsx
```diff
@@ -168,6 +169,23 @@ export function AuthProvider({ children }: { children: ReactNode }) {
...
+    // (the backend's `leave_session` closes a session whose last participant
+    // departs). This runs *before* the logout POST and the token clear because
+    // both `getMySession` and `leaveSession` go through `apiFetch`, which needs
+    // the still-live access token. Best-effort: not being in a session, a
+    // network error, or a rejected leave must not block the sign-out, so any
+    // failure is swallowed and we fall through to clear local state regardless.
+    // The stale-session sweeper remains the server-side backstop.
+    try {
+      const sessionId = await getMySession();
+      if (sessionId) await leaveSession(sessionId);
```
brendanbyrne, line 184: 🔵 **Confirm — `await`-sequential vs. the issue's fire-and-forget suggestion.** Issue #209's open decisions leaned *fire-and-forget* for the slow-network case; this PR awaits `getMySession()` then `leaveSession()` before the logout POST. I think the await choice is the **more correct** one, not just defensible: firing the leave without awaiting would let the subsequent `/auth/logout` POST (which bumps `refresh_token_version`) and `clearAuth()` (which nulls the access token) race the leave's `apiFetch`, which needs that live token. Awaiting removes that race. The only cost is the documented one — on a dead network, logout waits on the browser's fetch timeout with no busy-state — and that's **pre-existing** for the logout POST below, so no regression. No change needed to merge; just confirm you want to keep `await` vs. a `Promise.race` timeout follow-up for snappier logout.

frontend/src/hooks/useAuth.tsx
```diff
@@ -168,6 +169,23 @@ export function AuthProvider({ children }: { children: ReactNode }) {
...
+    } catch {
```
brendanbyrne: 🔵 **Suggestion (optional) — consider a debug log in the swallow.** The normal not-in-a-session path doesn't throw (`getMySession` returns `null` on a non-ok), so anything reaching this `catch` is abnormal — a network failure or a genuinely-rejected leave. A one-line `console.debug('logout: best-effort leave failed', e)` would make a persistently-broken leave diagnosable without blocking sign-out. Precedent for logging non-fatal anomalies: `api/result.ts` (`console.error` on parse-shape mismatch). Opinion, not a rule — fine to leave as-is given the intentional best-effort design + sweeper backstop.

brendanbyrne (commented):
## Code review — leave the current session on logout

**Verdict: Approve** — clean, well-scoped, well-tested, and the implementation matches every Issue #209 acceptance criterion. One decision to confirm and one optional polish below; no blockers. (Approval/merge stays yours — this review is `COMMENT`.)

**Areas covered:** `react.md` (§ 4, § 7, § 13), `testing.md` (§ 1–2), `api-contract.md` (§ 1.5), `user-workflows.md`, `design.md`, plus the backend handler and Issue #209 acceptance criteria for cross-checking.

**Automated checks:** `bun run lint` ✓ · `bun run typecheck` ✓ · `useAuth.test.tsx` 12/12 ✓. No Rust in the diff.

### Design-doc accuracy — best case ✅

- **`api-contract.md` § 1.5** adds the previously-*undocumented* `GET /sessions/mine`. Good catch — the endpoint already existed in the backend (`routes/sessions.rs::my_session`) and frontend (`getMySession`), and the auth flow now formally depends on it. The `{ session_id }`, null-if-not-in-one description matches `MySessionResponse { session_id: Option<SessionId> }` exactly. Leaving it un-annotated for auth matches the catalog's per-section format.
- **`user-workflows.md` § 1.5 step 6 + history entry** — accurate; records the always-leave + best-effort behavior and the rationale.
- **`design.md` / `data-model.md`** — no architecture or schema change → no update required.

### What's good

- **Acceptance criteria fully met, plus extra.** All three required tests (sole-participant, leave-fails, not-in-session) are present, and a 4th covers the `/mine`-errors branch the issue didn't ask for.
- **Test quality is solid** (delegated to `test-quality-review`). Every test pins the observable outcome (`isAuthenticated`/`user` cleared) per `testing.md` § 2, not just choreography. The `calls === ['mine','leave','logout']` ordering assertion is genuinely failure-sensitive — MSW has no fallback handlers, so a wrong/early/skipped call fails rather than silently passing. Test 1 also pins `leftSessionId === 's1'`, proving the *live* session id reached `leaveSession`.
- The **always-leave decision is correctly confirmed against the issue**, and the multi-participant no-regression case is reasoned through (backend transfers host / keeps the session open).
- The **explanatory comment** on the new block captures the *why* and the load-bearing ordering constraint — exactly what a future editor needs.
- `clearAuth`'s `useCallback` carve-out (`react.md` § 7(d)) is untouched and still correct.

Two inline notes follow.

brendanbyrne:
Thanks for the review! Both notes addressed:

- **🔵 `await` vs fire-and-forget** — confirming **keep `await`**. Agreed it's the more-correct choice, not just defensible: a non-awaited leave would let the logout POST (`refresh_token_version` bump) and `clearAuth()` (token null) race the leave's `apiFetch`, which needs the live token. The dead-network wait is pre-existing on the logout POST, so no regression. A `Promise.race` timeout stays a trivially-revisitable follow-up if logout snappiness ever becomes a real complaint — not worth the partial-operation complexity now.
- **🔵 debug log in the swallow** — added in `fac29bc`. `getMySession` returns `null` (no throw) when you're simply not in a session, so reaching that catch is abnormal; `console.debug('logout: best-effort leave failed', e)` makes a persistently-broken leave diagnosable without noise, matching the non-fatal-anomaly logging in `api/result.ts`. The line is covered by the existing leave-failure test.

No behavior change; re-running CI.

